Issue #19, opened 2022-05-31
Testing: Unit and E2E frameworks


ann-kilzer:
Let's try out Cypress and see what we think

ermish:
PR #44 resolves this

PR #44, opened 2022-07-26
build: install cypress for testing

theyokohamalife:
# Part of Issue #19 

1. Installs [Cypress](https://docs.cypress.io/)  so we can begin writing tests for the front end

## How to run

1. To get started, open the app with the following command:  `yarn run cypress open`
2. Select E2E Testing

## TODO

1. We'll need to write some specs. There are some sample ones that Cypress can install if you select the option. I'll spend some time studying them, but feel free to write specs in the meantime.


ann-kilzer (changes requested):
Yay Cypress, it's great.

What are these `.pnp` files... They don't look to be part of cypress, and are gigantic. Can we remove them or move them into another PR, if it's related to a different feature

theyokohamalife:
@ann-kilzer I thought the .pnp files were generated when I selected the E2E testing. It's hard to say, because running the project results in 1000+ changes due to the linting rules. I'll have a look at it again tonight and will confirm if they should be there or not.

ann-kilzer:
I'm doing cypress e2e at work and this is not part of the basics. Anyway the file size seems really unusual for a hello world

ermish:
The .pnp files are part of yarn. I fixed our gitignore and yarn config to properly support yarn 2 formats and netlify builds, but it's not merged yet.

ermish:
The pnp files are a much faster/smaller replacement of node-modules. Like 2gb vs 200mb difference!
However, it's still very experimental so I configured our project to use node-modules for now.

ann-kilzer:
> The pnp files are a much faster/smaller replacement of node-modules. Like 2gb vs 200mb difference!
> However, it's still very experimental so I configured our project to use node-modules for now.

Interesting. 

Should these be in .gitignore in that case? Seeing updates of `+44,854` looks like a lot of information. Or is it more like the `yarn.lock` file where there are just big changes when adding new dependencies

ann-kilzer:
> running the project results in 1000+ changes due to the linting rules

Do we need to configure the linter to ignore some things? Running the code should not be trigging the linter...

ermish:
> > running the project results in 1000+ changes due to the linting rules
> 
> Do we need to configure the linter to ignore some things? Running the code should not be trigging the linter...
> 
> we never want to lint:
> 
> * imported dependencies like node_modules
> * auto-generated code

yes and yes!

ermish:
> > The pnp files are a much faster/smaller replacement of node-modules. Like 2gb vs 200mb difference!
> > However, it's still very experimental so I configured our project to use node-modules for now.
> 
> Interesting.
> 
> Should these be in .gitignore in that case? Seeing updates of `+44,854` looks like a lot of information. Or is it more like the `yarn.lock` file where there are just big changes when adding new dependencies

excellent question!
@ann-kilzer @theyokohamalife  I highly recommending just reading this page so you can be yarn masters :) 
https://yarnpkg.com/features/pnp#the-node_modules-problem

We have two options: check in the dependencies or not. Here's some key points
- Checking them in means your app has everything needed to run (even offline)
- Not checking them in means smaller repo, but each person will download separately. 
- Still unclear how caching works during production builds. This might be better or worse 🤷 
- I was a little worried on Netlify support and caching, so I .gitignored them for now until this feature has been out a little longer

ermish:
@theyokohamalife I merged in #30, so you should be able to rebase and get the new yarn configs to solve the `pnp` file issue

ermish (approved):
LGTM :shipit:

PR #62, opened 2022-10-17
17 husky linting

### LINKED ISSUE #17: Github actions and husky for eslint / prettier
set up automation for linter and prettier

Based on discussion, it's preferred to not lint on commit, rather on push. Developers can configure local linters based on preference.

ann-kilzer:
Resolves #17 

# What changed
- add husky hooks
- set up a script to run linter

# Testing instructions
1. run `yarn prepare`
2. make a new branch `git checkout -b whatever`
3. commit a change
4. push to origin
5. husky should run

1. run `yarn lint`

1. run `yarn lint:fix`


package.json
```diff
@@ -3,8 +3,11 @@
   "version": "1.0.0",
   "private": true,
   "scripts": {
+    "prepare": "husky install",
     "dev": "yarn && nuxi dev",
     "build": "nuxi generate",
+    "lint": "yarn && eslint . --ext .js,.ts,.json",
+    "lint:check": "yarn && eslint . --ext .js,.ts,.json",
```
ermish, line 10: what;s the thought between having lint and lint:check? Could we just use lint?

ann-kilzer: Ahh... they were different when prettier was included but now are the same. If we decide not to add prettier, I will consolidate them

ermish: ah makes sense. Let's combine for now. We might not actually have to include prettier in the codebase, but rather just a local plugin for your IDE.

ermish (approved):
LGTM. added one comment

PR #81, opened 2022-10-21
Update README.md

ann-kilzer:
Switch the order of setup instructions

Aiden D. identified that we need yarn dependencies installed before running `yarn prepare` so I updated the README


theyokohamalife (approved):
LGTM

PR #85, opened 2022-10-22
Fix eslint warnings

RageZBla:
# What changed

- disable `vue/multi-word-component-names` rule as per conversation with @ann-kilzer on slack
- add default value to prop where required

# Testing instructions

run `yarn lint` and 0 warning should appear.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -113,7 +113,7 @@ module.exports = {
         'no-shadow': 2,
 
         // Vue
-        'vue/multi-word-component-names': 0,
+        'vue/multi-word-component-names': 1,
         'vue/html-indent': ['error', 4, {
             attribute: 1,
             baseIndent: 1,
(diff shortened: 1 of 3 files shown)


.eslintrc.js
```diff
@@ -113,7 +113,7 @@ module.exports = {
         'no-shadow': 2,
 
         // Vue
-        'vue/multi-word-component-names': 1,
+        'vue/multi-word-component-names': 0,
```
ermish, line 116: @ann-kilzer @RageZBla can we keep this as a warning? 
It's a good practice and we should try to enforce it, however, I moved to warning temporarily so we can pass build with our current code. We need to update our component names to follow vue standards (I'll create a ticket for this)

ermish (changes requested):
Everything looks good except the one change I highlighted

.eslintrc.js
RageZBla: @ermish thanks for the extra context. I have re-enabled the lint rule.

ermish (approved):
:shipit: thanks @RageZBla !

PR #96, opened 2022-10-25
Fix ESLint Workflows

### LINKED ISSUE #93: bug: Fix eslint CI
# Requirements 

- the eslint CI runner currently fails on the `install` step. Please fix!

### LINKED ISSUE #94: bug: eslint nuxt import issue
# Requirements

- Nuxt magically imports the `defineNuxtPlugin` method in the `vue-gtag.client` component. This causes an eslint lint bug. We should add a proper import for the method. 
- Note: `nuxt` is invalid to import from.
- We can't comment this out either as the CI linter doesn't accept eslint comments.

deek87:
Resolves #93 

Also fixes #94

# What changed

- [x] Changed the ESLint workflow to use node 16 and 18 (both LTS)
- [x] Fixed the ESLint problem ( #94)
- [x] Added the microsoft linter to dev
- [x] Added new commands for CI yarn setup:ci and yarn lint:ci
- [x] Use Yarn in workflows rather than npm
- [x] Use cache so we can re-use the same base for multiple linters

# Testing instructions

- Locally run yarn setup:ci and yarn lint:ci
- Github actions will run the rest


theyokohamalife:
@deek87 Thank you for helping us out with this! It looks like there are some errors in the CI. Could you have a look at them?

deek87:
Yeah its the delete request in the teardown.

Ive added the permissions for actions write, see here: https://docs.github.com/en/rest/actions/cache#delete-github-actions-caches-for-a-repository-using-a-cache-key

Or we can remove that step in the tear down and the caches will be kept up to 10GB

.github/workflows/lint.yml
```diff
@@ -0,0 +1,110 @@
+# ESLint is a tool for identifying and reporting on patterns
```
ermish, line 1: can you change the name back to `eslint` instead of `lint`? We have other linters that will be running

ermish: unless we can run stylelint as well from this? I'm not sure it would be parallellized then 🤔

package.json
```diff
@@ -7,6 +7,8 @@
         "dev": "yarn && nuxi dev",
         "build": "nuxi generate",
         "lint": "eslint  . -c .eslintrc.js --ext .vue,.js,.ts,.json",
+        "lint:ci": "eslint . --config .eslintrc.js --ext .js,.jsx,.ts,.tsx,.vue --format @microsoft/eslint-formatter-sarif --output-file eslint-results.sarif",
+        "setup:ci": "nuxi prepare",
```
ermish, line 11: let's change this to `prepare:ci` to be consistent with other scripts.

ermish (changes requested):
Thanks @deek87 for the contribution! A few small comments

.github/workflows/lint.yml
deek87: yeah we can run stylelint from this as well,
thats why we cache it in setup, then the cache is reused for the lint jobs

theyokohamalife:
> @theyokohamalife Yeah its the delete request in the teardown.
> 
> Ive added the permissions for actions write, see here: https://docs.github.com/en/rest/actions/cache#delete-github-actions-caches-for-a-repository-using-a-cache-key
> 
> Or we can remove that step in the teardown and the caches will be kept up to 7 days / 10GB and size and then auto removed by github. https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy

@deek87 For now, just to keep it simple could we remove that step and let GitHub automatically remove it?

deek87:
That should have fixed the permissions now 😃 👍

ermish:
hmmm teardown is still failing 🤔

deek87:
[code]
GITHUB_TOKEN Permissions
  Actions: read
  Contents: read
  Metadata: read
  SecurityEvents: read
[/code]

For some reason it is not giving the correct permissions at all, I'll just remove the delete action for now

theyokohamalife (approved):
LGTM! Thank you! Will also need approval from @ermish

PR #104, opened 2022-11-01
localePinia

### LINKED ISSUE #59: Navbar - Create a new state store for language selection
# Requirements

- Create a new pinia store for selecting the language
- For MVP, the two options will be `English` and `Japanese`. These values can be strings for now.
- Create an action/mutation for updating this state.

nasiranebi914:
Part I of #59 

# What changed
Created a new state store for locale selection 

# Testing Instructions

[ required ]


nasiranebi914:
Having a commit issue.. Trying to fix

ann-kilzer:
Great! I'll review when I return to Tokyo Tonight 😃

stores/localeStore.js
```diff
@@ -0,0 +1,21 @@
+import { defineStore } from 'pinia'
+
+export const useLocaleStore = defineStore('locale', {
+    state: () => ({
+        locale: { lable: 'English', code: 'en' }
```
ann-kilzer, line 5:
```suggestion
        locale: { label: 'English', code: 'en' }
```

stores/localeStore.js
```diff
@@ -0,0 +1,21 @@
...
+    }),
+    getters: {
+        currentLocale: state => state.locale
+    },
+    actions: {
+        changeLocale() {
+            if (this.currentLocale.code === 'en') {
+                this.locale.lable = '日本語'
```
ann-kilzer, line 13: After updating to typescript, it would be better to make this more object oriented and switch the entire locale at once rather than updating every field within. E.g. perhaps you would create a "type", "const", "enum" perhaps even, to represent `en` and `ja` locales

ann-kilzer (commented):
Thanks for starting this. Can we make new files in typescript? It is a little more advanced but it has better typesafety and some nice features.

stores/localeStore.js
nasiranebi914: Thank you for the advice! I will work on it :)

nasiranebi914: Hi @ann-kilzer, I just made some changes to the file by changing it to TypeScript. Is this the correct way to do it? :)

stores/localeStore.ts
```diff
@@ -0,0 +1,30 @@
...
+        changeLocale() {
+            if (this.currentLocale.code === 'en') {
+                this.locale = localeEn
+            } else {
+                this.locale = localeJa
+            }
+        }
+    }
+})
+
```
ann-kilzer, line 29: Looks like the linter is failing because there are two blank lines at the end and there should only be one

stores/localeStore.ts
```diff
@@ -0,0 +1,28 @@
+import { defineStore } from 'pinia'
+
+const localeEn:{label: string, code: string} = {
```
ann-kilzer, line 3: As a future enhancement (in a later PR) let's consider making a proper type rather than an anonymous type

ann-kilzer (approved):
I'm going to approve this as it's been on a branch for awhile and it would be nice to bring it into `main.` There are a few follow-up steps I'll add in #59 

Also going forward, please always include testing instructions in your PRs 🙏🏻 

Thanks!!!

PR #135, opened 2022-12-31
Update default.md

ann-kilzer:
Some suggestions for our PR template. I'd like to get a sense of what we're looking for from contributors, and then copy this to the server repo. Let me know your thoughts!


ermish (approved):
nice additions! Noticed it didn't have that resolves recently

PR #136, opened 2022-12-31
feat: updated footer to match mockups

### LINKED ISSUE #134: Search Page - Update footer
# Requirements
---

- [ ] The current footer is missing styling and elements needed to match our current design. Update it!
- [ ] Add missing links. Create empty new pages for now privacy and terms pages. Can implement them in separate issues. 
- [ ] Ensure the footer is responsive and displays appropriately across portrait and landscape modes. (desktop and mobile)

---
# Dependencies

ermish:
Resolves #134 

# What changed
- [x] added new footer elements
- [x] styling matches prototype
- [x] add empty pages for terms, privacy pages.
- [x] add link to feedback form (created an initial one! probably should move to findadoc google email instead of personal)
- [x] add link to github for contributions.
- [ ] ensure footer scales on desktop mode

NOTE: decided to keep mobile view outside the scope as it will change the design. 

# Testing instructions
- look at the beautiful new footer on desktop and mobile view. Nod head and think, "hmmm...iii ne"


ermish:
@ann-kilzer @theyokohamalife ready for review! The current design from the mockup doesn't responsively scale properly to mobile, so we'll need a different design in a separate issue.

.eslintrc.js
```diff
@@ -133,7 +133,7 @@ module.exports = {
                 max: 4
             },
             multiline: {
-                max: 3
+                max: 4
```
ermish, line 136: was conflicting with linter

.stylelintrc
```diff
@@ -7,6 +7,7 @@
         "stylelint-order"
     ],
     "rules": {
+        "block-closing-brace-newline-before": "always-multi-line",
```
ermish, line 10: looked really weird without this rule

assets/css/tailwind.css
```diff
@@ -1,3 +1,10 @@
 @tailwind base;
 @tailwind components;
 @tailwind utilities;
+
+@layer base {
```
ermish, line 5: had two choices, either but base styles in here or in the json config. Figured its better to use this file for base styles.

components/Footer.vue
```diff
@@ -1,10 +1,44 @@
...
+        <div class="middle-footer section flex flex-1 align-middle justify-center">
+            <span class="self-center">Powered by</span>
+            <nav class="flex mx-2 self-center">
+                <NuxtLink to="https://www.netlify.com/" target="_blank" class="underline">
+                    Netlify
+                </NuxtLink>
+            </nav>
+        </div>
+        <div class="right-footer-section flex flex-0 flex-row gap-6 justify-end px-8 text-xs">
+            <div class="flex flex-col flex-wrap mx-2">
```
ermish, line 29: added this in. I think it would be valuable to have a link to our github.  We can change the design however, just wanted to get it in the code

pages/privacypolicy.vue
```diff
@@ -0,0 +1,39 @@
+<template>
+    <div class="h-full">
+        <div class="flex flex-col md:flex-col h-full">
+            <h1 class="h1">Privacy Policy</h1>
```
ermish, line 4: dummy data for the page

tailwind.config.js
```diff
@@ -15,10 +15,24 @@ module.exports = {
         screens: {
             md: { min: '800px' }
         },
-        extend: {},
         colors: {
             transparent: 'transparent',
             current: 'currentColor',
+            currentColor: '#ED6C5A',
+            orange: {
```
ermish, line 22: I added the orange that was in the mockup from Kei. We can change this to our future colors. 
Set it up so we can change it easily. We could even allow theming to different colors.

assets/css/tailwind.css
theyokohamalife: Nit: Might want to consider differentiating h3 and h4.

components/Footer.vue
```diff
@@ -1,10 +1,44 @@
 <template>
-    <div class="flex bg-teal-700 text-white">
-        <div class="flex-auto">
-            <a href="https://www.netlify.com/" target="_blank">Deployed on Netlify</a>
+    <div class="footer flex bg-orange-100 text-sm text-current font-semibold py-2">
+        <div class="left-footer-section flex flex-0 justify-start">
+            <NuxtLink to="/" class="self-center ml-8 mr-2">
+                © 2023 Find a doc JP
```
theyokohamalife, line 5: Nit: Let's make an issue to add a function so the year is always up to date?

Also, the official name is: `Find a Doc, Japan`

assets/css/tailwind.css
ermish: also these header styles are just temporary until we have some designs

components/Footer.vue
theyokohamalife, line 5:
```suggestion
                © 2023 Find a Doc, Japan
```

components/LeftNavbar.vue
```diff
@@ -0,0 +1,5 @@
+<template>
+    <div class="grid">
```
theyokohamalife, line 2: Nit: Is this grid class needed? Would like to remove it if not.

components/SearchResult.vue
```diff
@@ -0,0 +1,48 @@
+<template>
+    <div class="flex flex-col">
```
theyokohamalife, line 2: This `flex flex-col` class shouldn't be needed since that's the default orientation. If this is removed then all of the `flex flex-row` can just be changed to `flex`.

components/SearchResult.vue
```diff
@@ -0,0 +1,48 @@
...
+            <img alt="profile icon" class="profile-icon" src="@/assets/images/profileIcon.svg" width="25"
+                height="25" />
+            <span class="flex-1 w-4 font-bold pl-2"> {{ name }} </span>
+        </div>
+        <div class="result-details flex flex-row my-2">
+            <span class="pr-2"> {{ specialty }}</span>
+            <span class="self-center">·</span>
+            <span class="px-2"> {{ facilityName }}</span>
+        </div>
+        <div class="result-tags flex flex-row flex-wrap w-64 mb-2">
```
theyokohamalife, line 13: Nit: Won't the flex-wrap cause the text items to overflow? I think we should leave it out and allow them to wrap normally like they do by default.

components/SearchResult.vue
```diff
@@ -0,0 +1,48 @@
...
+            <div :key="index" v-for="(spokenLanguage, index) in spokenLanguages"
+                class="pl-2 pr-2 py-[1px] mr-2 border border-solid border-orange/40 rounded-2xl shadow text-sm">
```
theyokohamalife, line 15:
```suggestion
                class="pl-2 pr-2 py-[1px] mr-2 border-2 border-orange-40 rounded-2xl shadow text-sm">
```

theyokohamalife: Borders are solid by default. I would simplify them like this.

theyokohamalife (approved):
You have a lot of `flex flex-col` that can just be removed, and the `flex flex-row` simplified as just `flex`.  Those are refactoring changes though and not necessary to hold up the PR. Please commit the project's name correction in the footer and other than that it LGTM!

assets/css/tailwind.css
ermish: I just noticed my comment was added after yours when I refreshed 😅  
Let's update these later when Kei has css rules for our headers

components/Footer.vue
ermish: ah good idea! I just copied from the mockup XD

components/LeftNavbar.vue
ermish: I need to do more grid work on the page, but the main advantage is that it doesn't jump when you first load or refresh the page. For structural components, I'm wanting to use grid instead of flex.

For this one though, I think you're right and it's not doing anything yet XD

components/SearchResult.vue
ermish: it will cause the flex items to wrap instead of overflowing to the right. it's different than text-wrap. 

My thinking is that if we have more than 3 tags, we'll want them to start wrapping into a second row. 

Here's an example
https://developer.mozilla.org/en-US/docs/Web/CSS/flex-wrap

components/SearchResult.vue
ermish: good call 👍 will update

The `/` is for the opacity level, not the color grade. 
https://tailwindcss.com/docs/border-color#changing-the-opacity

components/SearchResult.vue
theyokohamalife: Ah, I mistakenly thought this div was the tag itself and not the container!

components/Footer.vue
ermish: updated

ermish: updated

assets/css/tailwind.css
ermish: updated

components/SearchResult.vue
ermish: nice catch! updated

ermish: honestly, I'm used to being overexplicit on flex classes. 
Good for me to trim out defaults more often 👍

components/SearchResult.vue
ermish: updated

ermish:
@theyokohamalife updated with your recommendations 💯  Merging now. If I missed anything, let me know and I fit in an iterative PR.

PR #139, opened 2023-01-02
Update default.md

ann-kilzer:
I got PR and issue templates mixed up... resolved makes no sense here. Also make some adjustments


ermish (approved):
LGTM! 
Merging it so I don't have to keep fixing it for new issues. Should be same since it's just the MD file.

PR #140, opened 2023-01-02
Upgrade Node

### LINKED ISSUE #125: Upgrade node to 18.12.1 LTS
https://nodejs.org/en/

ann-kilzer:
Resolves #125 

# What changed
- update to Node 18

# Testing instructions
Site should build in netlify (check the deploy log that it uses node 18)
GitHub Actions should pass


ann-kilzer:
I notice we have an env variable in netlify `NODE_VERSION`... I think it would be better to pass that in from the codebase. Thoughts @ermish ?

ermish:
> I notice we have an env variable in netlify `NODE_VERSION`... I think it would be better to pass that in from the codebase. Thoughts @ermish ?

I think that's a good idea and I believe I tried to get it to run based on that version. I tried mode version in package json and the nvmrc but I didn't pick them up so I did the end var. 

If it works now/in the future, let's go for it!

ermish (approved):
Nice! :shipit:

ann-kilzer:
Yeah looking at Netlify's docs, I think we can just pass in the .nvmrc. I will also run the command on netlify to install the required version of node via nvm.

ann-kilzer:
Netlify is making using node 18 difficult... will investigate more at a later time or someone else feel free to make suggestions

RageZBla (approved):

netlify.toml
```diff
@@ -0,0 +1,6 @@
+# https://docs.netlify.com/configure-builds/file-based-configuration/
```
ann-kilzer, line 1: I'm encouraging us to use the toml for non-secrets because it makes the configuration really clear to non-admins

ermish: Nice! I'm all for it 🙌

ermish (approved):
Awesome

Issue #168, opened 2023-01-26
Start Cypress Tests for frontend


ann-kilzer:
# Requirements

- [x] cypress tooling and docs in place
- [x] Skeleton tests
- [x] Finish the skeleton tests in `home.cy.ts`

Please break this work into small parts and share intermediate progress via PR for feedback.

# Dependencies

- None (Add any issue dependencies here)

ann-kilzer:
If you're new to cypress I recommend taking the free fundamentals course here: https://cypress.tips/courses
Before starting this

ann-kilzer:
I'm gonna write some skeleton tests that can be divided up by people later

ann-kilzer:
@Tch4lla This is ready for you to get started. Let us know if you have questions

PR #186, opened 2023-02-06
Cypress Setup

ann-kilzer:
Part of #168 

Ah, we already have Cypress running in CI 😆 but anyway now there's a command to run it all in one go

# What changed
- [x] Cypress documentation
- [x] https://github.com/bahmutov/start-server-and-test
- [x] non-GUI command

# Testing instructions

Try the new instructions in the README and confirm they make sense


package.json
```diff
@@ -13,7 +13,9 @@
         "start": "nuxi start",
         "test:pinia": "vitest",
         "test:pinia:coverage": "vitest run --coverage",
-        "test:e2e": "cypress open --e2e --browser chrome --port 9090"
+        "test:e2e": "cypress open --e2e --browser chrome --port 9090",
+        "test:e2e:run": "cypress run --e2e --browser chrome",
```
ermish, line 17: Should we just replace the other command with run or do we need both of them separate?

ermish (approved):
:shipit:

package.json
ann-kilzer: Replacing makes sense. In general I try not to replace scripts in case I break something, but it sounds like the only place to fix is the github action

ermish (approved):
🚢 it

PR #218, opened 2023-02-20
feat: generating types from gql api

### LINKED ISSUE #204: Connect the generator and create a script task
# Connect the generator and create a script task 

- [ ] Connect the type generator
- [ ] Create a script task
- [ ] Ensure it connects to the server for the latest file

# Dependencies

- #203

ermish:
Resolves #204 

# What changed

- [x] It works! We can now generate our typescript types based on the graphql api types! This = consistency 😎 
- [x] Connected to server to pull types
- [x] Fixed generate task
- [x] Added types file to eslint ignore so we don't have to lint fix a generated file  

# Testing instructions
- run `yarn generate` and it should work. the `typedefs/gqlTypes.ts` file will update with any changes

### DIFF
--- a/typedefs/gqlTypes.ts
+++ b/typedefs/gqlTypes.ts
@@ -1,0 +1,387 @@
+import { GraphQLResolveInfo } from 'graphql'
+export type Maybe<T> = T | null;
+export type InputMaybe<T> = Maybe<T>;
+export type Exact<T extends { [key: string]: unknown }> = { [K in keyof T]: T[K] };
+export type MakeOptional<T, K extends keyof T> = Omit<T, K> & { [SubKey in K]?: Maybe<T[SubKey]> };
+export type MakeMaybe<T, K extends keyof T> = Omit<T, K> & { [SubKey in K]: Maybe<T[SubKey]> };
+export type RequireFields<T, K extends keyof T> = Omit<T, K> & { [P in K]-?: NonNullable<T[P]> };
+/** All built-in and custom scalars, mapped to their actual values */
+export type Scalars = {
+  ID: string;
+  String: string;
+  Boolean: boolean;
+  Int: number;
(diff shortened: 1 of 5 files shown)


theyokohamalife (approved):
So awesome! Thank you!!

typedefs/gqlTypes.ts
```diff
@@ -0,0 +1,387 @@
+import { GraphQLResolveInfo } from 'graphql'
```
ann-kilzer, line 1: What do you think about not checking in generated code into the repo?

ann-kilzer (changes requested):

typesgeneratorconfig.ts
```diff
@@ -2,9 +2,9 @@ import { CodegenConfig } from '@graphql-codegen/cli'
 
 const config: CodegenConfig = {
     overwrite: true,
-    schema: './src/typeDefs/schema.graphql',
+    schema: 'https://seal-app-5oq9w.ondigitalocean.app/graphql',
```
ann-kilzer, line 5: Great 👍

ann-kilzer:
Wait... since the types depend on the server... we probably do have to check it in. Let's make a ticket for our CI/CD to confirm that the generated file matches the production DB. We don't want people hand-editing that file

ann-kilzer (approved):
I take it back... ship it

typedefs/gqlTypes.ts
ermish: we need references to it otherwise we wouldn't be able to build properly. 
I don't think requiring generation per person would bring any benefit. 🤔 
Basically, if we have an updated schema, someone would just run it again and check it in so everyone has it. Also, it would allow us to version and only update the client-side schema when we know it works.

PR #224, opened 2023-03-09
Cypress testing tch4lla

### LINKED ISSUE #168: Start Cypress Tests for frontend

Tch4lla:
Resolves #168 

# What changed

- Added all testing queries using cypress
- Added data-testid to multiple elements in search bar and footer
- "changed linebreak-style": 0 in eslintrc under rules in order to tackle the linting issue

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -66,7 +66,7 @@ module.exports = {
         ],
         'key-spacing': ['error', { beforeColon: false, afterColon: true }],
         'keyword-spacing': ['error', { before: true, after: true }],
-        'linebreak-style': ['error', 'unix'], // no carriage returns
+        'linebreak-style': 0, // no carriage returns
         'max-len': ['error', { code: 140 }], // be friendly to laptops
         'newline-after-var': 'error',
         'require-atomic-updates': 'off',
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -66,7 +66,7 @@ module.exports = {
         ],
         'key-spacing': ['error', { beforeColon: false, afterColon: true }],
         'keyword-spacing': ['error', { before: true, after: true }],
-        'linebreak-style': 0, // no carriage returns
+        'linebreak-style': ['error', 'unix'], // no carriage returns
         'max-len': ['error', { code: 140 }], // be friendly to laptops
         'newline-after-var': 'error',
         'require-atomic-updates': 'off',
--- a/components/Footer.vue
+++ b/components/Footer.vue
@@ -7,13 +7,15 @@
             </NuxtLink>
             <span class="self-center">·</span>
             <nav class="flex mx-2">
-                <NuxtLink to="/privacypolicy" class="self-center hover:text-primary-hover transition-colors">
+                <NuxtLink to="/privacypolicy" class="self-center hover:text-primary-hover transition-colors"
+                data-testid="privacy-link">
                     Privacy
                 </NuxtLink>
             </nav>
             <span class="self-center">·</span>
             <nav class="flex mx-2">
-                <NuxtLink to="/terms" class="self-center hover:text-primary-hover transition-colors">
+                <NuxtLink to="/terms" class="self-center hover:text-primary-hover transition-colors"
+                data-testid="terms-link">
                     Terms
                 </NuxtLink>
             </nav>
@@ -21,14 +23,16 @@
         <div class="middle-footer section flex flex-1 align-middle justify-center">
             <span class="self-center">Powered by</span>
             <nav class="flex mx-2 self-center">
-                <NuxtLink to="https://www.netlify.com/" target="_blank" class="underline hover:text-primary-hover transition-colors">
+                <NuxtLink to="https://www.netlify.com/" target="_blank" class="underline hover:text-primary-hover transition-colors"
+                data-testid="netlify-link">
                     Netlify
                 </NuxtLink>
             </nav>
         </div>
         <div class="right-footer-section flex flex-0 flex-row gap-6 justify-end px-8 text-xs">
             <div class="mx-2 hover:text-primary-hover transition-colors">
-                <NuxtLink to="https://github.com/ourjapanlife/findadoc-web/" target="_blank" class="flex flex-col flex-wrap ">
+                <NuxtLink to="https://github.com/ourjapanlife/findadoc-web/" target="_blank" class="flex flex-col flex-wrap "
+                data-testid="github-link">
                     <div>Contribute on</div>
                     <div class="self-center">
                         GitHub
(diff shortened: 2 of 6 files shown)


.eslintrc.js
```diff
@@ -66,7 +66,7 @@ module.exports = {
         ],
         'key-spacing': ['error', { beforeColon: false, afterColon: true }],
         'keyword-spacing': ['error', { before: true, after: true }],
-        'linebreak-style': ['error', 'unix'], // no carriage returns
+        'linebreak-style': 0, // no carriage returns
```
ann-kilzer, line 69: Please don't change this, we do not want carriage returns. Make sure your editor is correctly configured with editorConfig

components/Footer.vue
```diff
@@ -1,34 +1,38 @@
 <template>
     <div class="footer flex border-t-4 border-secondary-bg/40 bg-secondary-bg text-sm
-    text-primary-text font-semibold py-2">
+        text-primary-text font-semibold py-2">
```
ann-kilzer, line 3: Seems like an unrelated change?

ann-kilzer (changes requested):
Thanks for starting this. See comments

.eslintrc.js
Tch4lla: I understand. In regards to this change, I've changed my end of line character to \n so that it is for the unix type, and the linter will return 1100+ errors in regards to the spacing. yarn lint --fix does not autofix anything, neither does the selction 'fix all line space errors'. 

Would you be able to offer any suggestions?

components/Footer.vue
Tch4lla: This probably was a lint format fix.

.eslintrc.js
Tch4lla: linter

I just changed it back to that error unix property and ran yarn lint --fix command

Tch4lla: I manually fixed all the errors. The linting should pass at least. Still, I'm not sure why the auto fix functions don't work...

components/Footer.vue
ann-kilzer: As a best practice, we should be separating linting fixes from logical changes. The files have been linted before and auto-checked as part of CI/CD. Unrelated changes distract from the core purpose of this PR: cypress testing

If you confirm that there is an error with the linter (can test by running against `main`), feel free to open a separate PR to address the issue.

cypress/e2e/home.cy.ts
```diff
@@ -10,34 +13,76 @@ describe('Visits the home page', () => {
...
+    it('can select "Japanese" from the language bar', () => {
+        cy.get('[data-testid="search-bar-language"]').should('be.visible')
+
+        cy.get('[data-testid="search-bar-language"]').select('Japanese')
+
+        cy.get('[data-testid="search-bar-language"]').should('have.value', 'Japanese')
+    })
 
+    it('shows doctors nearby', () => {
+        cy.contains('Doctors Nearby').should('be.visible')
```
ann-kilzer, line 50: Can we extend this one to ensure that the expected content is in the component?

cypress/e2e/home.cy.ts
```diff
@@ -10,34 +13,76 @@ describe('Visits the home page', () => {
...
+        })
 
         // verify Netlify citation
+        it('navigates to the netlify page', () => {
+            cy.get('[data-testid="netlify-link"]').should('be.visible')
+            cy.get('[data-testid="netlify-link"]').should('have.attr', 'href', netlify)
+        })
 
         // privacy
+        it('naviagtes to private policy', () => {
```
ann-kilzer, line 75: looking good 👍🏻

components/TopNav.vue
```diff
@@ -10,7 +10,7 @@
                 </svg>
                 <div class="title-text flex flex-col" data-testid="logo">
                     <div class="text-lg text-primary group-hover:text-primary-hover">
-                        Find a Doc 
+                        Find a Doc
```
ann-kilzer, line 13: 👀 looks like this one WAS unlinted 🤔

cypress/e2e/home.cy.ts
```diff
@@ -10,34 +13,76 @@ describe('Visits the home page', () => {
...
 
-    it.skip('displays the footer', () => {
+    describe('Checks footer links', () => {
+        it('navigates to github', () => {
         // TODO
 
-        // verify link to GitHub
+            // verify link to GitHub
+            cy.get('[data-testid="github-link"]').should('be.visible')
+            cy.get('[data-testid="github-link"]').should('have.attr', 'href', 'https://github.com/ourjapanlife/findadoc-web/')
```
ann-kilzer, line 59: Should we make a const for the github url, for consistency?

cypress/e2e/home.cy.ts
```diff
@@ -10,34 +13,76 @@ describe('Visits the home page', () => {
...
+    })
+
     it('allows setting search fields', () => {
         cy.get('[data-testid="search-button"]').should('be.visible')
         // TODO
     })
 
-    it.skip('selects a language', () => {
+    it('can select "select a language"', () => {
+        cy.get('[data-testid="search-bar-language"]').trigger('click')
```
ann-kilzer, line 29: Let's add some assertions to all the tests (or skip them if we don't have enough of the homepage working to finish)

ann-kilzer (commented):
Took a closer look at the tests and they're looking on the right track

.eslintrc.js
ann-kilzer: On our readme we note that everyone should be using the https://editorconfig.org/ tool, and if you use VSCode, you will need to install a special plugin. This file will set up expected line endings and formatting

🤔 Can we make this important setup step clearer somehow? Maybe we need to highlight it in CONTRIBUTING.md

Git on windows sometimes does sneaky things, replacing `\n` with `CRLF` on checkout, and converting back on upload.

cypress/e2e/home.cy.ts
Tch4lla: I'll take care of that

cypress/e2e/home.cy.ts
Tch4lla: I understand. I'll skip this one. I guess in the following tests if I can select the languages, then this test by default should pass. I'll keep that in mind as I go forward

components/TopNav.vue
Tch4lla: This might be the lf vs crlf line ending problem I was taking about?

.eslintrc.js
Tch4lla: Hmm, I have the editorconfig extension installed on my vs code and I have the editorconfig file pulled. That is interesting about the \n for git on windows.

components/Footer.vue
Tch4lla: Thank you for that. Previously when I was working on the other PR for the results store, I was always running into the linting issue, and was advised to run yarn lint fix, so I assumed that what I should always do before a push. I'll keep that in mind for the future

ann-kilzer (approved):
Going to remove my blocking "changes requested" as the line ending has been fixed. Tests are on the right track. There are a few suggested changes I've already left, but this is a good step forward. Thank you!

PR #267, opened 2023-07-17
Use Pinia State for Modals

theyokohamalife:
Resolves #212 

# What changed
1. @pinia/nuxt was added to better facilitate using Pinia to manage the state for components.
2. modalStore.ts was created to manage the Modal.vue component's state
3.  `.env` was added to .gitignore.

# Testing instructions
1. Rename app.vue to app-backup.vue and  app-copy.vue to app.vue in order to see the main page.
4. Change `isOpen: false` to `isOpen: true` in _modalStore.ts_ so the modal opens by default when you refresh the page.
Screenshot 2023-07-17 at 7 17 04 PM

5. Start the dev server and confirm that clicking the X on the modal closes it on all Chrome, Firefox, and Safari.

PR #269, opened 2023-07-17
Add Google Maps to MapContainer

theyokohamalife:
This PR follows #267 

# What changed
1. Adds `runtimeConfig` to nuxt.config.js so the Google Maps API Key is loaded by Nuxt.
2. Adds the `vue3-google-map` package to implement Google Maps more easily.
3. Adds `<GoogleMap>` to the MapContainer to render Google Maps in the component.
4. Adds blue pin icon for map. Other icons can be added in later PRs.
5. Creates a locationStore as a temporary measure to handle the state for the location the user selects. Let's discuss how this might be better implemented in a future PR.
6. Updates the `SearchResultDetails` and `SearchResultsList` components to use computed values from Pinia.

# Testing instructions
Same testing instructions as #267. Only difference is that you should now see the mockData populate the modal and search results list.

# Screenshot
Screenshot 2023-07-17 at 8 31 18 PM


.gitignore
```diff
@@ -3,7 +3,10 @@
 .output
 dist
 
-#package management and yarn3
+# Environment files
+.env*
```
ermish, line 7: see other PR comment

components/MainContentContainer.vue
```diff
@@ -4,9 +4,13 @@ import SearchResultDetails from './SearchResultDetails.vue';
 <template>
     <div class="flex flex-1 bg-secondary-bg/20 hover:shadow-inner hover:shadow-secondary-bg/90">
         <!-- <WelcomeSection /> -->
-        <Modal class="min-h-1/2 ml-8 mt-12">
+        <Modal v-show="useModalStore().$state.isOpen" class="min-h-1/2 ml-8 mt-12">
```
ermish, line 7: see other PR comments first

components/SearchResultDetails.vue
```diff
@@ -58,27 +58,17 @@
...
-
-const name = ref(dummyData.name)
-const specialty = ref(dummyData.specialty)
-const facilityName = ref(dummyData.facilityName)
-const spokenLanguages = ref(dummyData.spokenLanguages)
-const address = ref(dummyData.address)
-const hours = ref(dummyData.hours)
-const website = ref(dummyData.website)
-const phone = ref(dummyData.phone)
+const healthcareProfessionalName = computed(() => useLocationStore().$state.healthcareProfessionalName)
```
ermish, line 64: instantiate the store here once as well

stores/locationStore.ts
```diff
@@ -0,0 +1,36 @@
...
+        facilityName: null,
+        spokenLanguages: null,
+        address: null,
+        hours: null,
+        website: null,
+        phone: null,
+        position: { lat: 35.699615, lng: 139.545596 }
+    }),
+    actions: {
+        setLocationDetails(location: { healthcareProfessionalName: string;
```
ermish, line 18: would it make sense to have this be a typescript type?

ermish (changes requested):
looks great! nice work ⭐

components/MapContainer.vue
```diff
@@ -1,5 +1,55 @@
 <template>
-    <div>
-        <img src="@/assets/images/map-sample.png" class="w-full h-full opacity-80" data-testid="map-of-japan" />
-    </div>
+    <GoogleMap ref="mapRef" :api-key="runtimeConfig.public.googleMapsApiKey"
+        map-id="153d718018a2577e"
```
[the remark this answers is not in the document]
theyokohamalife: It's a unique key for our Google Maps API. It's not something I want to publish to the repo.

stores/locationStore.ts
theyokohamalife: Like I mentioned in the PR description, this file is probably temporary and will be rewritten as we're connecting this to the backend. Feel free to make changes!

components/MainContentContainer.vue
theyokohamalife: @ermish These will be updated when I rebase off #267

components/MapContainer.vue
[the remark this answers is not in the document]
ermish: I meant the `map-id`. Is that something that also should not be saved here?

ermish (approved):
LGTM when the changes are done ✅

components/MapContainer.vue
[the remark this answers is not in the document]
theyokohamalife: `map-id` is a unique key that won't be changed. However, it's not private. It could be stored in an env file or as a runtime variable instead.

PR #301, opened 2023-12-05
chore: update localization and credentials

theyokohamalife:
1. Puts footer text into localizable strings
2. Updates the runtime config variables


ermish (approved):
Ship it

PR #309, opened 2023-12-07
Fix address line in ja_JP

theyokohamalife:
Fix the query to include the 2nd line of the address for the facility.


Anissa3005 (approved):
LGTM

PR #318, opened 2023-12-08
fix: make top nav responsive

theyokohamalife:
# What changed
Adds a breakpoint for md size screens so only the hamburger menu is shown on smaller screens. Further improvements will need to be made to display the search filters on mobile, and have the overall UI look and flow as depicted on [Figma](https://www.figma.com/file/QXQxlw4jNFIhMxkKU7bbwq/Find-a-doc%2C-Japan?type=design&node-id=1207-9553&mode=design&t=yirhFIMLVvD6Wkwb-0).

This is just a 🩹 and a lot more work on the top nav will be needed!

# Screenshots

## Before
Screenshot 2023-12-09 at 12 46 40 AM

## After
Screenshot 2023-12-09 at 12 45 51 AM

### DIFF
--- a/components/TopNav.vue
+++ b/components/TopNav.vue
@@ -1,44 +1,62 @@
 <template>
-  <div
-    class="p-4 pl-8 border border-secondary-bg/20 flex justify-between items-center bg-gradient-to-t from-secondary-bg/30 via-primary-bg to-primary-bg"
-  >
-    <HamburgerComponent />
-    <div class="font-semibold text-xl group transition-colors items-start">
-      <NuxtLink class="flex" to="/">
-        <svg
-          role="img"
-          title="site icon"
-          class="mr-1 w-10 h-10 flex-shrink-0 align-middle fill-primary group-hover:fill-primary-hover"
-        >
-          <use xlink:href="../assets/images/site-logo.svg#site-logo-svg" />
-        </svg>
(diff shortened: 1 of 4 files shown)


stores/menuStore.ts
```diff
@@ -0,0 +1,22 @@
...
+export const useMenuStore = defineStore('menuStore', () => {
+    const menuOpen = ref(false)
+    const menuItems = {
+        "home": { "displayText": "topNav.home", "route": "/" },
+        "about": { "displayText": "topNav.about", "route": "/about" },
+        "submit": { "displayText": "topNav.submit", "route": "/submit" },
+    }
+
+    function toggleMenu() {
+        console.log("menu= ", menuOpen.value)
```
ermish, line 13: should remove this log. we can check vue plugins to track state. 
for future reference, we should only keep logs for data fetching/submission and warnings/errors

components/TopNav.vue
```diff
@@ -1,44 +1,62 @@
...
+                        data-testid="logo"
+                    >
+                        <div
+                            class="text-lg text-primary group-hover:text-primary-hover"
+                        >
+                            Find a Doc
+                        </div>
+                        <div
+                            class="text-sm text-primary leading-none group-hover:text-primary-hover"
+                        >
```
ermish, line 31: these closing tags on their own line is strange (to me at least) Should we consider changing whatever rule is doing this?

PR #325, opened 2023-12-11
changed from apolloclient to graphql-request

### LINKED ISSUE #348: Display of duplicate locations
When you want to select location there are duplicate locations

Image

# Requirements

- [ ] Filter out he duplicate location in filter bar

# Dependencies

### LINKED ISSUE #366: Doctors nearby not rendering on mobile

### LINKED ISSUE #369: Change from apollojs client to graphql-request
Apollo graphql client has been a pain for us. Here's a few reasons: 
- There's no a clear way to trigger fetching data. It tries to work through ref variables that, when changed, trigger a data fetch. This causes a problem for us is that 1) the trigger for fetching is based on the user pressing search when no variables might change 2) We want to search after the user has changed all the filters, not a search for each one. 
- the client is heavier than we need. (we don't need heavy caching abilities for example)
- the "previously fetched" cache problem was causing our search results to only update when a new query was ran and ignored previously ran queries. 

Let's try switching to a simpler querying framework! 
[`graphql-request` ](https://www.npmjs.com/package/graphql-request) seems to be simple, fast, and actively maintained. 

# Requirements

- [ ] Remove apollo client
- [ ] Add graphql-request
- [ ] Rework our state stores to support the new fetching format.

ermish:
Resolves #369
Resolves #348 
Resolves #366 

# What changed

- changed from apolloclient to graphql-request (see #369 for more details)
- Fixed the state stores to use the new fetching format
- Tested new data fetching design

# Testing instructions
- Tests aren't currently running (need to fix ASAP), so manually verify the page looks good and data is fetching properly


Anissa3005 (approved):
As a Junior I approve 👌

theyokohamalife (changes requested):
Netlify fails to build with these changes. It looks like we're making a bad request to the API: 
Screenshot 2024-02-18 at 4 44 21 PM

NabbeunNabi:
The checks pass now. It is not resolving the duplicate locations still from what I see but I fixed the build problem I believe. I can remove those semicolons as well if they aren't supposed to be there. The problem was the yarn.lock needed to be rebuilt since since it was throwing errors for toUpperCase, and then the endpoint wasn't correct in the `utils/graphql.ts`

NabbeunNabi:
Just as a question. Does it resolve the bottom two issues as well or just the original one?

NabbeunNabi:
I do not think this currently resolves the duplicate search results. I do not know what the best solution would be but I think if there is a script that creates a lookup like  `{ city: count}` as the key value pair we could dynamically render it using `v-for` then later. But I do not know if we need this extra function in the script or if there is a better solution. Should it be solved on the frontend/backend/in the store?

ermish:
> I do not think this currently resolves the duplicate search results. I do not know what the best solution would be but I think if there is a script that creates a lookup like `{ city: count}` as the key value pair we could dynamically render it using `v-for` then later. But I do not know if we need this extra function in the script or if there is a better solution. Should it be solved on the frontend/backend/in the store?

that's a good question! we could solve in either the front end or backend. For now, I think front-end will be easier and should be fine for the result quantity. If it gets too heavy, we can shift to the backend.

PR #342, opened 2023-12-16
Tch4lla/mobile map display

Tch4lla:
# What changed
The left nav bar now extends to the width of the screen 
The map would be dsplayed under 

before
image

after
image

# Testing instructions
Change screen width to 640px


ermish:
@Tch4lla  I think this is outdated at this point
Let's work together on porting over some of your mobile styling effort to the current state 👍

Issue #347, opened 2023-12-20
Map not visible on mobile


Anissa3005:
When we view the application with a smartphone, the map is not visible anymore and we only see the search result. I think we need something similar like the google map mobile app. When we click on tag it open on full screen the doctor. And have a different page/modal/slide for all search results? (something we should discuss)

![Image](https://github.com/ourjapanlife/findadoc-web/assets/114712265/d91f5699-8002-44f5-b100-02eacac7ab5d)

# Requirements

- [x] Create a responsive search results and map for mobile

# Dependencies

- None (Add any issue dependencies here)

PR #365, opened 2024-01-24
Make the map visible again on desktop

Anissa3005:
Part of #347 

# What changed

- I have refactored the Mapcontainer component. In Vue3 we can add the `setup` in `<script>`. Since we are doing this in the other components it seemed better for consistency.
- To make the the map visible in desktop, I have changed some tailwind elements. During testing and reviewing I have notice that some elements were  unnecessary and have removed them.


theyokohamalife (approved):
Looks good!

PR #367, opened 2024-02-03
feat: Add email address to SearchResultDetails

### LINKED ISSUE #336: Healthcare Professional Email not displayed
# Problem
We currently store the email addresses for healthcare professionals in the database, but they aren't displayed anywhere on the site.

# Success
The SearchResultDetails component displays the email address for a healthcare professional if one is available.

evan-desu:
Resolves #336 

Previously a facility's email address was not displayed in the `SearchDetailsResult`. This renders the email.

# What changed
1. Implemented logic to conditionally display the facility's email address.
2. Added an SVG icon for email addresses.
3. Email addresses with a value of `none` or `email@email.com` are filtered out with a `v-if` statement. (See the images below. I can remove this if necessary.)

Screenshot 2024-02-03 at 17 22 35

Screenshot 2024-02-03 at 17 21 44

# Testing instructions
Open the app, click on a doc and check out their email!

### DIFF
--- a/components/SearchResultDetails.vue
+++ b/components/SearchResultDetails.vue
@@ -59,6 +59,13 @@
                     </svg>
                     {{ phone }}
                 </div>
+                <div class="email flex my-4" v-if="email !== 'none' && email !== 'email@email.com'">
+                    <svg role="img" alt="Facility Banner Image" title="banner image"
+                        class="banner-icon w-6 h-6 stroke-primary mr-2 self-center">
+                        <use xlink:href="../assets/images/email-icon.svg#email-icon-svg" />
+                    </svg>
+                    {{ email }}
+                </div>
             </div>
         </div>
     </div>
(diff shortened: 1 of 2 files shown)


components/SearchResultDetails.vue
```diff
@@ -59,6 +59,13 @@
                     </svg>
                     {{ phone }}
                 </div>
+                <div class="email flex my-4" v-if="email !== 'none' && email !== 'email@email.com'">
```
theyokohamalife, line 62: Just curious, where are `none` and `email@email.com` coming from? Are these placeholders? If so, it's better to use variables instead of hardcoded strings in case these can change in the future.

theyokohamalife (changes requested):
Looking good so far! Left a question regarding the conditionals you have. Also just a nit, but make sure your PR titles and commits are in the present tense instead of past for good practice. 🙏🏾

components/SearchResultDetails.vue
evan-desu: I've added an array of the placeholders. That way if we find more, we can just add them to the list.

theyokohamalife:
https://github.com/all-contributors please add @evan-desu  for code

theyokohamalife:
@all-contributors please add @evan-desu for code

PR #380, opened 2024-02-19
Update Cypress Home Tests

Tch4lla:
Resolves  #224 

# What changed
The page would reload after each test. when going from "displays the logo" to 'allows setting search fields', the page wasn't loaded in the DOM, making the test fail when trying to locate the search bar. 

I added "testIsolation: false" to the cypress config file in order to stop the page from reloading after every test. 

I also ran into the issue where the tests would run, but the page would not have finished loading, so in turn they would fail. I added a wait to the test in order to make sure all elements were loaded. 
image

# Testing instructions
in the command line run 
`yarn run test:e2e`


Tch4lla:
I skipped the testing for the map as I don't believe I have access to the Google map API

theyokohamalife (approved):
Awesome work fixing all the tests!

PR #393, opened 2024-02-20
fix: add test for modal store to fix failing vitest workflow

theyokohamalife:
Prior to this change the Vitest/Pinia workflow was failing because there weren't any tests to run. This change introduces a test for the workflow to execute and no longer fail.


Anissa3005 (approved):
LGTM!

PR #406, opened 2024-03-04
feat: add photos, titles, and names on about us

### LINKED ISSUE #295: Create about us page
# Requirements

- [ ] Update the page
- [ ] Add in all the board members
- [ ] Use Kei's design in [Figma](https://www.figma.com/file/QXQxlw4jNFIhMxkKU7bbwq/Find-a-doc%2C-Japan?type=design&node-id=704-12700&mode=design&t=BbqUi0FwYYnAvpqO-0)

NabbeunNabi:
Resolves #295 

## 🔧 What changed
-Before:  There were not photos or linked for the people on the about us page. The heart icon also was not visible and no links to urls for members were set up correctly.  
-After: Pictures, titles, and names, with links based on what I could find were implemented. It is mobile responsive and after lots of spacing consideration the style is set to not have a lot of whitespace. Not everything is lined up perfectly on mobile but to do that requires strange whitespace that doesn't look aesthetically appealing. The `app.vue` file style was changed from `h-screen` to `h-full`. To fix the map showing up on the main page the style on the div containing the `MapContainer.vue` in `MainContentContainer.vue` was changed from `h-full` to `h-screen`

## 🧪 Testing instructions
-run `yarn test:e2e` as tests have been added to make sure different html elements of the page contain the text desired

## 📸 Screenshots

-   ### Before
image
image

-   ### After
image
image


member_directory/members.json
```diff
@@ -0,0 +1,103 @@
+{
+    "members": [
+        {
+            "avatarImg": "",
```
theyokohamalife, line 4: The avatar images aren't rendering, because the value for `avatarImg` is an empty string for each object in the `members` array.

You're using the value for `avatarImg` in the for loop here:

```ts
<img
  :src="member.avatarImg"
  class="member-image object-cover rounded-full w- 
  16 h-16"
/>
```

The problem is that the function you wrote to map the paths for the images doesn't execute on mount. I think we could simplify things by just putting the full paths in the json since it's a separate file.

That will clean up the vue file even more.

Screenshot 2024-03-12 at 9 10 08 PM

theyokohamalife (changes requested):
This is looking really good! Left a suggestion.

NabbeunNabi:
I removed the function and found if you put in the json file `/_nuxt/` before the pathname it can dynamically render

assets/images/heart-plus.svg
ermish, line 1: why did we move this a png instead of svg? it would be better to keep this as is

cypress/e2e/home.cy.ts
```diff
@@ -92,4 +92,28 @@ describe('Visits the home page', () => {
...
+
+    describe('Visits the about page', () => {
+        before(() => {
+            cy.visit('/about')
+        })
+    
+        it('Displays the main content', () => {
+            cy.get('[data-testid="about-heading"]').should('be.visible').contains('About Find a Doc, Japan')
+            cy.get('[data-testid="about-subheading"]').should('be.visible').contains('Connecting people in Japan to the healthcare services they need, in the languages they need.')
+            cy.get('[data-testid="about-paragraph1"]').should('be.visible').contains('At Find a Doc, Japan, our mission is to promote a healthier and more inclusive society. We offer quality health information for free in many languages for everyone, including foreigners and medical professionals. We aim for a Japan where all people, regardless of nationality or language, can confidently navigate the healthcare system, find the right doctors and get the healthcare they need.')
```
ermish, line 104: we should avoid putting actual text to compare for a few reasons:
- text may change over time, but doesn't mean the test should fail or the page is invalid. 
- this doesn't account for viewing in a different language

Two options you can do is 1) just test that the element exists and has content or 2) compare content against the text in the language json file.

pages/about.vue
```diff
@@ -1,102 +1,132 @@
...
+            <p data-testid="about-subheading" class="mb-12 text-2xl">
+                {{ $t("about.subheading") }}
+            </p>
+            <p
+                data-testid="about-paragraph1"
+                class="mb-6 text-md text-primary-text-muted"
+            >
+                {{ $t("about.paragraph1") }}
+            </p>
+            <p
```
ermish, line 21: this formatting seems a bit strange, is this after running the linter? 
It could be a conflict if you have `prettier` running as well

pages/about.vue
```diff
@@ -1,102 +1,132 @@
...
+                id="members"
+                class="flex flex-wrap justify-between items-start w-4/5 mx-4 md:flex-wrap mx-32"
+            >
+                <div
+                    data-testid="member"
+                    class="members-list flex flex-col justify-center p-4 md:flex-row basis-1/3"
+                    :key="index"
+                    v-for="(member, index) in data.members"
+                >
+                    <div class="member flex flex-col items-center text-center">
```
ermish, line 67: nit: you could make a separate memberComponent here

NabbeunNabi: Are you referring to the whole container or it should loop through a component that is a card that takes the `member` and `index` as props?

assets/images/heart-plus.svg
NabbeunNabi: The `svg` file wasn't loading. LaShawn and I paired it and couldn't figure out why this is the only one that wouldn't load on the page. So as a quick fix we went with PNG and made an issue to later make it an `svg` file

pages/about.vue
```diff
@@ -1,102 +1,132 @@
...
+        </div>
+        <div
+            data-testid="members-container"
+            id="members-container"
+            class="flex justify-center items-center mx-4"
+        >
+            <div
+                data-testid="members"
+                id="members"
+                class="flex flex-wrap justify-between items-start w-4/5 mx-4 md:flex-wrap mx-32"
```
ermish, line 59: nit: we could use a little more whitespace at the bottom of the page after the members

```suggestion
                class="flex flex-wrap justify-between items-start w-4/5 mx-4 mb-8 md:flex-wrap mx-32"
```

ermish, line 59: you have two `mx`'s listed here

ermish (changes requested):
added a few comments!

pages/about.vue
NabbeunNabi: The first mx is on small resolutions per tailwind and the second one is when it moves to `md` resolution the `mx` changes to update the styling. Is there a better way to write this?

member_directory/members.json
NabbeunNabi: I changed the files to be in the json file and got rid of the function as suggested

pages/about.vue
ermish: the latter where you can have a member card. that you loop through. 
You can check out `SearchResultsList` and `SearchResultListItem` as an example 👍

member_directory/members.json
NabbeunNabi: Comment below describes what I did to fix it

assets/images/heart-plus.svg
ermish: 👍 I can live with that

theyokohamalife:
> So I do not know why it isn't building anymore. I didn't change paths. It all runs locally. Without any problems. Can generate the images. All I did was add the component like asked and made the changes.

You can click on the "Details" button for a test to see the log and the reason why it's failing. For example, if you look at the Cypress tests in the CI you'll see the following:

Screenshot 2024-03-15 at 9 41 35 AM

Based on the error written in yellow, we can see that it's not able to resolve the file path.

NabbeunNabi:
> > So I do not know why it isn't building anymore. I didn't change paths. It all runs locally. Without any problems. Can generate the images. All I did was add the component like asked and made the changes.
> 
> You can click on the "Details" button for a test to see the log and the reason why it's failing. For example, if you look at the Cypress tests in the CI you'll see the following:
> 
> Screenshot 2024-03-15 at 9 41 35 AM
> Based on the error written in yellow, we can see that it's not able to resolve the file path. You may want to try using `~/assets/` so it knows that the assets directory is in the root, instead of using `../assets/`.

I do not think I communicated my dilemma too well as it was late. I was able to locate the problem with the test that you showed me and see the path wasn't resolving. I am not sure as to why it doesn't resolve. Relative to the changes before, the path is the same just located in a different file, but where the file is in a different folder. Before it was in `paths`, and now it is in` components` I tried to use all these solutions on build. `@/, ~/, /_nuxt/`, but none of them work. I tried to have it as a hardcoded path instead of variables to see if there was an issue there. Locally it runs and I am able to see the `svg` field.

theyokohamalife:
> I do not think I communicated my dilemma too well as it was late. I was able to locate the problem with the test that you showed me and see the path wasn't resolving. I am not sure as to why it doesn't resolve. Relative to the changes before, the path is the same just located in a different file, but where the file is in a different folder. Before it was in `paths`, and now it is in` components` I tried to use all these solutions on build. `@/, ~/, /_nuxt/`, but none of them work. I tried to have it as a hardcoded path instead of variables to see if there was an issue there. Locally it runs and I am able to see the `svg` field.

I just checked again and noticed there isn't a file called `/assets/images/social-linkedIn.svg` in that folder. You have a typo causing it to fail.

NabbeunNabi:
I made the changes requested. I am not quite sure again as to why the socket security is going off. I have read the details. But it is not 100% clear to me as to what changed. I did commit a `yarn.lock` file. But that was built based off of what was already in` main`.

theyokohamalife:
@NabbeunNabi Please make sure to squash your fixup commits between reviews. It helps us to see only what changed since the last time we checked the PR. Also, you probably shouldn't need to commit the `yarn.lock` changes since you aren't making any dependency changes. After you remove that and squash your fixup commits I think this will be ready to go.

NabbeunNabi:
> @NabbeunNabi Please make sure to squash your fixup commits between reviews. It helps us to see only what changed since the last time we checked the PR. Also, you probably shouldn't need to commit the `yarn.lock` changes since you aren't making any dependency changes. After you remove that and squash your fixup commits I think this will be ready to go.

Done and Done. Thank you

theyokohamalife (approved):
LGTM ✨  Nice work!

theyokohamalife:
@ermish Will also have to approve it before it can be merged.

ermish (approved):
looks great! nice updates! and all green CI passing 🙌

PR #428, opened 2024-03-19
Debug undefined name when name does not exist in locale language.

### LINKED ISSUE #368: [Bug] Professionals with Japanese names and no english names show as undefined in details
Professionals with a Japanese name and no english name show as undefined in details when English is the selected page language. 

# Requirements

- [ ] Fix it!

Sample
image

Anissa3005:
Resolves #368 

## 📝 Description

Previously, if a professional's profile only contained a Japanese name (and vice versa) and the user switched the language locale to anything other than Japanese, the name would appear empty or undefined. Now it also shows a name when toggle to Swahili.

## 🔧 What changed

1. I have refactor the conditional statement, by using a switch case.

## 🧪 Testing instructions

Currently we actually don't have a professional that misses a English name or Japanese name. But if you want to test it please change the following  variables and toggle the locale:
```
    const englishFullName = undefined
    const japaneseFullName = `${japaneseName?.lastName} ${japaneseName?.firstName}`
```

### DIFF
--- a/components/SearchResultDetails.vue
+++ b/components/SearchResultDetails.vue
@@ -94,11 +94,21 @@ const healthcareProfessionalName = computed(() => {
         resultsStore.$state.activeResult?.professional.names.find(
             (n) => n.locale === Locale.JaJp
         )
+  
     const englishFullName = `${englishName?.firstName} ${englishName?.lastName}`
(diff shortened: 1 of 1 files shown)


theyokohamalife (approved):
LGTM ✨

PR #429, opened 2024-03-19
fix: change image from png to svg

### LINKED ISSUE #407: Convert PNG to SVG on About us Page
# Requirements

Currently, when using the svg from [figma](https://www.figma.com/file/QXQxlw4jNFIhMxkKU7bbwq/Find-a-doc%2C-Japan?type=design&node-id=704-12700&mode=design&t=EUnLF8LZ9sr77XjX-0 ) the image doesn't render properly, so we resorted to a png file for now, but in the future would like it to be an svg file

- [ ] Change the `heart-plus.png` file to an svg file that is rendered in `about.vue`

# Dependencies

- It is a follow up to PR #406

NabbeunNabi:
import the svg into the aboutus page and use it as the source of the image like the other svg being used in the aboutus page and MemberComponent.vue

Resolves #407 

## 🔧 What changed
I changed the image from a `png` to `svg`. I imported it in order to use the svg since the `xlink` wouldn't work correctly

## 🧪 Testing instructions
None

## 📸 Screenshots
It is the same as before. Just different type of image

-   ### Before
image

-   ### After
image


theyokohamalife (approved):
LGTM 💟

PR #435, opened 2024-03-21
chore: remove console logs and alerts

theyokohamalife:
🧹 Doing a bit of cleaning up.  ✨ 

I noticed that the console is overrun with data that we're already rendering in the UI. I removed any that weren't logging errors. Eventually, we'd like to use a service like [Datadog](https://www.datadoghq.com/) so we can remove those as well.

I also removed the alerts, because not only are they bad for the user experience, but they could also cause our site to be penalized by Google if they pop up too often.

### DIFF
--- a/stores/searchResultsStore.ts
+++ b/stores/searchResultsStore.ts
@@ -88,15 +88,12 @@ async function queryProfessionals(searchSpecialty?: Specialty, searchLanguage?: 
             } satisfies HealthcareProfessionalSearchFilters
         }
 
-        console.log('searching professionals')
         const response = await gqlClient.request<{ healthcareProfessionals: HealthcareProfessional[] }>(searchProfessionalsQuery, searchProfessionalsData)
-        console.log(`Fetched professionals: ${JSON.stringify(response)}`)
 
         const professionalsSearchResult = (response?.healthcareProfessionals ?? []) as HealthcareProfessional[]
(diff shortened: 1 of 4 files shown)


stores/searchResultsStore.ts
```diff
@@ -88,15 +88,12 @@ async function queryProfessionals(searchSpecialty?: Specialty, searchLanguage?:
...
 
-        console.log('searching professionals')
         const response = await gqlClient.request<{ healthcareProfessionals: HealthcareProfessional[] }>(searchProfessionalsQuery, searchProfessionalsData)
-        console.log(`Fetched professionals: ${JSON.stringify(response)}`)
 
         const professionalsSearchResult = (response?.healthcareProfessionals ?? []) as HealthcareProfessional[]
         return professionalsSearchResult
     } catch (error) {
         console.log(`Error getting professionals: ${JSON.stringify(error)}`)
-        alert(`Error getting data! Please contact our support team by clicking the bottom right link on the page!`)
```
ermish, line 99: While it doesn't have to be an alert, we should let the user know when there's an error so they don't keep waiting and they can message us if they want. (Many people do surprisingly)

A nice toast message would be better here, but that's out of scope for this....how about keep it and replace ina new ticket?

ermish (approved):
looks good! ship the goods!

PR #442, opened 2024-04-01
Update default template for Issues

theyokohamalife:
Updating the template for new issues to help provide devs with more context and details.


ermish (approved):
LGTM! 
I don't think the last part we need to call out as may confuse some people what to put in there

theyokohamalife:
> LGTM! I don't think the last part we need to call out as may confuse some people what to put in there

@ermish Removed the last bit! 😎

PR #458, opened 2024-05-08
fix: disables telemetry on local dev

Ntlja:
## 🔧 What changed
This change fixes a nuxt error to do with telemetry opt-in that is preventing `yarn dev` from working. It will set the opt-in to no by default.
image


theyokohamalife (approved):
LGTM ✨

PR #464, opened 2024-05-10
feat: add page for moderation panel

### LINKED ISSUE #461: Add the Submission Portal as a page
# Problem
We do not want to link to the submission portal but we want a url that allows us to see the components that are being worked on as a page in Nuxt. Currently we do not have this

# Success
Can type the url of the page to view the submssion portal, but has no button that currently links from the main page to this page

### Requirements

- [x] Implement page for submission portal

# Implementation notes
Look at how the other pages are implemented and research what urls are best practice for page views in production

NabbeunNabi:
Resolves #461 

## 🔧 What changed
- Before there was no route for the moderation panel, but the route `/moderation` was added.
- It conditionally renders the page when `ENABLE_MODERATION_PANEL=true` is in `.env` file.

## 📸 Screenshots

-   ### After

image

### DIFF
--- a/nuxt.config.js
+++ b/nuxt.config.js
@@ -144,7 +144,9 @@ export default defineNuxtConfig({
     runtimeConfig: {
         public: {
             // eslint-disable-next-line
-            GOOGLE_MAPS_API_KEY: process.env.GOOGLE_MAPS_API_KEY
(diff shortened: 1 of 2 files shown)


theyokohamalife (approved):
LGTM ✨

PR #465, opened 2024-05-13
Create form to edit submissions in moderation panel

### LINKED ISSUE #460: Component to Edit Facilities in the Moderation Panel
# Problem
Create the form UI component for the Edit Facilities in the Moderation Panel. The form should be scrollable.

Image
# Success
Follow the [Figma](https://www.figma.com/file/QXQxlw4jNFIhMxkKU7bbwq/Find-a-doc%2C-Japan?type=design&node-id=1611-23459&mode=design&t=Sh8sFMmYgDiy40di-0 ). The sidebar of this should scroll to the area that is clicked in the form.

### Requirements

- [ ] Scroll Functionality 
- [x] Create all the form fields with placeholder:
  - [x] Contact information
  - [x] Addresses
  - [x] Google Maps Information
- [ ] Having basic validation

# Implementation notes
Being able to send a mutation to the backend will be added later on. This issue we will focus on the UI aspect

Anissa3005:
Resolves #460 

## 🔧 What changed

- Create the UI for the edit submission form for the facilities
- Create validations for the input field 
- Implement a pinia store for storing our field values

## 📸 Screenshots

The form is quite long could not fit it on one screenshot.

image

image

## To Do 

Write validations for google url, at the moment it only checks if it's a general url.


theyokohamalife:
@Anissa3005 What do you think about renaming the PR title to something like, "Create form to edit submissions in moderation panel"?

components/ModerateSubmissionsEditForm.vue
```diff
@@ -0,0 +1,97 @@
...
+            </div>
+
+        </form>
+    </div>
+</template>
+
+<script lang="ts" setup>
+import {ref} from 'vue'
+
+let chosenFormName = ref<string>('Contact information')
```
ermish, line 87: let's i18n'ify these! Can you move these names to the i18n json files? This will allow headers in different languages

components/ModerateSubmissionsEditForm.vue
```diff
@@ -0,0 +1,97 @@
...
+                    v-else
+                    type="text" 
+                    class="mb-5 px-3 py-3.5 w-[350px] h-[50px] bg-white rounded-lg border border-zinc-400 text-neutral-600 text-sm font-normal font-['Noto Sans JP']"
+                />
+                </div>
+            </div>
+
+            <div v-else-if="chosenFormName == 'Healthcare professional IDs'">
+                <span class="mb-3.5 text-center text-zinc-900 text-2xl font-bold font-['Noto Sans JP'] leading-normal">
+                    Healthcare professional IDs
```
ermish, line 70: same as other comment, let's move these to i18n files

components/ModerateSubmissionsEditForm.vue
```diff
@@ -0,0 +1,97 @@
...
+                </span>
+            </div>
+
+            <div v-else-if="chosenFormName == 'Change Log'">
+                <span class="mb-3.5 text-center text-zinc-900 text-2xl font-bold font-['Noto Sans JP'] leading-normal">
+                    Change Log
```
ermish, line 76: same as other comment, let's move these to i18n files

components/ModSubmissionsEditForm.vue
```diff
@@ -0,0 +1,163 @@
...
+            </div>
+        </div>
+        <button type="submit" class="bg-currentColor text-white font-bold py-2 px-4 rounded">submit</button>
+    </form>
+</template>
+
+<script lang="ts" setup>
+import {ref} from 'vue'
+
+const inputFields = ref({
```
Anissa3005, line 42: Maybe we can store this field object somewhere else, so we can make this mare dynamic, if we need the form for HP we can import the object with all the fields HP need and vica versa.

stores/moderationFormInputStore.ts
```diff
@@ -0,0 +1,254 @@
...
+                label: 'Name Jp',
+                placeholder: 'e.g. 立川中央病院',
+                type: 'text',
+                required: true,
+                inputValid: true,
+                inputValidationMessage: ''
+            },
+            phone: {
+                label: 'Phone',
+                placeholder: 'e.g. 08080901022',
```
theyokohamalife, line 26: I would change the placeholder to something that (probably) isn't a real phone number.

Anissa3005: Like this ?: 08XXXXXXXXXX

theyokohamalife (commented):
It's looking really good! I know that adding tests would be really hard since this isn't connected to any data yet. I think we should at least add `data-testid=""` with the name included for each of the fields so it'll be easier to write the tests as soon as possible.

components/ModSubmissionsEditFormFacility.vue
```diff
@@ -0,0 +1,252 @@
...
+            <span class="mb-3.5 text-center text-zinc-900 text-2xl font-bold font-['Noto Sans JP'] leading-normal">
+                {{ $t('moderateSubmissionForm.googleMapsInformation') }}
+            </span>
+            <ModInputField
+                data-testid="submission-form-google-maps"  
+                label="Google maps URL"
+                type="url"
+                placeholder="e.g. https://www.google.co.jp/maps/"
+                :required="true"
+                :inputValidationCheck="validateWebsite"
```
Anissa3005, line 158: This validation check is here as 'placeholder' since Google URLs are very long the existing regex doesn't work for this field. At the moment when we submit it just checks for a general url, but that's all.

i18n/locales/en.json
```diff
@@ -98,6 +98,12 @@
         "status": "Status",
         "modified": "Modified",
         "submitted": "Submitted"
+    },
+    "moderateSubmissionForm": {
+        "contactInformation": "Contact Information",
+        "addresses": "Addresses",
+        "googleMapsInformation": "Google maps information",
+        "healthcareProfessionalId": "Healthcare Professional IDs"
```
theyokohamalife, line 106: Sorry I forgot to mention this before, but let's add these strings to ja.json as well. They don't need to be translated yet, but we at least want them there so they'll be included in the next round of translations.

cypress/e2e/moderationDashboard.cy.ts
theyokohamalife, line 1: Let's add a ticket to make localization strings for all of the placeholder texts as well.

theyokohamalife (approved):
LGTM ✨

PR #476, opened 2024-05-19
feat: implement main container for submission

### LINKED ISSUE #455: Create the `SubmissionList` component for Moderation Panel
# Problem
There isn't a list component that shows the requested submissions. This should use the component from issue #452 to list the submissions.

# Success
Be able use the fetched data from the API and then display the data using the component listed above. Should have styling matching [Figma](https://www.figma.com/file/QXQxlw4jNFIhMxkKU7bbwq/Find-a-doc%2C-Japan?type=design&node-id=1621-17300&mode=design&t=YlT64bYAuVCI537V-4 )

Image

This uses the component from #452  and maps through them to show the whole list

One of this 
### Requirements

- [ ] Able to pass props to the following component
- [ ] Show the List of Submissions

# Implementation notes
Currently the `Status` is not implemented so we need to put filler data for now in order to wait for the backend

NabbeunNabi:
Resolves #455 

## 🔧 What changed
Before there was no main component for us to hold the submissions. This created a list component that can hold the data based on the state and will change what is displayed under the heading based on what is clicked in the dashboard. There is commented out code to show how it can be implemented in the future to use data to preview the data we want. We will set the data it is looping through based on the states in the sidebar component that in the future will run a function that can filter the data based on the states.
The grid system was chosen due to recommendations to avoid semantic tables from @ermish due to css problems that could arise.

## 📸 Screenshots

-   ### Before
image

-   ### After

image

image

### DIFF
--- a/components/ModMainContent.vue
+++ b/components/ModMainContent.vue
@@ -1,7 +1,7 @@
 <template>
     <div class="h-full">
-        <div v-if="store.activeScreen === ModerationScreen.dashboard ">
-            <p class="text-xl font-bold">MAIN CONTAINER MODERATION DASHBOARD</p>
+        <div v-if="store.activeScreen === ModerationScreen.dashboard">
+            <ModSubmissionListContainer />
         </div>
         <div v-else-if="store.activeScreen === ModerationScreen.editSubmission">
             <p class="text-xl font-bold">MAIN CONTAINER REVIEW SUBMISSION</p>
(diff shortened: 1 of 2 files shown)


NabbeunNabi:
I am sure there will be lots of changes requested. But I thought I would get this started as I can implement the logic with the other component which I will work on and replace the placeholder data with

theyokohamalife (approved):
Nice start!

PR #490, opened 2024-05-25
chore: resize and rename members pics

theyokohamalife:
## 🔧 What changed

The About page was loading very slowly (as high as 8 seconds), so a few changes were made to the images in the Members section:
1. All files were converted to `.jpg`
2. All files were reduced in size to 200 kb or less
3. All file names were changed to separate first and last names with an underscore

## 🧪 Testing instructions
1. Check that the pictures of all Members are visible on the About page
2. Check and compare the payload size on prod vs the preview

## 📸 Screenshots

-   ### Before
Screenshot 2024-05-25 at 9 28 35 PM

-   ### After
Screenshot 2024-05-25 at 9 28 18 PM


NabbeunNabi (approved):
Great! On the payload for the local running of the new pics I see this
image


Which is significantly smaller and faster than production which was told to test at this:
image

PR #491, opened 2024-05-26
chore: remove the coming soon page

theyokohamalife:
## 🔧 What changed
This will remove the coming soon page that we haven't needed for a while. The component isn't used anywhere, so the only change necessary is deleting the file.

## 🧪 Testing instructions
1. Confirm that CI still passes and the preview looks fine.


NabbeunNabi (approved):
LGTM. I checked the preview. The only error was getting the data. Which my guess is just that the preview URL doesn't have access to the backend like on other PR previews which I checked also do not get the data for the facilities.

PR #493, opened 2024-05-26
fix: improve responsiveness and component visibility

theyokohamalife:
## 🔧 What changed

When a user is viewing the site on a desktop, the TopNav and Footer should always be visible without having to scroll. Meaning, the main content should be contained between them and be the only component that scrolls up and down.

This behavior wouldn't work that great on tablet and mobile screens due to their much smaller sizes. So, some adjustments were made to what items will be displayed depending on what screen size the user is using.

1. Makes the TopNav and Footer always visible on desktop
2. Makes the TopNav display the HamburgerComponent on screens smaller than desktop. This remedies the issue of menu items overlapping for sizes between tablets and desktop.
3. Creates separate menu items for the hamburger menu. Items that were previously included in the footer were moved here, but formatting will be needed to improve the UI/UX.
4. Improves the responsiveness of the map and search results list on screen sizes smaller than desktop.

## 🧪 Testing instructions
1. Check the preview and confirm that the TopNav and Footer are visible without scrolling on all pages on desktop.
2. Check the preview and confirm that the hamburger menu is visible and the footer is invisible on all pages for screens smaller than desktop.
3.  Check that the hamburger menu items include the Privacy, Terms, Github, and Netlify links.

### Desktop
https://github.com/ourjapanlife/findadoc-web/assets/31802656/367f092b-227c-44ce-a865-64e093f63260

### Mobile
https://github.com/ourjapanlife/findadoc-web/assets/31802656/85278198-5eb0-4c2b-a1f1-5bbb5a4cec77

### DIFF
--- a/app.vue
+++ b/app.vue
@@ -1,22 +1,31 @@
 <template>
-    <div>
+    <!-- The main app should be set to h-screen. This is to ensure that the TopNav and Footer are
+        always visible on desktop screens without scrolling.
+    -->
+    <div class="h-screen w-full">
         <div v-if="store.enableModerationPanel" class="h-full flex flex-col font-sans text-primary-text bg-primary-bg">
             <NuxtPage class="flex flex-1" />
         </div>
-        <div v-else-if="!store.enableModerationPanel" class="h-full flex flex-col font-sans text-primary-text bg-primary-bg">
+        <div v-else-if="!store.enableModerationPanel && $viewport.isGreaterThan('desktop')" class="h-full w-full flex flex-col font-sans text-primary-text bg-primary-bg">
             <TopNav />
-            <NuxtPage class="flex flex-1" />
+            <NuxtPage />
             <Footer />
-        </div>      
+        </div>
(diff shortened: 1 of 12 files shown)


NabbeunNabi (commented):
### Question 
The desired effects happen when I check them. but I was wondering if you consulted with anyone on the design team about the hamburger menu. It seems very crowded with a lot of links

theyokohamalife:
> ### Question
> The desired effects happen when I check them. but I was wondering if you consulted with anyone on the design team about the hamburger menu. It seems very crowded with a lot of links

@NabbeunNabi You raise a good point! I think we'll be able to display all of the links neatly in the hamburger menu once it is styled. We should probably think of it more as a side drawer that opens up when clicked. Then it could 3 (or so) separate sections. I posted [an example on the design channel](https://find-a-doc.slack.com/archives/C062ACS4RJA/p1716818680897639).

NabbeunNabi (approved):
Alright. Since the design is in process and for now the functionality is important I think it is good. It works for me. Maybe once the design is made it should be high priority as a UI/UX ticket

PR #494, opened 2024-05-26
test: add cypress testing for pages

theyokohamalife:
Part of #502 

## 🔧 What changed

This PR focuses on improving code quality by implementing testing for pages. Focus was put on testing pages rather than individual components, to cover how the layout appears when everything is rendered together.

## 🧪 Testing instructions
1. Check that CI is passing
2. Start the frontend server with `yarn dev` and run Cypress locally with `yarn cypress open`
3. Confirm that everything in the preview still works


theyokohamalife:
@SocketSecurity ignore-all

cypress/e2e/submit.cy.ts
```diff
@@ -0,0 +1,79 @@
...
+            cy.get('[data-testid="hamburger-component"]').should('not.exist')
+        });
+
+        it('has a heading', () => {
+            cy.get('h1').should('have.attr', 'data-testid', 'submit-heading').contains('Add a Doctor')
+        });
+
+        it('has a subheading', () => {
+            cy.get('[data-testid="submit-subheading"]').contains(
+                "Want to add a doctor to the database? Complete the form below and it will be reviewed by our moderators.")
```
NabbeunNabi, line 22: Same as before. Can we import the JSON file that holds the text?

cypress/e2e/about.cy.ts
```diff
@@ -0,0 +1,77 @@
...
+            cy.get('[data-testid="hamburger-component"]').should('not.exist')
+        });
+
+        it('has a heading', () => {
+            cy.get('[data-testid="about-heading"]').contains('About Find a Doc, Japan')
+        });
+
+        it('has a subheading', () => {
+            cy.get('[data-testid="about-subheading"]').contains('Connecting people in Japan to the healthcare services they need, in the languages they need.')
+        });
```
NabbeunNabi, line 22: This is from when I was writing tests a few months ago. @ermish said we should avoid hard coding text into the tests. I do not know a better way. Can we import the JSON we are using and use template literals. This way any small change that could happen to that text file is automatically updated and checked in both places?

NabbeunNabi (changes requested):

cypress/e2e/submit.cy.ts
theyokohamalife: Oh, I meant to leave a note that importing the JSON works, but the `$t` doesn't. I'll look into a workaround.

cypress/e2e/about.cy.ts
```diff
@@ -0,0 +1,79 @@
+import enUS from '~/i18n/locales/en.json'
```
NabbeunNabi, line 1: I see the checks are failing. So I checked if it was the bug. But also failed on my device. I think because we do not have the `cypress.config.ts` setup to import files this way you should change it here to `import enUS from "../../i18n/locales/en.json"`. When I did this locally all the checks passed for the tests your wrote.

cypress/e2e/submit.cy.ts
```diff
@@ -0,0 +1,80 @@
+import enUS from '~/i18n/locales/en.json'
```
NabbeunNabi, line 1: You should change it here as well to import enUS from "../../i18n/locales/en.json". I could make a ticket that has us fix the config file later but this is the quickfix that I found works.

NabbeunNabi (changes requested):

cypress/e2e/submit.cy.ts
theyokohamalife: Hmm that's strange, because it should follow the paths defined in the /cypress/tsconfig.json.

NabbeunNabi (approved):
Tested locally and CI/CD all working

PR #496, opened 2024-05-27
fix: remove duplicate variable key breaking submit page

theyokohamalife:
## 🔧 What changed

We removed the `variable` key from the input, because we noticed on the backend that it was creating an additional nested key for one that already exists.

## 🧪 Testing instructions
1. Run the backend dev db using `npm run dev:startlocaldb`
2. Run the backend dev server using `npm run dev`
3. Point the frontend endpoint to the backend server
4. Fill out the Submit form on the local frontend server and verify the payload that prints in the backend console.

## 📸 Screenshots
Screenshot 2024-05-27 at 8 30 13 PM


Anissa3005 (approved):
LGTM

PR #501, opened 2024-05-27
chore: wrap submit request in try catch

theyokohamalife:
Part of #502 

## 🔧 What changed

Small PR that wraps the Submit request in a try/catch. Further changes are currently being made in #494 to add Cypress testing for the page. Validation will also be added in a separate PR.


NabbeunNabi (approved):
I look forward to the 1000s of cypress tests we will have

PR #506, opened 2024-05-27
feat: create topbar for edit submission form

### LINKED ISSUE #459: Create Navigation Bar Status template UI for the Moderation Panel
# Problem
Currently there is no way to save, reject, or approve the submissions. We need a component that just logs for now with the UI. Functionality can be added later

# Success
Looks the same as this photo

Image

### Requirements

- [ ] Has the same UI as the photo uploaded from Figma
- [ ] Logs when each button is clicked to the console, so those onClick events can have functionality later
- [ ] Contains a field has an ID and on click copies the ID to the clipboard. Only the ID value not the text `ID: `

# Implementation notes

NabbeunNabi:
Resolves #459 

## 🔧 What changed
Before there was no top navbar. Now there is partial functionality with styling matching that of the UI

## 🧪 Testing instructions
Go to the branch and go to the edit Submission part of the mod panel. When you see the topbar you can click the copy to clipboard. It should show you a check mark for 2 seconds to show that you successfully copied the id. There are comments about how to implement the autosave later. The Reject and Approve work with on click, but we need to create the call to the backend still within the store to put into those functions. But currently it logs. `You are rejected` or `You are accepted`

## 📸 Screenshots

-   ### Before
Nothing was there. Just text saying `TopBar Edit Submission`
Similar to this
image
 

-   ### After
Before clicking to copy
image

After clicking copy for 2 seconds
image


components/ModTopbar.vue
```diff
@@ -1,18 +1,19 @@
...
     <!-- styling will change from px to rm -->
-    <div class="h-[76px] bg-emerald-100">
+    <div class="h-[76px] w-full">
         <div v-if="store.activeScreen === ModerationScreen.dashboard ">
             <p class="text-xl font-bold">TOPBAR MODERATION DASHBOARD</p>
         </div>
-        <div v-else-if="store.activeScreen === ModerationScreen.editSubmission">
-            <p class="text-xl font-bold">TOPBAR REVIEW SUBMISSION</p> 
+        <div class="h-[76px] w-full" v-else-if="store.activeScreen === ModerationScreen.editSubmission">
+           <ModEditSubmissionTopbar id="MattIsAGreatHumanBeing"/>
```
theyokohamalife, line 8: I agree that Matt is a great human being! Let's change this id to one that we can test later using `data-testid=""`

components/ModEditSubmissionTopbar.vue
```diff
@@ -0,0 +1,73 @@
...
+                class="flex flex-row w-90 bg-neutral p-2 m-2 border-2 border-slate-400 rounded hover">
+                ID: {{ id }}
+                <SuccessCheckMark v-if="copySuccess" role="img" title="clipboard copy" class="ml-2 w-6" />
+                <CopyContent v-else role="img" title="successful copy" class="ml-2 w-6" />
+            </button>
+        </div>
+        <div class="flex flex-row justify p-2 font-bold ">
+            <button
+                class="flex justify-center items-center rounded-full bg-white border-slate-400 border-2 w-28 text-sm mr-2 ">Save
+                & Exit</button>
```
theyokohamalife, line 14: I would add this string to the locale files. The "&" looks like it could potentially cause problems. Git doesn't seem too happy with it. 

Screenshot 2024-05-28 at 12 45 09 AM

components/ModTopbar.vue
NabbeunNabi: That id there is a prop at the moment that we are passing but we will add a data test id as well!

components/ModEditSubmissionTopbar.vue
NabbeunNabi: Moved all the text except ID to the text.json file. No Japanese but has the translation string for now

components/ModEditSubmissionTopbar.vue
```diff
@@ -0,0 +1,75 @@
...
+
+const props = defineProps({
+    submissionId: {
+        type: String,
+        required: true,
+    }
+}
+)
+
+let copySuccess: Ref<boolean> = ref(false)
```
theyokohamalife, line 38: I think we could make this variable name more specific as well. We should know from the name what it is that we're copying. Perhaps something like `copiedSubmissionId`?

components/ModTopbar.vue
```diff
@@ -1,18 +1,19 @@
...
-    <div class="h-[76px] bg-emerald-100">
-        <div v-if="store.activeScreen === ModerationScreen.dashboard ">
+    <div class="h-[76px] w-full">
+        <div v-if="store.activeScreen === ModerationScreen.dashboard">
             <p class="text-xl font-bold">TOPBAR MODERATION DASHBOARD</p>
         </div>
-        <div v-else-if="store.activeScreen === ModerationScreen.editSubmission">
-            <p class="text-xl font-bold">TOPBAR REVIEW SUBMISSION</p> 
+        <div class="h-[76px] w-full" v-else-if="store.activeScreen === ModerationScreen.editSubmission">
+            <ModEditSubmissionTopbar data-test-id="mod-edit-submission-topbar" submissionId="MattIsAGreatHumanBeing" />
```
theyokohamalife, line 8: I think we can initialize `submissionId` as a variable with an empty string below.

components/ModEditSubmissionTopbar.vue
```diff
@@ -0,0 +1,75 @@
...
+
+    navigator.clipboard.writeText(id).then(() => {
+        copySuccess.value = true;
+        setTimeout(() => {
+            copySuccess.value = false;
+        }, 2000);
+    }).catch(err => {
+        console.error('Failed to copy ID: ', err);
+    });
+}
```
theyokohamalife, line 51: I think we should add a Cypress test for this behavior. You could write a test that clicks on the button and asserts that the expected checkmark is shown.

components/ModEditSubmissionTopbar.vue
```diff
@@ -0,0 +1,75 @@
...
+    submissionId: {
+        type: String,
+        required: true,
+    }
+}
+)
+
+let copySuccess: Ref<boolean> = ref(false)
+
+const copyId = () => {
```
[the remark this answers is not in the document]
NabbeunNabi: Updated the name of this and other variables

components/ModEditSubmissionTopbar.vue
```diff
@@ -0,0 +1,75 @@
...
+                $t('modEditSubmissionTopNav.reject') }}</button>
+            <button @click="acceptSubmission"
+                class="flex justify-center items-center rounded-full bg-primary w-28 text-white text-sm mr-2 ">{{
+                $t('modEditSubmissionTopNav.approve') }}</button>
+        </div>
+    </div>
+</template>
+
+<script setup lang="ts">
+import CopyContent from '~/assets/icons/content-copy.svg'
```
ermish, line 26: let's prefix the svg component names with `SVG*` to make it easy to identify

components/ModEditSubmissionTopbar.vue
```diff
@@ -0,0 +1,75 @@
...
+}
+)
+
+let copySubmissionIdSuccess: Ref<boolean> = ref(false)
+
+const copySubmissionId = () => {
+    const submissionIdFromProps = props.submissionId
+
+    navigator.clipboard.writeText(submissionIdFromProps).then(() => {
+        copySubmissionIdSuccess.value = true;
```
ermish, line 44: nit: remove the semicolons

components/ModEditSubmissionTopbar.vue
```diff
@@ -0,0 +1,75 @@
...
+import CopyContent from '~/assets/icons/content-copy.svg'
+import SuccessCheckMark from '~/assets/icons/checkmark-square.svg'
+import { Ref, ref } from 'vue';
+
+const props = defineProps({
+    submissionId: {
+        type: String,
+        required: true,
+    }
+}
```
ermish, line 35: can you run the linter? I'm surprised these are on separate lines and also not tabbed

components/ModEditSubmissionTopbar.vue
```diff
@@ -0,0 +1,75 @@
...
+
+let copySubmissionIdSuccess: Ref<boolean> = ref(false)
+
+const copySubmissionId = () => {
+    const submissionIdFromProps = props.submissionId
+
+    navigator.clipboard.writeText(submissionIdFromProps).then(() => {
+        copySubmissionIdSuccess.value = true;
+        setTimeout(() => {
+            copySubmissionIdSuccess.value = false;
```
ermish, line 46: I was thinking it failed if it hit this timeout at first glance. I think we should rename the variable something like `showCopySuccessMessage` or something like that

ermish (changes requested):
coming along!

components/ModEditSubmissionTopbar.vue
NabbeunNabi: Just ran the linter. It is setting it up the same way as the props in the `SearchResultsListItem.vue`

components/ModEditSubmissionTopbar.vue
NabbeunNabi: changed it to `showCopySuccessIcon` since it shows the SVG, not a message

theyokohamalife (approved):
LGTM

components/ModSubmissionListContainer.vue
```diff
@@ -1,49 +1,56 @@
...
-    return `grid-template-columns: ${defaultColumnWidth}% repeat(${numColumns - 1}, ${columnWidth}%);`;
+    const columnWidth = remainingWidth / (numOfColumns - 1);
+    return `grid-template-columns: ${defaultColumnWidth}% repeat(${numOfColumns - 1}, ${columnWidth}%);`;
 });
 
-const accessSubmissionForm = (id:String) => {
-    // logic on how to pass the data to the submission form that is being edited
-}
+const handleClickToSubmissionForm = (id: string) => {
+    console.log("WE CLICKED THE THING!!")
```
ermish, line 50: should remove if the logic is working now

i18n/locales/ja.json
```diff
@@ -14,6 +14,37 @@
...
+        "about": "About Find a Doc, Japan",
+        "home": "Home",
+        "contact": "Contact",
+        "submit": "Submit",
+        "privacy": "Privacy",
+        "terms": "Terms",
+        "github": "Github",
+        "netlify": "Powered by Netlify"
+    },
+    "privacyPage": {
```
ermish, line 27: these changes should realistically be a separate PR

ermish: for the sake of time, I won't block on it, but realistically should try to make more atomic PRs

pages/moderation.vue
```diff
@@ -12,11 +12,11 @@
 
 <script setup lang="ts">
 import { useModerationScreenStore, ModerationScreen } from "~/stores/moderationScreenStore"
-import { useRoute } from 'vue-router'
+// import { useRoute } from 'vue-router'
```
ermish, line 15: I think this code can be removed?

ermish (approved):
a few cleanup things and it looks good!

pages/moderation.vue
NabbeunNabi: If you remove this code make sure all the other cypress tests are passing first. This is what was keeping the layout from shifting due to the env variable being implemented in development. Since the page layout is different for every other route once the env variable is introduced it can mess up the layout of every other page. I do not know either how on production we will make it know that it is on the moderation page to show the correct layout.

NabbeunNabi: Assuming this code isn't being utilized in the place it was previously being implemented

ermish: oh I was just referring to the commented out portion

i18n/locales/ja.json
theyokohamalife: @ermish Agreed! These were only included, because we noticed they were missing while adding the new strings. I think we might want to implement an i18n plugin like we had in 1.0 to make the locale files easier to maintain.

pages/moderation.vue
theyokohamalife: Actually, this file shouldn't change. Will undo them.

PR #512, opened 2024-05-29
fix: change the old import to new style

NabbeunNabi:
## 🔧 What changed
The images were not appearing properly due to the change of the `svg` from img to component style. Updated to be the way we import the `svg` now

## 📸 Screenshots

-   ### Before
image

-   ### After
image

### DIFF
--- a/components/ModDashboardLeftNavbar.vue
+++ b/components/ModDashboardLeftNavbar.vue
@@ -12,7 +12,7 @@
                 :class="{ 'border-b border-gray-300': index !== submissionListViewOptions.length - 1 }"
                 v-for="(submissionListEntry, index) in submissionListViewOptions ">
                 <div class="flex items-center justify-center w-1/5">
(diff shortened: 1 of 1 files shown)


components/ModDashboardLeftNavbar.vue
```diff
@@ -38,19 +38,19 @@ const modDashboardViewOptions = [{
 const submissionListViewOptions = [{
```
ermish, line 38: nit: unrelated to this, but I think the UI elements would be easier if they were static vs a dynamic list. 
Then you can just import the icons with the nice named component style. 

for this PR, this is fine though

ermish (approved):
ship ship

components/ModDashboardLeftNavbar.vue
theyokohamalife: @NabbeunNabi Could you make a ticket for this suggestion?

PR #513, opened 2024-05-30
fix: change handling of mod store visibility to use path

NabbeunNabi:
## 🔧 What changed
Before we were handling the visibility of the route for  `/moderation` with only and environmental variable which is mocking the feature flag variable for now. This now conditionally handles the variable and the visibility of the route with not only the `env` variable being set to true, but if the `pathname` includes `moderation`. I implemented it this way in case we have nested routes later.

## 🧪 Testing instructions
-This will allow for successful testing of moderation panel testing in the future since it will not change the layout of the other pages, which would cause all those cypress tests to fail

## 📸 Screenshots

-   ### Before
-This is how the layout was being affected before.
WhatsApp Image 2024-05-30 at 13 19 48_609f8283

-   ### After
This is what it looks like with new implementaiton
image

-Moderation panel still displays

image

### DIFF
--- a/stores/moderationScreenStore.ts
+++ b/stores/moderationScreenStore.ts
@@ -1,6 +1,5 @@
 import { defineStore } from 'pinia'
 import { Ref, ref } from 'vue'
-import { useRuntimeConfig } from "#imports"
 
 export enum ModerationScreen {
     'dashboard',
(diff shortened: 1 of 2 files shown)


stores/moderationScreenStore.ts
```diff
@@ -10,11 +10,15 @@ export enum ModerationScreen {
...
-    const enableModerationPanel: Ref<boolean> = ref(runtimeConfig.public.ENABLE_MODERATION_PANEL || false)
+    const enableModerationPanel: Ref<boolean> = ref(false)
 
     function setActiveScreen(newValue: ModerationScreen) {
         activeScreen.value = newValue
     }
 
-    return { activeScreen, enableModerationPanel, setActiveScreen }
+    function setEnableModerationPanelToTrue (currentPath:string){  
+        runtimeConfig.public.ENABLE_MODERATION_PANEL && currentPath.includes("moderation") ? enableModerationPanel.value = true : null
```
theyokohamalife, line 20: Could we avoid using `runtimeConfig` and just use `process.env.ENABLE_MODERATION_PANEL` directly? That would make testing way easier.

theyokohamalife (approved):
LGTM ✨

PR #528, opened 2024-06-04
fix/show loading leftnav

NabbeunNabi:
## 🔧 What changed
Before it was showing no results instead of loading while fetching results from the store. Now it displays no results found if there are only no results and shows the loading svg when first fetching from backend

## 🧪 Testing instructions
Run `yarn cypress open` and run the home tests. Since it is now using the dev database instead of production one test will fail but maybe we can address this in a different PR.

## 📸 Screenshots

-   ### Before
image

-   ### After

image

If No search results after selecting urology from local db

image

### DIFF
--- a/components/SearchResultsList.vue
+++ b/components/SearchResultsList.vue
@@ -13,7 +13,20 @@
                 </span>
             </button>
         </div>
-        <div v-if="hasResults">
+        <div v-if="loadingStore.isLoading">
+            <div class="h-full flex flex-col justify-center items-center w-full">
+<<<<<<< HEAD
+                <SVGLoading class="w-4/5 p-12" />
+=======
+                <div class="flex justify-center align-middle">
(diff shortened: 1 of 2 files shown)


components/SearchResultsList.vue
```diff
@@ -43,15 +53,22 @@
...
+import { useLoadingStore } from '../stores/loadingStore'
 import { Locale } from '~/typedefs/gqlTypes.js';
 import noSearchResultsGraphic from '@/assets/icons/no-search-results-graphic.svg'
 const resultsStore = useSearchResultsStore()
 const localeStore = useLocaleStore()
-//let's start with some initial data
-resultsStore.search()
+const loadingStore = useLoadingStore()
+
+onMounted(() => {
```
ermish, line 68: nice improvement!

ermish (approved):
great work!

PR #534, opened 2024-06-05
Create healthcare professional search component

### LINKED ISSUE #529: Create a search healthcare professional by facility id component
# Problem
Create a component were we can search a healthcare professionals by facility id. This component should be added in the facility edit submission form.

# Success

Image

### Requirements

- [x] Create a new component
- [x] Create the UI for the search bar
- [x] Add this component to the facility edit submission form
- [x] Add basic tests for the component

# Implementation notes

The UI for this component already has been created in the facility edit submission form. We just have to move that code to a separate component and edit the testing.

Anissa3005:
Resolves #529 

## 🔧 What changed

- Create healthcare professional search bar component for the facility edit form
- Applied the UI for the search bar
- Removed some small bugs that I notice in the form component

## 🧪 Testing instructions

Run the cypress tests.

## 📸 Screenshots

image

## To Do

Fetch the healthcare professionals for the facility

### DIFF
--- a/components/ModHealthcareProfessionalSearchbar.vue
+++ b/components/ModHealthcareProfessionalSearchbar.vue
@@ -1,0 +1,23 @@
+<template>
+    <div>
+        <span class="mb-3.5 text-center text-primary-text text-2xl font-bold font-['Noto Sans JP'] leading-normal">
+            {{ $t('modSubmissionForm.healthcareProfessionalId') }}
+        </span>
+        <div class="flex flex-col mt-4">
+            <div class="justify-start items-start flex">
+                <input 
+                    type='text'
+                    :placeholder="$t('modSubmissionForm.placeholderTextHealthcareProfessionalSearchbar')"
+                    class="mb-5 px-3 py-3.5 w-96 h-12 bg-white rounded-lg border border-primary-text-muted text-primary-text text-sm font-normal font-['Noto Sans JP'] placeholder-primary-text-muted"
+                />
+                <SVGLookingGlass role="img" title="searching icon" class="relative right-8 top-3 w-6 h-6"/>
+            </div>
+        </div>
+    </div> 
+</template>
+
+<script lang="ts" setup>
+import {ref, Ref} from 'vue'
+
+import SVGLookingGlass from '~/assets/icons/looking-glass.svg'
+</script>
--- a/components/ModSubmissionsEditFormFacility.vue
+++ b/components/ModSubmissionsEditFormFacility.vue
@@ -2,7 +2,7 @@
     <form @submit="validateFields" class="p-4 h-full overflow-y-auto">
         <div>
             <span class="mb-3.5 text-center text-zinc-900 text-2xl font-bold font-['Noto Sans JP'] leading-normal">
-                {{ $t('moderateSubmissionForm.contactInformation')}}
+                {{ $t('modSubmissionForm.contactInformation')}}
             </span>
             <ModInputField  
                 data-testid="submission-form-nameEn"
@@ -53,7 +53,7 @@
 
         <div>
             <span class="mb-3.5 text-center text-zinc-900 text-2xl font-bold font-['Noto Sans JP'] leading-normal">
-                {{ $t('moderateSubmissionForm.addresses') }}
+                {{ $t('modSubmissionForm.addresses') }}
             </span>
             <ModInputField
                 data-testid="submission-form-postalCode"  
@@ -147,7 +147,7 @@
         
         <div>
             <span class="mb-3.5 text-center text-zinc-900 text-2xl font-bold font-['Noto Sans JP'] leading-normal">
-                {{ $t('moderateSubmissionForm.googleMapsInformation') }}
+                {{ $t('modSubmissionForm.googleMapsInformation') }}
             </span>
             <ModInputField
                 data-testid="submission-form-google-maps"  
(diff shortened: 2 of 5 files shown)


components/ModHealthcareProfessionalSearchbar.vue
```diff
@@ -0,0 +1,21 @@
+<template>
+    <div>
+        <span class="mb-3.5 text-center text-zinc-900 text-2xl font-bold font-['Noto Sans JP'] leading-normal">
```
ermish, line 3: avoid using custom colors, try using one of our theme variables like `text-primary`, `text-secondary`, or `text-tertiary` or one of their subsets like this:

image

components/ModHealthcareProfessionalSearchbar.vue
```diff
@@ -0,0 +1,21 @@
...
+            {{ $t('modSubmissionForm.healthcareProfessionalId') }}
+        </span>
+        <div class="flex flex-col mt-4">
+            <div class="justify-start items-start flex">
+                <input 
+                    type='text'
+                    placeholder='Search by name or ID to add a doctor'
+                    class="mb-5 px-3 py-3.5 w-[350px] h-[50px] bg-white rounded-lg border border-zinc-400 text-neutral-600 text-sm font-normal font-['Noto Sans JP'] placeholder-gray-300"
```
ermish, line 11: for inputs, I recommend avoiding fixed pixel widths, try to use a related size or grid, like `w-1/3` for example. this will help it be more responsive across difference resolutions and device sizes

components/ModHealthcareProfessionalSearchbar.vue
```diff
@@ -0,0 +1,21 @@
...
+                />
+                <SVGLookingGlass role="img" title="searching icon" class="relative right-8 top-3 w-6 h-6"/>
```
ermish, line 13: does this need to be relative?

ermish (changes requested):
nice component extraction!

theyokohamalife:
Nit: 'hp' could mean a lot of things, like hit points! 😆 Also let's be careful not to include spelling mistakes in the PR title, like "serachbar". I'd suggest something along the lines of "Create healthcare professional search component", and add the "moderation panel" label to the PR.

components/ModHealthcareProfessionalSearchbar.vue
```diff
@@ -0,0 +1,21 @@
+<template>
+    <div>
+        <span class="mb-3.5 text-center text-zinc-900 text-2xl font-bold font-['Noto Sans JP'] leading-normal">
+            {{ $t('modSubmissionForm.healthcareProfessionalId') }}
+        </span>
+        <div class="flex flex-col mt-4">
+            <div class="justify-start items-start flex">
+                <input 
+                    type='text'
+                    placeholder='Search by name or ID to add a doctor'
```
theyokohamalife, line 10: Let's save the placeholder text to a variable.

components/ModSubmissionsEditFormFacility.vue
```diff
@@ -177,31 +177,15 @@
                 invalidInputErrorMessage="Invalid Longitude"
```
theyokohamalife, line 177: Not part of this change, but we should avoid using raw text in the template like here.

components/ModHealthcareProfessionalSearchbar.vue
Anissa3005: In the Figma the image is in the search box, and this was the only way to put an image in a  input field. 
I can also use a div to create a search bar and than I don't have to use the relative, which one do you recommend?

ermish: I think the container div makes more sense. Relative is also ok, but it inherently comes with more risk of UI abnormalities since it renders dependent on another element.

components/ModHealthcareProfessionalSearchbar.vue
Anissa3005: This is a preview using the theme colors for the search box:

image

I read you can also set 'classes'  (in the tailwind.ts file) in tailwind, for example create an input class and use that style on all the input fields. Is this something we can do, since we have so many input fields?

components/ModHealthcareProfessionalSearchbar.vue
Anissa3005: Do we want to do this for all input fields? I'm thinking about the longer forms this means we have to create 14-16 variables + the ErrorMessage variables that you mention. 

If so let me know than I will create an issue for the form so I can apply it there as well.

components/ModHealthcareProfessionalSearchbar.vue
Anissa3005: Ah I know why I went with input, because of the placeholder text. I think that requires some css/tailwind skill. I will try if I can figure it out.

components/ModHealthcareProfessionalSearchbar.vue
ermish: yup! every error message, every piece of text.....

components/ModHealthcareProfessionalSearchbar.vue
```diff
@@ -0,0 +1,25 @@
...
+                <SVGLookingGlass role="img" title="searching icon" class="relative right-8 top-3 w-6 h-6"/>
+            </div>
+        </div>
+    </div> 
+</template>
+
+<script lang="ts" setup>
+import {ref, Ref} from 'vue'
+
+const placeholderText: Ref<string> = ref('Search by name or ID to add a doctor')
```
ermish, line 22: this should still be pulled from i18n. I think don't think it need to be a ref vs just a variable?

ermish (changes requested):
just the i18n string and it's good to go!

components/ModHealthcareProfessionalSearchbar.vue
Anissa3005: Ah sorry I totally misunderstood what you and LaShawn meant, but now I understand thank you! Will fix this 🙏

theyokohamalife (approved):
LGTM ✨

ermish (approved):
💯  nice!

PR #535, opened 2024-06-05
fix: update scripts with nuxt env and semicolon

NabbeunNabi:
## 🔧 What changed

1. I changed the setting of the `env` variable to `NUXT_USE_LOCAL_API` and added a semicolon to the script after the setting of the `env` variable.
2. I changed the `nuxt.config` to then have that variable accessible during the `runtime.config` and then used it as a variable passed into the `InitializeGqlClient`.
 **Explanation**: On the server the `utils/graph.ts` was able to access the set `NUXT_USE_LOCAL_API=development` `env` variable. But since we use that function within the app, when it ran the function `initializeGqlClient` in `app.vue` the client didn't have access to that set variable from the script. This led the function call to default to the prod `url` client side. Now in `app.vue` we pass the `runtimeConfig` as a parameter that was set in the `nuxt.config` as the `NUXT_USE_LOCAL_API=development` into the `initializeGqlClient`  function. This allows us to access the `env` during runtime to correctly set the `apiUrl` in `initializeGqlClient` .

## 🧪 Testing instructions
Run `yarn dev` and that should connect to the prod db.
image

Run `yarn dev:localserver` and that should show the following error.
image

Start your local backend server and then:
Run `yarn dev:localserver` and that should show the page with the dummy data.

image


theyokohamalife:
@SocketSecurity ignore-all

theyokohamalife (approved):
Awesome work! I've confirmed that it works for me, but let's wait for @Anissa3005 and @ermish to also confirm it!

NabbeunNabi:
@theyokohamalife Good idea! Let's hope!

nuxt.config.js
```diff
@@ -150,7 +150,9 @@ export default defineNuxtConfig({
             // eslint-disable-next-line
```
Anissa3005, line 150: The problem I have now is it doesn't set the `process.env.ENABLE_MODERATION_PANEL` I have it set to true but returns undefined so it doesn't set the env variable. My guess is that the nuxt config file just ignores our .env file

Anissa3005 (changes requested):
The problem I have now is it doesn't set the process.env.ENABLE_MODERATION_PANEL I have it set to true but returns undefined so it doesn't set the env variable. My guess is that the nuxt config file just ignores our .env file

Anissa3005 (approved):

nuxt.config.js
Anissa3005: Solved!

package.json
```diff
@@ -5,7 +5,7 @@
     "scripts": {
         "prepare": "husky",
         "dev": "yarn && nuxi dev",
-        "dev:localserver": "NODE_ENV=development yarn && nuxi dev",
+        "dev:localserver": "NUXT_USE_LOCAL_API=development; yarn && nuxi dev",
```
ermish, line 8: why `development` vs just `true` 😆

app.vue
```diff
@@ -3,10 +3,12 @@
...
     -->
     <div data-testid="app" class="h-screen w-full">
-        <div v-if="store.enableModerationPanel" data-testid="moderation-content" class="h-full w-full flex flex-col font-sans text-primary-text bg-primary-bg">
+        <div v-if="store.enableModerationPanel" data-testid="moderation-content"
+            class="h-full w-full flex flex-col font-sans text-primary-text bg-primary-bg">
             <NuxtPage class="flex flex-1" />
         </div>
-        <div v-else-if="!store.enableModerationPanel && $viewport.isGreaterThan('desktop')" class="h-full w-full flex flex-col font-sans text-primary-text bg-primary-bg">
+        <div v-else-if="!store.enableModerationPanel && $viewport.isGreaterThan('desktop')"
+            class="h-full w-full flex flex-col font-sans text-primary-text bg-primary-bg">
```
ermish, line 11: you can move the viewport check to be a css media query
like this
`portrait:hidden`

```suggestion
            class="portrait:hidden h-full w-full flex flex-col font-sans text-primary-text bg-primary-bg">
```

utils/graphql.ts
```diff
@@ -2,14 +2,14 @@ import { GraphQLClient } from "graphql-request";
...
-export const initializeGqlClient = () => {
+export const initializeGqlClient = (useLocalApi: string | undefined) => {
     if (gqlClient) {
         return gqlClient
     }
 
     let apiURL
 
-    if (process.env.NODE_ENV == 'development') {
+    if (useLocalApi === "development") {
```
ermish, line 12: this is ....confusing... let's set useLocalApi to true/false instead of strings.

ermish (changes requested):
quick change, then let's merge this 👍

app.vue
ermish: you forgot to add the class 😆

ermish: no wait, I see it on the one above 💯

NabbeunNabi: This needed to go into line 7

ermish (approved):
ya'll fixed it! ⭐

ermish (commented):
ya'll fixed it! ⭐ 

Before I hit merge, let's make sure those cypress tests didn't break from this. Can you confirm @NabbeunNabi ?

ermish (approved):
💯

NabbeunNabi:
> 💯

The `portrait hidden` was still making it show the footer? I am a little confused as to how those work so do you mind if I add the CSS changes in its own PR?

Issue #540, opened 2024-06-07
BUG: fix mod dashboard tests


ermish:
# Problem
There are some tests failing in CI
- `"it shows the moderation top nav"` this is one failing locally as well, so I added `.skip()` to it until it's fixed. It seems like the clipboard is being triggered before it's available

# Success
All tests pass!

NabbeunNabi:
Blocked while @ermish and team discuss the best way to create a testing environment

NabbeunNabi:
This was fixed within another ticket using lifecycle hooks that @gminetoma noticed and we fixed. So We should be able to safely remove it now

PR #547, opened 2024-06-07
refactor: update problematic tests

theyokohamalife:
Part of #540 

## 🔧 What changed

1. This PR adds `.skip` to the test that is making fake submissions to the prod database. We'll re-enable this once we figure out mocking. 

2. This PR also changes `.blur()` in tests to `.tab()` because the command was causing Cypress to fail consistently. The purpose of the blur was to make the error messages appear, but it also had the effect of changing the DOM. The [cypress-plugin-tab](https://github.com/Bkucera/cypress-plugin-tab) allows us to use tabs in our tests, which is helpful for triggering the input validations.

Screenshot 2024-06-07 at 8 14 20 PM

3. Increases the wait time between moderation tests that were failing due to the elements not having enough time to load.
4. Upgrades the [Cypress Github Action](https://github.com/cypress-io/github-action) to (hopefully) fix the "server not running" errors that sometimes pop up in CI.

## How to test
1. Run `yarn dev`
2. Run `yarn cypress run`

### DIFF
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -15,11 +15,11 @@ jobs:
         runs-on: ubuntu-latest
         steps:
             - name: Checkout code
-              uses: actions/checkout@v3
+              uses: actions/checkout@v4
             # Install yarn dependencies, cache them correctly
             # and run all Cypress tests
             - name: Cypress run
-              uses: cypress-io/github-action@v5.6.1
+              uses: cypress-io/github-action@v6
               with:
                   build: yarn prod:build
                   start: yarn prod:start
--- a/package.json
+++ b/package.json
@@ -54,6 +54,7 @@
         "all-contributors-cli": "^6.26.1",
         "autoprefixer": "^10.4.18",
         "cypress": "^13.10.0",
+        "cypress-plugin-tab": "^1.0.5",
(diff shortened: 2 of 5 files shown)


.github/workflows/cypress.yml
```diff
@@ -15,11 +15,11 @@ jobs:
         runs-on: ubuntu-latest
         steps:
             - name: Checkout code
-              uses: actions/checkout@v3
+              uses: actions/checkout@v4
             # Install yarn dependencies, cache them correctly
             # and run all Cypress tests
             - name: Cypress run
-              uses: cypress-io/github-action@v5.6.1
```
ermish, line 22: not part of this atomic PR context, but the refactor is in the same code space so approved by me! 😸

package.json
```diff
@@ -54,6 +54,7 @@
         "all-contributors-cli": "^6.26.1",
         "autoprefixer": "^10.4.18",
         "cypress": "^13.10.0",
+        "cypress-plugin-tab": "^1.0.5",
```
ermish, line 57: what a dumb plugin need 🤣  I read how people have been waiting since 2017 for the ability to tab so this still plugin still exists

ermish (approved):
:shipit:

ermish (approved):

PR #548, opened 2024-06-07
refactor: improve visibility of search results

theyokohamalife:
## 🔧 What changed
Currently, it's difficult for users to understand when there are no search results. The information is rendered off screen on mobile in both portrait and landscape mode. This PR improves the visibility of the "No Results Found" message on mobile, so most users will be aware that it's there without having to scroll. 

It also improves the visibility of the Search Results List, by enabling users to scroll to the bottom of the last result. For iPhone users, this information is currently blocked by the toolbar.

## 🧪 Testing instructions
1. Test the preview on an iPhone in portrait and landscape mode.
2. Test the preview on a desktop browser and drag the size.

## 📸 Screenshots
Desktop before
Screenshot 2024-06-08 at 12 27 55 AM

Desktop after**
Screenshot 2024-06-08 at 12 35 19 AM

Mobile landscape before:
IMG_5247

Mobile landscape after:
IMG_5242

Mobile portrait before:
IMG_5248

Mobile portrait after:
IMG_5241

### DIFF
--- a/components/MainContentContainer.vue
+++ b/components/MainContentContainer.vue
@@ -17,10 +17,10 @@ import SearchResultDetails from './SearchResultDetails.vue';
             <Modal v-show="store.$state.isOpen" class="min-h-1/2 ml-8 mt-12">
                 <SearchResultDetails />
             </Modal>
-            <div class="h-1/2">
+            <div class="h-1/4">
                 <MapContainer />
             </div>
-            <div class="h-1/2 overflow-y-auto">
+            <div class="h-3/4">
                 <SearchResultsList />
             </div>
         </div>
--- a/components/SearchResultsList.vue
+++ b/components/SearchResultsList.vue
@@ -1,5 +1,5 @@
 <template>
-    <div class="h-full flex flex-col">
+    <div class="h-full flex flex-col overflow-y-auto">
         <div class="results-header flex flex-row ml-9 mr-5 mb-6 pt-5">
             <span class="flex-1 w-1/2 font-bold self-center">
                 {{ $t('searchResultsList.doctorsNearby') }}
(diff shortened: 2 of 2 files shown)


components/MainContentContainer.vue
```diff
@@ -17,10 +17,10 @@ import SearchResultDetails from './SearchResultDetails.vue';
             <Modal v-show="store.$state.isOpen" class="min-h-1/2 ml-8 mt-12">
                 <SearchResultDetails />
             </Modal>
-            <div class="h-1/2">
+            <div class="h-1/4">
                 <MapContainer />
             </div>
-            <div class="h-1/2 overflow-y-auto">
+            <div class="h-3/4">
```
ermish, line 23: Future idea: The map being smaller is fine for now, but we'll probably want it bigger when we can make the results more like an expanding drawer.

components/SearchResultsList.vue
```diff
@@ -38,13 +38,12 @@
                 </div>
             </div>
         </div>
-
         <div v-else>
-            <div class="h-full flex flex-col">
-                <SVGNoSearchResults class="p-12" />
-                <span class="text-primary-text pl-1 font-bold self-center group-hover:text-primary-text-inverted/50">{{
+            <div class="flex flex-col">
+                <SVGNoSearchResults class="portrait:px-4 h-12" />
```
ermish, line 43: Design nit: personally I think the svg is a little too small now on both landscape and portrait mode. Could we make it a little bigger? (Doesn't have to be as big as the original)

ermish (approved):
Added a potential change but not blocking 👍

components/SearchResultsList.vue
ermish: Looking on my phone, I think about double looks right. (Originally was like 4x)

theyokohamalife: @ermish If we make it bigger on landscape, the "No Results Found" text won't be visible, which I think is more important for the user to see. We could consider moving the text above the SVG.

ermish: Valid point. Let's swap it whenever you have time

PR #550, opened 2024-06-08
fixed failing cypress tests

Tch4lla:
Resolves #[Issue number]

## 🔧 What changed
In regards to moderation, Cypress doesn't have a '.tab()' command. I had to replace all tab commands to manually focus on the next input field in order for the error messages to be displayed.  I also needed to rewrite the structure of the clipboard function as cypress has a special way of dealing with the clipboard. 

In regards to the home test, due to the way that vue refreshes, I needed to add a small wait in order for the elements to appear in the dom 

## 🧪 Testing instructions

run cypress tests 

## 📸 Screenshots
pas test


theyokohamalife (changes requested):
I'm not sure from the PR description why we need this change. A package was installed that allows us to use tab() in tests. Is it not working as intended? Could you provide screenshots in the description?

Tch4lla:
> I'm not sure from the PR description why we need this change. A package was installed that allows us to use tab() in tests. Is it not working as intended? Could you provide screenshots in the description?

I reinstalled the plugin and tested it again. When running the tests, it never actually tabs and focuses on the next input value, which is what would trigger the error message to be displayed 

image

Tch4lla (commented):
If there is any more information needed please let me know

theyokohamalife (changes requested):
I ran the tests again and confirmed that the `.tab()` is working on my local machine. It's also passing on CI. I think we should figure out why it's not working on your computer instead of removing it.

https://github.com/ourjapanlife/findadoc-web/assets/31802656/44e4857f-4942-4042-a6d9-776363fe359b

cypress/e2e/moderationDashboard.cy.ts
```diff
@@ -42,20 +41,24 @@ describe(
...
-                    })
+                    cy.window().then((win) => {
+                        win.navigator.clipboard.readText().then((text) => {
+                            expect(text).to.eq('13243214');
+                        });
+                    });
 
                     // the timeout is to give time for the clipboard to be read
-                    clipboardResult.should("exist", 10000);
+                    // clipboardResult.should("exist", 10000);
```
theyokohamalife, line 61: Let's remove this instead of commenting it out.

cypress/e2e/moderationDashboard.cy.ts
```diff
@@ -1,5 +1,4 @@
 import "cypress-real-events";
-import "cypress-plugin-tab"
```
theyokohamalife, line 2: This package is needed in order for the `.tab()` to work.

cypress/e2e/moderationDashboard.cy.ts
```diff
@@ -121,48 +125,62 @@ describe('Moderation Facility Submission Form', () => {
         })
 
         it('should be display error messages', () => {
-
-            cy.get('[data-testid="submission-form-nameEn"]').find('input').type('立川中央病院').tab()
+            //there is no cy.tab(), so manually need to shift focus from one input to another
```
theyokohamalife, line 128: Let's remove this comment. Also please see the screen recording I shared of `.tab()` working with the plugin installed.

Seeing as this is passing on CI and only failing on your local machine, I think we should try to figure out why before making changes to these tests.

Tch4lla (commented):
Latest PR has updates to address these issues

ermish:
@theyokohamalife can you review this one again? there's overlap with your test fixes PR, so you'll have the most context

PR #556, opened 2024-06-11
chore: increase wait between submit tests

theyokohamalife:
## 🔧 What changed
Increases the wait time before each Submit validation test, because sometimes the elements are slow to load. This was failing on `main` for #534, but passed after re-running it.

image


NabbeunNabi (approved):
LGTM

PR #560, opened 2024-06-12
CHORE: adding nuxt support to typescript compiler and intellisense

ermish:
## 🔧 What changed
- Adding support for nuxt variables and plugins into typescript compiler. 
  - We should not have any more `$t()` errors!
  - import from `'#app'` should work now. We can get `vueApp` variable
 
## 🧪 Testing instructions
- app should still work! 
- you shouldn't see any more $t() errors

### DIFF
--- a/nuxt.config.js
+++ b/nuxt.config.js
@@ -83,6 +83,8 @@ export default defineNuxtConfig({
         ],
         link: [{ rel: 'icon', type: 'image/x-icon', href: '/assets/favicon.ico' }]
     },
+    // this helps the aliases like ~ or @ to start at '/' instead of '/src'.
+    srcDir: '',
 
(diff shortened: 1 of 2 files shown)


ermish:
splitting these out is a pain and taking forever..... trying to figure out what's missing that's causing it to fail @theyokohamalife

theyokohamalife:
> splitting these out is a pain and taking forever..... trying to figure out what's missing that's causing it to fail @theyokohamalife

@ermish It looks like CI is failing because the module in `./.nuxt` doesn't exist. This isn't a directory that we check into the repository, so I would check if it's being generated when the CI runs.

Screenshot 2024-06-13 at 10 50 42 PM

ermish:
duplicated by #566. Closing in favor of that one

PR #565, opened 2024-06-14
refactor: adjust screen resolutions and wait times for submit tests

theyokohamalife:
## 🔧 What changed
Currently, tests are failing assertions for certain elements—some that are not rendering, and some that are offscreen. I discovered that the screen size can change between tests if it's only specified in the `before()` and not the `beforeEach()`. This was causing failures when looking for specific elements in either portrait or landscape mode.

### Test with local dev server
1. Run `yarn dev`
2. Run `yarn cypress run`
3. Confirm that the tests are passing

Additionally:
1. Run `yarn cypress open`
2. Run each e2e test and confirm that the tests are running against the proper screen resolutions

### Test with prod server
1. Run `yarn prod:build`
2. Run `yarn prod:start`
3. Run `yarn cypress run` to confirm that the tests pass.
4. Follow the Additional steps above

### DIFF
--- a/cypress.config.ts
+++ b/cypress.config.ts
@@ -7,6 +7,4 @@ export default defineConfig({
         baseUrl: 'http://localhost:3000',
         testIsolation: false
     },
-    viewportHeight: 927,
-    viewportWidth: 1600
 })
--- a/cypress/e2e/about.cy.ts
+++ b/cypress/e2e/about.cy.ts
@@ -2,9 +2,17 @@ import enUS from '../../i18n/locales/en.json'
 
 describe('About page', () => {
     context('Desktop resolution', () => {
+        before(() => {
+            cy.visit("/about")
+            // This wait time is to give the page time to load from Prod when ran in CI.
+            cy.wait(3000)
+        })
+
         beforeEach(() => {
+            // The resolution is in the beforeEach() instead of before() to
+            // prevent Cypress from defaulting to other screen sizes between tests.
             cy.viewport(1920, 1080)
-            cy.visit('/about')
+            cy.wait(500)
         })
 
         it('shows the desktop top nav', () => {
@@ -51,9 +59,19 @@ describe('About page', () => {
     })
 
     context('Portrait mode', () => {
-        beforeEach(() => {
-            cy.viewport('iphone-5')
+        before(() => {
             cy.visit('/about')
+            // This wait time is to give the page time to load from Prod when ran in CI.
+            cy.wait(3000)
+        })
+
+        beforeEach(() => {
+            // The resolution is in the beforeEach() instead of before() to
+            // prevent Cypress from defaulting to other screen sizes between tests.
+
+            // An iPhone 5 screen resolution is used to test portrait mode.
+            cy.viewport(640, 1136)
+            cy.wait(500)
         })
 
         it('shows the hamburger component', () => {
@@ -74,6 +92,17 @@ describe('About page', () => {
             cy.get('[data-testid="member-title"]').should('exist')
             cy.get('[data-testid="member-linkedin"]').should('exist')
             cy.get('[data-testid="member-github"').should('exist')
+        })
+
+        it("Allows navigating to external links", () => {
+            cy.get('[data-testid="member-linkedin"]')
+                .first()
+                .should("have.attr", "href")
+                .and("include", "linkedin.com")
+            cy.get('[data-testid="member-github"]')
+                .first()
+                .should("have.attr", "href")
+                .and("include", "github.com")
         })
     })
 })
--- a/cypress/e2e/home.cy.ts
+++ b/cypress/e2e/home.cy.ts
@@ -132,6 +132,8 @@ describe("Visits the home page", () => {
         beforeEach(() => {
             // The resolution is in the beforeEach() instead of before() to
             // prevent Cypress from defaulting to other screen sizes between tests.
+
+            // An iPhone 5 screen resolution is used to test portrait mode.
             cy.viewport(640, 1136)
             cy.wait(500)
         })
--- a/cypress/e2e/home.cy.ts
+++ b/cypress/e2e/home.cy.ts
@@ -4,19 +4,22 @@ describe("Visits the home page", () => {
 
     context('Landscape mode', () => {
         before(() => {
-            cy.viewport(1920, 1080)
             cy.visit("/")
+            // This wait time is to give the page time to load from Prod when ran in CI.
+            cy.wait(3000)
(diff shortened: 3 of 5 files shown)


cypress.config.ts
```diff
@@ -7,6 +7,4 @@ export default defineConfig({
         baseUrl: 'http://localhost:3000',
         testIsolation: false
     },
-    viewportHeight: 927,
-    viewportWidth: 1600
```
theyokohamalife, line 11: This does not appear to be having an effect in tests.

cypress/e2e/moderationDashboard.cy.ts
```diff
@@ -4,151 +4,142 @@ import enUS from "../../i18n/locales/en.json"
...
+                                    lastName: 'Doe',
+                                    locale: 'en'
+                                }
+                            ],
+                            spokenLanguages: ['English'],
+                            degrees: ['MD'],
+                            specialties: ['General Practice'],
+                            acceptedInsurance: ['Insurance1'],
+                            facilityIds: ['1']
+                        }
```
theyokohamalife, line 109: I don't know if this is how the formatting should be after the linter is fixed or not. These were automatic formatting changes that happened after I fixed the tests in the same file below. Can remove these if it doesn't align with the linting rules.

ermish: I think I'll ignore for now. when it's working, it'll correct itself then

cypress.config.ts
ermish: it's duplicate, let's strip it out 👍

cypress/e2e/about.cy.ts
```diff
@@ -2,9 +2,17 @@ import enUS from '../../i18n/locales/en.json'
 
 describe('About page', () => {
     context('Desktop resolution', () => {
+        before(() => {
+            cy.visit("/about")
+            // This wait time is to give the page time to load from Prod when ran in CI.
+            cy.wait(3000)
```
ermish, line 8: adding waits like this as a solution is a really testing bad practice and also compounds as tests grow, however, I don't think this is the right time to build  more elegant loading approaches so this is good for now 👌

cypress/e2e/about.cy.ts
```diff
@@ -2,9 +2,17 @@ import enUS from '../../i18n/locales/en.json'
 
 describe('About page', () => {
     context('Desktop resolution', () => {
+        before(() => {
```
ermish, line 5: do we need before AND beforeeach? seems weird to have both 🤔

cypress/e2e/about.cy.ts
```diff
@@ -51,9 +59,17 @@ describe('About page', () => {
     })
 
     context('Portrait mode', () => {
-        beforeEach(() => {
-            cy.viewport('iphone-5')
+        before(() => {
             cy.visit('/about')
+            // This wait time is to give the page time to load from Prod when ran in CI.
+            cy.wait(3000)
```
ermish, line 65: do we really need 3 secs for every test group? it's gonna be so slow

cypress/e2e/home.cy.ts
```diff
@@ -119,43 +120,19 @@ describe("Visits the home page", () => {
                 cy.url().should("include", "/terms")
             })
         })
-
-
-        describe("Visits the about page", () => {
```
ermish, line 124: I think it's fine to remove most of this, but we should validate navigating to a link works. (especially since we generate our urls.)

cypress/e2e/about.cy.ts
```diff
@@ -51,9 +59,17 @@ describe('About page', () => {
...
+        })
+
+        beforeEach(() => {
+            // The resolution is in the beforeEach() instead of before() to
+            // prevent Cypress from defaulting to other screen sizes between tests.
+            cy.viewport(640, 1136)
```
ermish, line 71: can you add a comment to explain this resolution

cypress/e2e/home.cy.ts
```diff
@@ -177,59 +154,56 @@ describe("Visits the home page", () => {
 
         describe("Hamburger Menu tests", () => {
             it('shows the hamburger icon', () => {
-                //this conditionally renders based on screen orientation
+                // The hamburger menu conditionally renders based on screen orientation
```
ermish, line 157: before today's negotiations, I would have said these comment improvements aren't technically part of this PR 😆  but I think these are excellent refactors in the files you were working in ✨

ermish (changes requested):
yay, hopefully we can merge this soon! added a few comments

cypress/e2e/about.cy.ts
theyokohamalife: @ermish So for this specific context, we only need to visit `/about` once and all of the tests can run after it loads. It takes quite a while for it to fully render—probably because of the images and whatnot. I figured it was probably best to load the page once, wait for it to finish loading, and then use the `beforeEach()` just to make sure the resolution stays at 1920x1080 for each test. Open to suggestions on how to better implement this though!

cypress/e2e/about.cy.ts
theyokohamalife: @ermish It can take that long for the page to fully render when it's ran in the prod environment (like it's done by our GitHub actions). That's why it sometimes fails and sometimes doesn't. Maybe you might know how to do this, but we could shorten the wait time if the tests are running against the local dev server instead of prod. That would at least make them run more quickly while developing. I'm not sure how to check for that in Cypress and set the conditional.

cypress/e2e/about.cy.ts
ermish: can you add the same to the other ones too? ♻️

cypress/e2e/about.cy.ts
ermish: we can improve it in future commits 👍  I have some ideas. For now, let's get this in to fix the tests

cypress/e2e/about.cy.ts
ermish: let's chat in a future issue about how to optimize it

cypress/e2e/about.cy.ts
theyokohamalife: @ermish Good call! I added those comments in the latest fixup commits.

ermish (approved):
ship it!

cypress/e2e/home.cy.ts
ermish: can you add an issue for adding one? would be good beginner dev opportunity

PR #566, opened 2024-06-14
fix: fix typescript errors in the app

NabbeunNabi:
## 🔧 What changed

- Adding support for nuxt variables and plugins into typescript compiler.
- We should not have any more $t() errors!
- import from '#app' should work now. We can get vueApp variable

## 🧪 Testing instructions

- app should still work!
- you shouldn't see any more $t() errors

### DIFF
--- a/app.vue
+++ b/app.vue
@@ -26,7 +26,7 @@ import { useNuxtApp } from "#app";
 import { useRuntimeConfig } from '#imports'
 import { useModerationScreenStore } from "~/stores/moderationScreenStore"
 import { useRoute } from 'vue-router'
-const { $viewport } = useNuxtApp();
+const { $viewport } = useNuxtApp()
 
 const store = useModerationScreenStore()
 const route = useRoute()
--- a/app.vue
+++ b/app.vue
@@ -24,9 +24,9 @@
 import { initializeGqlClient } from './utils/graphql.js'
 import { useNuxtApp } from "#app";
 import { useRuntimeConfig } from '#imports'
-const { $viewport } = useNuxtApp();
 import { useModerationScreenStore } from "~/stores/moderationScreenStore"
 import { useRoute } from 'vue-router'
+const { $viewport } = useNuxtApp();
 
 const store = useModerationScreenStore()
 const route = useRoute()
@@ -34,9 +34,8 @@ const currentPath = route.path.replace("/", "")
 
 store.setEnableModerationPanelToTrue(currentPath)
 
-const useLocalAPI = useRuntimeConfig().public.NUXT_USE_LOCAL_API
-
-initializeGqlClient(useLocalAPI as string | undefined)
+const useLocalAPI = useRuntimeConfig().public.NUXT_USE_LOCAL_API as string | undefined
+initializeGqlClient(useLocalAPI)
 
 
 </script>
--- a/app.vue
+++ b/app.vue
@@ -36,7 +36,7 @@ store.setEnableModerationPanelToTrue(currentPath)
 
 const useLocalAPI = useRuntimeConfig().public.NUXT_USE_LOCAL_API
 
-initializeGqlClient(useLocalAPI)
+initializeGqlClient(useLocalAPI as string | undefined)
 
 
 </script>
--- a/components/ModDashboardLeftNavbar.vue
+++ b/components/ModDashboardLeftNavbar.vue
@@ -36,7 +36,8 @@
 </template>
 
 <script setup lang="ts">
-import { ref, Ref } from 'vue'
+import { ref} from 'vue'
+import type { Ref } from 'vue'
 import SVGNoteStackAddSvg from "../assets/icons/note-stack-add.svg"
 import SVGCheckBoxSvg from "../assets/icons/check-box.svg"
 import SVGDisabledByDefault from "../assets/icons/disabled-by-default.svg"
--- a/components/ModDashboardLeftNavbar.vue
+++ b/components/ModDashboardLeftNavbar.vue
@@ -36,8 +36,7 @@
 </template>
 
 <script setup lang="ts">
-import { ref} from 'vue'
-import type { Ref } from 'vue'
+import { ref, type Ref} from 'vue'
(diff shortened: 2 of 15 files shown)


app.vue
```diff
@@ -36,7 +36,7 @@ store.setEnableModerationPanelToTrue(currentPath)
 
 const useLocalAPI = useRuntimeConfig().public.NUXT_USE_LOCAL_API
 
-initializeGqlClient(useLocalAPI)
+initializeGqlClient(useLocalAPI as string | undefined)
```
ermish, line 39: move this up to when useLocalAPI is instantiated

ermish, line 39: move this up to when useLocalAPI is instantiated

components/ModDashboardLeftNavbar.vue
```diff
@@ -36,7 +36,8 @@
 </template>
 
 <script setup lang="ts">
-import { ref, Ref } from 'vue'
+import { ref} from 'vue'
+import type { Ref } from 'vue'
```
ermish, line 40: you can keep these in the same line, just put `type` in front of `Ref`  within the curlies

package.json
```diff
@@ -14,7 +14,7 @@
         "prod:start": "nuxi start",
         "prepare:ci": "nuxi prepare",
         "test": "yarn test:pinia --run && yarn test:e2e",
-        "test:pinia": "vitest",
+        "test:pinia": "nuxi prepare && nuxi generate && vitest",
```
ermish, line 17: I don't think we need this

tsconfig.json
```diff
@@ -30,5 +26,7 @@
     ],
     "exclude": [
         "node_modules"
-    ]
+    ],
+    
+    "extends": "./.nuxt/tsconfig.json"
```
ermish, line 31: let's move this to the top of the file. It makes it easier to the next reader to know we're building on rules from a base config.

components/ModEditSubmissionTopbar.vue
```diff
@@ -26,7 +26,8 @@
 <script setup lang="ts">
 import SVGCopyContent from '~/assets/icons/content-copy.svg'
 import SVGSuccessCheckMark from '~/assets/icons/checkmark-square.svg'
-import { Ref, ref } from 'vue';
+import {  ref } from 'vue';
```
ermish, line 29: ♻️

components/ModInputField.vue
```diff
@@ -41,4 +39,10 @@ const checkValidationOfInputField =(inputValue:string) => {
     isTheInputValueValid.value = props.inputValidationCheck(inputValue)
 }
 
+const handleInput = (event: Event) => {
```
ermish, line 42: not really related to this PR context, but I'm fine with improvements 😉

ermish: I guess unless the compiler is complaining

components/SearchBar.vue
```diff
@@ -51,12 +51,15 @@
...
-import { ref, Ref, watchEffect } from 'vue'
+import { ref,  watchEffect } from 'vue'
 import { useSearchResultsStore } from '~/stores/searchResultsStore.js'
 import { useLocationsStore } from '~/stores/locationsStore.js'
-import { useSpecialtiesStore, SpecialtyDisplayOption } from '~/stores/specialtiesStore.js'
+import { useSpecialtiesStore } from '~/stores/specialtiesStore.js'
 import { Locale, Specialty } from "~/typedefs/gqlTypes.js"
-import { useLocaleStore, LocaleDisplay } from '~/stores/localeStore.js'
+import { useLocaleStore } from '~/stores/localeStore.js'
+import type { LocaleDisplay } from '~/stores/localeStore.js'
```
ermish, line 60: ♻️

components/SubmissionForm.vue
```diff
@@ -93,7 +93,8 @@
 import { useSubmissionStore } from "~/stores/submissionStore"
 import { useLocaleStore } from "~/stores/localeStore"
 import * as validations from '../utils/formValidations'
-import { computed, ComputedRef } from 'vue'
+import { computed } from 'vue'
```
ermish, line 96: ♻️

package.json
ermish: oh n/m I guess we do.

ermish (changes requested):
it works! a few cleanups
this duplicates #560 so I'll close that one in favor of this

app.vue
ermish: best practice for types: 
Add `as` casts when you want to change the type to something different or declare a new variable. If the variable is supposed to be of that type, put it in that line instead so it's more visible that it was expected to be of that type.

components/ModInputField.vue
NabbeunNabi: The typescript compiler was complaining about the type so this was the best practice I could find to fix it

app.vue
```diff
@@ -24,18 +24,17 @@
 import { initializeGqlClient } from './utils/graphql.js'
 import { useNuxtApp } from "#app";
 import { useRuntimeConfig } from '#imports'
-const { $viewport } = useNuxtApp();
 import { useModerationScreenStore } from "~/stores/moderationScreenStore"
 import { useRoute } from 'vue-router'
+const { $viewport } = useNuxtApp();
```
ermish, line 29: there's a semicolon added

tsconfig.json
```diff
@@ -30,5 +27,5 @@
     ],
     "exclude": [
         "node_modules"
-    ]
+    ],
```
ermish, line 30: added unnecessary comma

ermish (approved):
looks good! :shipit:

components/ModInputField.vue
```diff
@@ -10,18 +10,15 @@
                 :placeholder='placeholder'
                 :required='required'
                 :value="inputValue"
-                @input="event => {
-                    inputValue = event.target.value
-                    store.setInputField(label, inputValue)
-                }"
+                @input="handleInput"
```
theyokohamalife, line 13: Nice! I was mentioning to @Anissa3005 that this should be moved to a function in the script.

PR #580, opened 2024-06-17
chore: update generated types on db schema

NabbeunNabi:
## 🔧 What changed

This updated the generated types based on the graphQl schema change.

## 🧪 Testing instructions
Locally build the production and see that the types are the same as seen in the PR using 

```
yarn prod:build
```


Anissa3005 (approved):
LGTM 🚀

PR #581, opened 2024-06-17
chore: remove defineProps import

NabbeunNabi:
## 🔧 What changed
I removed an unnecessary import due to it being a compiler macro

## 🧪 Testing instructions
The cypress tests will make sure that it still renders correctly


ermish (approved):
Launch it

PR #589, opened 2024-06-18
style: match styling of language select to searchbar

NabbeunNabi:
## 🔧 What changed
This was mentioned a while back by @Ntlja. She noticed that our selects were not consistent in styling. So this updates the `LocaleSelector` to use the same styling as the `SearchBar` in terms of padding. I kept the styling in the hamburger menu the same due to it being styled in a way that fits into the hamburger menu text sizing.

## 🧪 Testing instructions
Run `yarn dev` and confirm it matches the screenshots

## 📸 Screenshots

-   ### Before
image

Mobile:
image

-   ### After
image

Mobile:
image

### DIFF
--- a/components/LocaleSelector.vue
+++ b/components/LocaleSelector.vue
@@ -2,7 +2,7 @@
   <div class="w-24">
     <select
       v-model="selectedLocale"
-      class="rounded-full  w-full px-1 border-2 border-primary/60 py-1.5 drop-shadow-md text-primary-text bg-primary-bg hover:bg-primary-hover/10 transition-all"
+      class="rounded-full md:w-full px-2 py-1 md:px-1 md:py-1.5 border-2 border-primary/80 md:border-primary/60  drop-shadow-md text-primary-text bg-secondary-bg/5 md:bg-primary-bg hover:bg-primary-hover/10 transition-colors"
     >
       <option
         v-for="(localeOption) in localeStore.mvpLocaleDisplayOptions"
@@ -28,3 +28,6 @@ watch(selectedLocale, (newLocale: any) => {
   locale.value = newLocale;
 });
 </script>
+
+
+rounded-full px-2 py-1 border-2 border-primary/80 drop-shadow-md text-primary-text bg-secondary-bg/5 hover:text-primary-hover transition-colors
--- a/components/LocaleSelector.vue
+++ b/components/LocaleSelector.vue
@@ -2,7 +2,7 @@
   <div class="w-24">
     <select
       v-model="selectedLocale"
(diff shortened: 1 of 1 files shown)


components/LocaleSelector.vue
```diff
@@ -2,7 +2,7 @@
   <div class="w-24">
     <select
       v-model="selectedLocale"
-      class="rounded-full px-2 py-1 border-2 border-primary/80 drop-shadow-md text-primary-text bg-secondary-bg/5 hover:text-primary-hover transition-colors"
+      class="rounded-full md:w-full px-2 py-1 md:px-1 md:py-1.5 border-2 border-primary/80 md:border-primary/60  drop-shadow-md text-primary-text bg-secondary-bg/5 md:bg-primary-bg hover:bg-primary-hover/10 transition-colors"
```
ermish, line 5: I need to add documentation around this, but we aren't using `md:` anymore in the project. Instead use `portrait:` or `landscape:`. 
This let's us design around device orientation and we don't have to have custom resolutions for every new device  that comes out.

NabbeunNabi: @ermish should I be writing it as if portrait is the default? Or landscape is the default?

ermish: that's a fair question we haven't decided yet 🤔 
we've been building as desktop mode first, but maybe we should default the mobile view....we don't know what our user base is yet. 
@theyokohamalife you up for mobile first from now on? (obviously we need to add documentation around this eventually)

NabbeunNabi: Just my two cents for @ermish and @theyokohamalife before you make the decision. The project uses tailwind which in the documentation is [mobile first](https://tailwindcss.com/docs/responsive-design#working-mobile-first ). So I would suggest that we default to portrait and then change to landscape. Just due to the css preprocessor we are using and the way it was designed.

ermish: sounds good to me! @theyokohamalife ?

components/LocaleSelector.vue
```diff
@@ -28,3 +28,6 @@ watch(selectedLocale, (newLocale: any) => {
   locale.value = newLocale;
 });
 </script>
+
+
+rounded-full px-2 py-1 border-2 border-primary/80 drop-shadow-md text-primary-text bg-secondary-bg/5 hover:text-primary-hover transition-colors
```
ermish, line 33: uhhh

ermish: can we delete this random css string floating in the void?

ermish (changes requested):
two fixes then it's in!

ermish (approved):
LGTM

PR #564, opened 2024-06-14
Added the portuguese json file and add the option

Diego-Bittencourt:
## 🔧 What changed
- Added the portuguese json file with all text in portuguese.

- Add the portuguese json file in the index.ts file to export it.

- Added the portuguese option in the mvpLocaleDisplayOptions type, in the localeStore.

- Changed from "Portuguese" to "Português" and switched from "Brazil" to "Brasil" in the localeDisplayOptions type in the store.

## 🧪 Testing instructions

Access the application, change the language settings to "Português" and note the text changes to words in portuguese.

## 📸 Screenshots

-   ### Before
image

-   ### After
image

PR #595, opened 2024-06-19
implement fade in photo component

NabbeunNabi:
This component can be used anywhere in the app. It allows for a smoother load of images no matter the size. Even after compressing the about us files they would sometimes load in a manner that was less than ideal. Now it loads with a fade in effect. Tests were refactored to be updated for the component.

## 🔧 What changed
Before the loading image on the about us page was a bit choppy

## 🧪 Testing instructions
Go to the branch and run: 
```
yarn dev
yarn cypress open
```
Run the about specs

## 📸 Screenshots

-   ### Before
https://www.loom.com/share/d1249542d49c4113934080f20ef8ffd2

-   ### After
https://www.loom.com/share/5dd53ead6c424cf2b51ebb518feaff68

### DIFF
--- a/components/MemberComponent.vue
+++ b/components/MemberComponent.vue
@@ -54,7 +54,6 @@
 import SVGLinkedinIcon from '~/assets/icons/social-linkedin.svg'
 import SVGGithubIcon from '~/assets/icons/social-github.svg'
 import type { PropType } from "vue"
-import SmoothLoadingImage from './SmoothLoadingImage.vue';
 
 defineProps({
     avatarImg: {
--- a/components/MemberComponent.vue
+++ b/components/MemberComponent.vue
@@ -1,9 +1,13 @@
 <template>
     <div class="member flex flex-col items-center text-center">
-        <img
-            data-testid="member-avatar"
-            :src="avatarImg"
-            class="member-image object-cover rounded-full w-32 h-32"
+        <SmoothLoadingImage
+            :imgSrc="avatarImg"
+            :testId="`member-avatar-${index}`"
+            :alt="name"
+            object-styling="cover"
+            height="32"
+            width="32"
+            borderRadius="full"
         />
         <div
             data-testid="member-name"
@@ -50,11 +54,16 @@
 import SVGLinkedinIcon from '~/assets/icons/social-linkedin.svg'
 import SVGGithubIcon from '~/assets/icons/social-github.svg'
 import type { PropType } from "vue"
+import SmoothLoadingImage from './SmoothLoadingImage.vue';
 
 defineProps({
     avatarImg: {
         type: String,
         required: true,
+    },
+    index: {
+        type: Number,
+        require: true
     },
     name: {
         type: String,
@@ -71,6 +80,6 @@ defineProps({
     githubUrl: {
         type: [String, null] as PropType<string | null>,
         required: false,
-    },
+    }
 })
 </script>
--- a/components/SmoothLoadingImage.vue
+++ b/components/SmoothLoadingImage.vue
@@ -1,7 +1,7 @@
 <template>
   
-    <SmoothLoadingImageFadeTransition><div v-show="!show.showing"><SVGLoadingIcon /></div></SmoothLoadingImageFadeTransition>
-   <SmoothLoadingImageFadeTransition><img v-show="show.showing" :data-testid=testId :class="`object-${objectStyling} rounded-${borderRadius} w-${width} h-${height}`" :src="src" @load="showImg()"/></SmoothLoadingImageFadeTransition>
+    <SmoothLoadingImageFadeTransition><div v-show="!isVisible"><SVGLoadingIcon /></div></SmoothLoadingImageFadeTransition>
+   <SmoothLoadingImageFadeTransition><img v-show="isVisible" :data-testid=testId :class="`object-${objectStyling} rounded-${borderRadius} w-${width} h-${height}`" :src="src" @load="showImg()"/></SmoothLoadingImageFadeTransition>
 
 </template>
 
@@ -41,16 +41,16 @@
   })
   
   const src:Ref<string | undefined> = ref()
-  const show = reactive({showing: false});
+  const isVisible:Ref<boolean> = ref(false)
   const showImg = function () {
-    show.showing = true
+    isVisible.value = true
   }
   
   watch(
     () => props.imgSrc,
     () => {
     src.value = props.imgSrc
-    show.showing = false
+    isVisible.value = false
   }, {immediate: true})
   
 
--- a/components/SmoothLoadingImage.vue
+++ b/components/SmoothLoadingImage.vue
@@ -1,7 +1,7 @@
 <template>
   
     <SmoothLoadingImageFadeTransition><div v-show="!show.showing"><SVGLoadingIcon /></div></SmoothLoadingImageFadeTransition>
-   <SmoothLoadingImageFadeTransition><img v-show="show.showing" :class="`object-${objectStyling} rounded-${borderRadius} w-${width} h-${height}`" :src="src" @load="showImg()"/></SmoothLoadingImageFadeTransition>
+   <SmoothLoadingImageFadeTransition><img v-show="show.showing" :data-testid=testId :class="`object-${objectStyling} rounded-${borderRadius} w-${width} h-${height}`" :src="src" @load="showImg()"/></SmoothLoadingImageFadeTransition>
 
 </template>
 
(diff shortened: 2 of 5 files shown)


components/MemberComponent.vue
```diff
@@ -1,9 +1,13 @@
 <template>
     <div class="member flex flex-col items-center text-center">
-        <img
-            data-testid="member-avatar"
-            :src="avatarImg"
-            class="member-image object-cover rounded-full w-32 h-32"
+        <SmoothLoadingImage
```
NabbeunNabi, line 3: These props allow you to customize the image to the place you want to use this component since we might have more assets in the future

cypress/e2e/about.cy.ts
```diff
@@ -45,7 +45,7 @@ describe('About page', () => {
         })
 
         it('shows member details', () => {
-            cy.get('[data-testid="member-avatar"]').should('exist')
+            cy.get('[data-testid="member-avatar-0"]').should('exist')
```
NabbeunNabi, line 48: This tests each avatar of the members their own  unique id and can also be tested in other files when images are used in the future using the prop.

cypress/e2e/about.cy.ts
```diff
@@ -87,7 +87,7 @@ describe('About page', () => {
...
```
NabbeunNabi, line 90: ♻️

components/SmoothLoadingImage.vue
```diff
@@ -0,0 +1,77 @@
...
+    src.value = props.imgSrc
+    show.showing = false
+  }, {immediate: true})
+  
+
+</script>
+
+
+<style scoped>
+  .loader {
```
ermish, line 61: move these to tailwind style!  🐈 💨  (closest I can make to a tail-wind emoji XD)

components/SmoothLoadingImage.vue
```diff
@@ -0,0 +1,77 @@
...
+        required: true,
+    },
+    borderRadius:{
+        type: String,
+        required: true,
+    },
+  })
+  
+  const src:Ref<string | undefined> = ref()
+  const show = reactive({showing: false});
```
ermish, line 44: reactive is good for objects with more than one field (like a whole entity where the fields are all related.  ex. const person = { .... }
we should simplify this to just `isVisible = ref(false)`

components/SmoothLoadingImageFadeTransition.vue
```diff
@@ -0,0 +1,20 @@
...
+</script>
+
+<template>
+  <Transition name="fade">
+    <slot></slot> 
+  </Transition>
+</template>
+
+<style>
+  .fade-enter-active,
```
ermish, line 11: I'm flipping back and forth if we really need a separate component. I can see the value if we plan to have more effects than just the fade, but we probably won't. 
According to "best practices!": only build for resubility if you know you are going to need it or you already have 3 or more recurring usages. 

also....we could make it conditional css instead of a component as a design alternative 🤔

pages/about.vue
```diff
@@ -32,7 +32,7 @@
                 <div data-testid="member" class="members-list grid"
                     :key="index" v-for="(member, index) in data.members">
                     <MemberComponent :avatarImg="member.avatarImg" :name="member.name" :title="member.title"
-                        :linkedInUrl="member.linkedInUrl" :githubUrl="member.githubUrl" />
+                        :index=index :linkedInUrl="member.linkedInUrl" :githubUrl="member.githubUrl" />
```
ermish, line 35: it's usually recommended to avoid index as the index if possible. Usually I just pick a unique field on the object as the key.

pages/about.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-    <div class="overflow-y-auto w-full">
```
ermish, line 2: I think @theyokohamalife wants to keep this in

ermish (changes requested):
love the upgrade!

pages/about.vue
theyokohamalife: Correct! This is to make sure the footer is stickied at the bottom of the page.

theyokohamalife (changes requested):
Weird things are happening with the image proportions that you can see in the preview.
Screenshot 2024-06-19 at 4 06 04 PM

components/SmoothLoadingImageFadeTransition.vue
NabbeunNabi: You are the one that had me make a separate `MemeberComponent` for the about us page. That is used only in the About us page. So I was following your previous instructions. The previous standups doctors were wondering if they could use the photo space in the modal. Thought this could be used there as well.

pages/about.vue
NabbeunNabi: I was passing this index down so it can be used in the test in the `data-testid` the way it is used in the `SubmissionListComponent`. I could add an id into the json?

pages/about.vue
NabbeunNabi: To me this should stay as it creates two y-axis scrollbars and css should be used to make the footer sticky in the footer component.

On the production site it isn't sticky due to the main scrollbar existing.

image

theyokohamalife: @NabbeunNabi That is a bug caused by #564 and noted in issue  #596. Let's fix that first and exclude this for now. 🙏🏾

components/SmoothLoadingImage.vue
```diff
@@ -0,0 +1,77 @@
+<template>
+  
+    <SmoothLoadingImageFadeTransition><div v-show="!isVisible"><SVGLoadingIcon /></div></SmoothLoadingImageFadeTransition>
+   <SmoothLoadingImageFadeTransition><img v-show="isVisible" :data-testid=testId :class="`object-${objectStyling} rounded-${borderRadius} w-${width} h-${height}`" :src="src" @load="showImg()"/></SmoothLoadingImageFadeTransition>
```
Anissa3005, line 4: Linting Police 🚓

- This lines have weird indentation (maybe it's only on Github but make sure the indentation is correct)
- On line for the line is to long and the attributes should be under each other for better readability like this:
```
    <SVGHamburgerMenuIcon
      data-testid="hamburger-menu-icon"
      alt="hamburger menu"
      class="h-8 w-8"
      @click="openMenu()"
    />
```  

Other example: 
```
    <button
      data-testid="hamburger-menu-close-button"
      class="px-2 py-.5 mr-2"
      @click="closeMenu()"
    >
      <svg
        class="stroke-primary"
        width="20"
        heigh="20"
        viewBox="4 0 15 25"
      >
        <path
          stroke-width="3"
          fill="none"
          d="M6.25,6.25,17.75,17.75"
        />
        <path
          stroke-width="3"
          fill="none"
          d="M6.25,17.75,17.75,6.25"
        />
      </svg>
    </button>
```

components/SmoothLoadingImage.vue
```diff
@@ -0,0 +1,77 @@
...
+        required: true,
+    },
+    width:{
+        type: String,
+        required: true,
+    },
+    borderRadius:{
+        type: String,
+        required: true,
+    },
```
Anissa3005, line 40: Styling Police 🚓🖌️

No trailing commas

components/SmoothLoadingImage.vue
```diff
@@ -0,0 +1,77 @@
...
+  }, {immediate: true})
+  
+
+</script>
+
+
+<style scoped>
+  .loader {
+    border: 4px solid grey;
+    border-top: 4px solid darkGrey;
```
Anissa3005, line 63: Styling Police 🚓🖌️

No SEMICOLON don't be a Philip

components/SmoothLoadingImageFadeTransition.vue
```diff
@@ -0,0 +1,20 @@
...
+<template>
+  <Transition name="fade">
+    <slot></slot> 
+  </Transition>
+</template>
+
+<style>
+  .fade-enter-active, 
+  .fade-leave-active {
+    transition: opacity 1s ease;
```
Anissa3005, line 13: Styling Police 🚓🖌️

No SEMICOLON

components/SmoothLoadingImageFadeTransition.vue
```diff
@@ -0,0 +1,20 @@
...
+  }
+
+  .fade-enter-from, 
+  .fade-leave-to {
+    opacity: 0;
```
Anissa3005, line 18: Styling Police 🚓🖌️

No SEMICOLON

components/MemberComponent.vue
```diff
@@ -1,9 +1,13 @@
...
+            :imgSrc="avatarImg"
```
Anissa3005, line 4: This part is PERFECT linting/styling, if you can do this for your SmoothLoadingImage component as well would be great ✨

Anissa3005 (changes requested):
Some linting/styling requests 🚨

PR #610, opened 2024-06-25
feat: implement autofill for counts

### LINKED ISSUE #509: Implement the correct count in the `ModDashboardLeftNavbar`
# Problem
Right now the left navbar is hardcoded at 0. 

Image

# Success
It should show the correct number for each choice based on the data fetched. 
### Requirements

- [ ] Get the data from the store and show the count

# Implementation notes
Along with issue #508  a lot of this logic could be handled in the store. Then called in the components

NabbeunNabi:
Resolves #509 
In the submission dashboard there is a count for how many of each type of submission there are based on its status in the database for approved, rejected, or under review. Now it is dynamically calculated and tested based on the Boolean from the database.
## 🧪 Testing instructions
You can go to the branch and run:
```
yarn dev
yarn cypress open
```
From here you should run the `moderationDashboard.cy.ts` specs. They all should pass.

## 📸 Screenshots

-   ### Before
image

-   ### After
image

### DIFF
--- a/components/ModDashboardLeftNavbar.vue
+++ b/components/ModDashboardLeftNavbar.vue
@@ -19,7 +19,7 @@
                     class="flex flex-row items-center text-start p-1"
                     @click="updateSubmissionListViewState('forReview')"
                 >
-                    {{ $t("modDashboardLeftNav.forReview") }} ({{ 0 }})
+                    {{ $t("modDashboardLeftNav.forReview") }} ({{ autofillStatusCount.forReviewCount }})
                 </button>
             </div>
             <div class="flex w-4/5 justify-start items-center border-b-2 border-slate-200">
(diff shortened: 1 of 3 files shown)


components/ModDashboardLeftNavbar.vue
```diff
@@ -63,4 +67,34 @@ const updateSelectedDashboardView = (event: Event) => {
     const selectedValue = (event.target as HTMLSelectElement).value
     selectedDashboardView.value = selectedValue
 }
+
+interface SubmissionStatusDisplayCount {
```
ermish, line 71: I'm seeing some good naming going on! 💯

components/ModDashboardLeftNavbar.vue
```diff
@@ -63,4 +67,34 @@ const updateSelectedDashboardView = (event: Event) => {
...
+}
+
+function createCountsForSubmissionListView(submissions: Submission[]) {
+    const statusCountObject: SubmissionStatusDisplayCount = {
+        forReviewCount: 0,
+        approvedCount: 0,
+        rejectedCount: 0
+    }
+
+    submissions.forEach((submission: Submission) => {
```
ermish, line 84: You can actually use reduce or map for this instead of for each

ermish (changes requested):
Nice update! Like the choice for where to generate the counts

components/ModDashboardLeftNavbar.vue
NabbeunNabi: updated to use reduce

ermish (approved):
Commencing Production delivery 🚀

PR #611, opened 2024-06-25
feat: create logic for fetching healthcare professionals

### LINKED ISSUE #530: FEAT: Fetch healthcare professionals
# Problem

This issue relies on #529 

We need to create a fetch request to retrieve a healthcare professional by full name and locale. This will allow us to associate existing healthcare professionals with new facilities.

### Requirements

- [x] Create a store to manage our fetching logic.
- [x] Implement the fetching logic and GraphQL query for retrieving a healthcare professional by full name and locale.
- [x] Verify the data by logging it to the console (for now).

# Implementation notes
- The display of the healthcare professional data will be handled in a separate PR to keep changes small and manageable.

Anissa3005:
Resolves #530 

## 🔧 What changed

**❗Changes Made Against Design ❗**

I have decided to only implement search a healthcare professional by id only, this is why;

At the moment we don't have a fuzzy search in the backend. This means that we need to search by exact name type like following: 
```
type LocalizedName {
  firstName: String!
  middleName: String
  lastName: String!
  locale: Locale!
}
```
So if you would make a request it should exactly match those fields. In my opinion I think it's better to just use the id for now. If we have a fuzzy search in the future we can implement searching by name.

- Fetch the healthcare professionals for the ModSearchHealthcareProfessional component
- Create graphQl query
- Create a moderationSearchHealthcareProfessionalStore

## To do
- Implement the UI for this part

### DIFF
--- a/components/ModHealthcareProfessionalSearchbar.vue
+++ b/components/ModHealthcareProfessionalSearchbar.vue
@@ -6,10 +6,12 @@
         <div class="flex flex-col mt-4">
             <div class="justify-start items-start flex">
                 <input
+                    v-model="requestedHealthcareProfessionalId"
                     type="text"
                     :placeholder="$t('modSubmissionForm.placeholderTextHealthcareProfessionalSearchbar')"
                     class="mb-5 px-3 py-3.5 w-96 h-12 bg-secondary-bg rounded-lg border border-primary-text-muted
                     text-primary-text text-sm font-normal font-sans placeholder-primary-text-muted"
+                    @input="searchHealthcareProfessionalById"
                 >
                 <SVGLookingGlass
                     role="img"
@@ -22,5 +24,19 @@
 </template>
 
 <script lang="ts" setup>
+import { ref, type Ref } from 'vue'
 import SVGLookingGlass from '~/assets/icons/looking-glass.svg'
+import type { HealthcareProfessional } from '~/typedefs/gqlTypes'
+import { getHealthcareProfessionalById } from '~/stores/healthcareProfessionalsStore'
+
+const requestedHealthcareProfessionalId: Ref<string> = ref('')
+// `healthcareProfessionalById` will be used in the UI component in an upcoming PR
+const healthcareProfessionalById: Ref<HealthcareProfessional[] | []> = ref([])
+
+async function searchHealthcareProfessionalById() {
+    if (!requestedHealthcareProfessionalId.value || requestedHealthcareProfessionalId.value.trim().length < 1) {
+        return
+    }
+    healthcareProfessionalById.value = await getHealthcareProfessionalById(requestedHealthcareProfessionalId.value)
+}
 </script>
--- a/stores/healthcareProfessionalsStore.ts
+++ b/stores/healthcareProfessionalsStore.ts
@@ -52,6 +52,26 @@ async function queryHealthcareProfessionals() {
     }
 }
 
+export async function getHealthcareProfessionalById(id: string) {
+    try {
+        const queryData = {
+            healthcareProfessionalId: id
+        }
+        const result = await gqlClient.request<{ healthcareProfessional: HealthcareProfessional[] }>(
+            getHealthcareProfessionalByIdGqlQuery,
+            queryData
+        )
+
+        if (!result.healthcareProfessional) {
+            throw new Error(`The Healthcare Professional ID doesn't exist`)
+        }
+        return result.healthcareProfessional
+    } catch (error: unknown) {
+        console.log(`Error retrieving healthcare professional by id: ${id}: ${JSON.stringify(error)}`)
+        return []
+    }
+}
+
 const getAllHealthcareProfessionalsData = gql`
 query Query($filters: HealthcareProfessionalSearchFilters!) {
   healthcareProfessionals(filters: $filters) {
@@ -71,6 +91,19 @@ query Query($filters: HealthcareProfessionalSearchFilters!) {
     updatedDate
   }
 }`
+
+const getHealthcareProfessionalByIdGqlQuery = gql`
+    query HealthcareProfessionals($healthcareProfessionalId: ID!) {
+        healthcareProfessional(id: $healthcareProfessionalId) {
+        names {
+            firstName
+            lastName
+        }
+        id
+        specialties
+    }
+}`
+
 
 const updateHealthcareProfessionalGqlMutation = gql`
 mutation Mutation($updateHealthcareProfessionalId: ID!, $input: UpdateHealthcareProfessionalInput!) {
--- a/stores/healthcareProfessionalsStore.ts
+++ b/stores/healthcareProfessionalsStore.ts
@@ -104,7 +104,6 @@ const getHealthcareProfessionalByIdGqlQuery = gql`
     }
 }`
 
-
 const updateHealthcareProfessionalGqlMutation = gql`
 mutation Mutation($updateHealthcareProfessionalId: ID!, $input: UpdateHealthcareProfessionalInput!) {
   updateHealthcareProfessional(id: $updateHealthcareProfessionalId, input: $input) {


theyokohamalife:
@Anissa3005 Is this PR still a WIP?

Anissa3005:
@theyokohamalife Yes, we needed to implemented something first in the backend to make this work. The backend implementation is merged yesterday, so now working on this again!

components/ModHealthcareProfessionalSearchbar.vue
```diff
@@ -22,5 +24,20 @@
...
 <script lang="ts" setup>
+import { ref, type Ref } from 'vue'
 import SVGLookingGlass from '~/assets/icons/looking-glass.svg'
+import { useModerationSearchHealthcareProfessionalsStore } from '~/stores/moderationSearchHealthcareProfessionalsStore'
+
+const id: Ref<string> = ref('')
+
+const modSearchHealthcareProfessionalStore = useModerationSearchHealthcareProfessionalsStore()
+// This code is required for implementing a feature in HTML. It's currently commented out due to ESLint rules.
+// const healthcareProfessional = computed(() => modSearchHealthcareProfessionalStore.healthcareProfessional)
```
theyokohamalife, line 35: Is this something that should be fixed before this gets merged? We should probably avoid commenting out code a feature is dependent on due to ESLint.

Rather than commenting it out, you may consider leaving a comment for this line on your PR with the ESLint error message so reviewers can pitch in with suggestions.

stores/moderationSearchHealthcareProfessionalsStore.ts
```diff
@@ -0,0 +1,49 @@
...
+async function queryHealthcareProfessionalById(id: string) {
+    try {
+        const healthcareProfessionalId = {
+            healthcareProfessionalId: id
+        }
+        const result = await gqlClient.request<{ healthcareProfessional: HealthcareProfessional[] }>(
+            getHealthcareProfessionalsGqlQuery,
+            healthcareProfessionalId
+        )
+        return result.healthcareProfessional ?? []
```
theyokohamalife, line 32: I'd like to know what @ermish thinks, but I think we should return something other than an empty list here. Empty lists are usually used when you expected multiple items to be returned, but there were none. In this case we were only expecting one item.

stores/moderationSearchHealthcareProfessionalsStore.ts
```diff
@@ -0,0 +1,49 @@
...
+    } catch (error: unknown) {
+        console.log(`Error querying the healthcare professionals: ${JSON.stringify(error)}`)
```
theyokohamalife, line 34:
```suggestion
        console.log(`Error querying healthcare professional ${id}: ${JSON.stringify(error)}`)
```
Eventually when we have a logging service it could be really helpful to include the `id` of the healthcare professional that caused the error.

components/ModHealthcareProfessionalSearchbar.vue
Anissa3005: This logic will be necessary for the person that will implement the UI for this part.

The problem is if a variable is unused it cause an eslint error, so this logic I had to comment out otherwise my eslint wouldn't pass. But this code  is necessary for when UI is implemented.

stores/moderationSearchHealthcareProfessionalsStore.ts
```diff
@@ -0,0 +1,49 @@
...
+        }
+
+        return {
+            getHealthcareProfessionals,
+            healthcareProfessional
+        }
+    }
+)
+
+async function queryHealthcareProfessionalById(id: string) {
```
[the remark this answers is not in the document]
Anissa3005: In the other stores we use this naming like `queryProfessionals` and `queryFacilities` so I think It's better to use the same naming as the other stores to keep it consistent.

components/ModHealthcareProfessionalSearchbar.vue
```diff
@@ -22,5 +24,20 @@
...
+import { useModerationSearchHealthcareProfessionalsStore } from '~/stores/moderationSearchHealthcareProfessionalsStore'
+
+const healthcareProfessionalId: Ref<string> = ref('')
+
+const modSearchHealthcareProfessionalStore = useModerationSearchHealthcareProfessionalsStore()
+// This code is required for implementing a feature in HTML. It's currently commented out due to ESLint rules.
+// const healthcareProfessional = computed(() => modSearchHealthcareProfessionalStore.healthcareProfessional)
+
+async function searchHealthcareProfessionalById() {
+    if (healthcareProfessionalId.value.trim().length < 1) {
```
ermish, line 38: This will break if the trim is called on undefined

stores/moderationSearchHealthcareProfessionalsStore.ts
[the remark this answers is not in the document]
ermish: I think @theyokohamalife has a point that they're overlapping. Getbyid is very specific which is good. I recommend using the term query or search for methods that can have different search conditions. (Id is not really a search for example)

stores/moderationSearchHealthcareProfessionalsStore.ts
ermish: I agree. For getbyid it should return the value. 
If it doesn't exist, usually returning undefined would be appropriate for a single item. 

throwing an exception is also a valid option if we expect to always have a response.

components/ModHealthcareProfessionalSearchbar.vue
```diff
@@ -22,5 +24,20 @@
 </template>
 
 <script lang="ts" setup>
+import { ref, type Ref } from 'vue'
 import SVGLookingGlass from '~/assets/icons/looking-glass.svg'
+import { useModerationSearchHealthcareProfessionalsStore } from '~/stores/moderationSearchHealthcareProfessionalsStore'
+
+const healthcareProfessionalId: Ref<string> = ref('')
```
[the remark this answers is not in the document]
theyokohamalife: @ermish Good point! It was my suggestion to change it from `id` to `healthcareProfessionalId`. For this particular component I believe it'll only be used to to search for a healthcare professional. I don't think it'll be used to mutate the healthcare professional at all.

stores/moderationSearchHealthcareProfessionalsStore.ts
```diff
@@ -0,0 +1,49 @@
+import { gql } from 'graphql-request'
```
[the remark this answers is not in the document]
theyokohamalife: Hmm just a thought, but do we need this store? We already have a `healthcareProfessionalsStore` that these could be added to.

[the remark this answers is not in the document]
Anissa3005: It's tied to this:

image

At the moment the functionality how it's designed in the UI is not yet possible. But I can see that this would change in the future. So with that I thought for long term.

@theyokohamalife This only would be possible if we add an Id filter, what I actually wanted to ask about. Would it not be cleaner if we just have one searchHealthcareProfessional and add the id as filter in the backend?

components/ModHealthcareProfessionalSearchbar.vue
Anissa3005: Can this be undefined? Since healthcareProfessionalId (name already changed) is a ref set to an empty string, will the value not always be a string value?

stores/moderationSearchHealthcareProfessionalsStore.ts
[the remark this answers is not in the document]
theyokohamalife: I guess what I'd propose is defining a `getHealthcareProfessionalById` function in the `healthcareProfessionalsStore.ts`. I think that we should write our queries and mutations in a way that they aren't tied to specific components and can be used throughout the app.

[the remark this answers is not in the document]
ermish: > I guess what I'd propose is defining a getHealthcareProfessionalById function in the healthcareProfessionalsStore.ts. I think that we should write our queries and mutations in a way that they aren't tied to specific components and can be used throughout the app.

this makes sense to me. Your point @Anissa3005 that we MAY have additional functions for moderation in the future, is a good design example where we shouldn't build for something that may not happen. The reason is if it doesn't, we've overdesigned/added complexity. It's a habit that when multiplied x50 devs, the codebase becomes more complex than it needs. You can always split it out later into it's own section later, but it's hard to do the opposite direction and consolidate.

components/ModHealthcareProfessionalSearchbar.vue
ermish: in theory it could be cleared out since it's the selected id and not always pulled from a healthcareprof object so there's a risk even if it's unlikely

components/ModHealthcareProfessionalSearchbar.vue
```diff
@@ -22,5 +24,18 @@
 </template>
 
 <script lang="ts" setup>
+import { ref, type Ref } from 'vue'
 import SVGLookingGlass from '~/assets/icons/looking-glass.svg'
+import { useSearchHealthcareProfessionalStore } from '~/stores/searchHealthcareProfessionalStore'
+
+const requestedHealthcareProfessionalId: Ref<string> = ref('')
+
+const modSearchHealthcareProfessionalStore = useSearchHealthcareProfessionalStore()
```
ermish, line 33: need to update this name to match

stores/searchHealthcareProfessionalStore.ts
```diff
@@ -0,0 +1,53 @@
+import { gql } from 'graphql-request'
+import { defineStore } from 'pinia'
+import { type Ref, ref } from 'vue'
+import { gqlClient } from '../utils/graphql.js'
+import type { HealthcareProfessional } from '~/typedefs/gqlTypes'
+
+export const useSearchHealthcareProfessionalStore = defineStore(
+    'SearchHealthcareProfessionalStore',
```
ermish, line 8: let's make this camel case to match the other stores
```suggestion
    'searchHealthcareProfessionalStore',
```

components/ModHealthcareProfessionalSearchbar.vue
```diff
@@ -22,5 +24,18 @@
...
+
+async function searchHealthcareProfessionalById() {
+    if (requestedHealthcareProfessionalId.value.trim().length < 1) {
+        return
+    }
+    await modSearchHealthcareProfessionalStore.getHealthcareProfessional(requestedHealthcareProfessionalId.value)
```
ermish, line 39: bug: we're not actually returning any value here to set the id

stores/searchHealthcareProfessionalStore.ts
```diff
@@ -0,0 +1,53 @@
...
+    () => {
+        const healthcareProfessional: Ref<HealthcareProfessional[]> = ref([])
```
ermish, line 10: I don't think we actually need ` getHealthcareProfessional` or `healthcareProfessional`. We can just expose `getHealthcareProfessionalById` and let the caller use the value to set the ref in the component

ermish (changes requested):
few updates

components/ModHealthcareProfessionalSearchbar.vue
```diff
@@ -22,5 +24,18 @@
 </template>
 
 <script lang="ts" setup>
+import { ref, type Ref } from 'vue'
 import SVGLookingGlass from '~/assets/icons/looking-glass.svg'
+import type { HealthcareProfessional } from '~/typedefs/gqlTypes'
+import { getHealthcareProfessionalById } from '~/stores/healthcareProfessionalsStore'
+
+const requestedHealthcareProfessionalId: Ref<string> = ref('')
+const healthcareProfessionalById: Ref<HealthcareProfessional[] | []> = ref([])
```
ermish, line 33: is this being used by anything? I think we can remove it

stores/healthcareProfessionalsStore.ts
```diff
@@ -39,6 +39,26 @@ async function queryHealthcareProfessionals() {
     }
 }
 
+export async function getHealthcareProfessionalById(id: string) {
+    try {
+        const healthcareProfessionalId = {
```
ermish, line 44: nit: would rename this to something like data, payload, params, etc.
since other stores are using `data` I'd recommend this:
```suggestion
        const queryData = {
```

ermish (changes requested):
looks really good now!! two small changes and it's done :)

components/ModHealthcareProfessionalSearchbar.vue
Anissa3005: I set the value like this:
```
healthcareProfessionalById.value = await getHealthcareProfessionalById(requestedHealthcareProfessionalId.value)
```

on line 39

ermish: I see it being set, but I don't see it used anywhere after it's set

Anissa3005: It's set already for the PR after this, when we implement the UI, just wanted to make it a bit easier for the person/myself implementing the when working on the UI ticket. 😅

stores/healthcareProfessionalsStore.ts
ermish: Is this one updated? I don't see any changes

components/ModHealthcareProfessionalSearchbar.vue
ermish: Ahhh ok, that's worth putting a comment on in the future since it's unused code. 👍

stores/healthcareProfessionalsStore.ts
Anissa3005: Yes line 44, when you click on the fixup commit you can also see the change

stores/healthcareProfessionalsStore.ts
```diff
@@ -52,6 +52,26 @@ async function queryHealthcareProfessionals() {
     }
 }
 
+export async function getHealthcareProfessionalById(id: string) {
+    try {
+        const queryData = {
```
Anissa3005, line 57: @ermish Here it is, had to rebase so the fixup commit is gone

ermish (approved):
ship it!

PR #612, opened 2024-06-27
feat: show filtered data in ModSubmissionList

### LINKED ISSUE #508: Use local Server data to show correct filtered data to `ModSubmissionList` components
# Problem
Right now we are using hardcoded data to test the data. We need to fetch the local DB data using the ReadMe from the [server](https://github.com/ourjapanlife/findadoc-server ) to set it up. From there we want to have it default to showing Facilities and for Review

# Success
Shows the correct data based on the local state of the `ModDashboardLeftNavbar` component. They are all in the same Container so they should be able to communicate.

### Requirements

- [ ] Successfully display the data based on the state with successful filtering

# Implementation notes
Decide where to filter data. Might be a good idea to do it in the store.

NabbeunNabi:
Resolves #508 

## 🔧 What changed
Before we would show all the different types of submissions in the `SubmissionListComponent`. Now it is filtered based on which tab is selected.

## 🧪 Testing instructions
There are tests added so you can go to the branch and run
```
yarn dev
yarn cypress open
```
and then run the `moderationDashboard.cy.ts` spec

## 📸 Screenshots

-   ### Before
https://www.loom.com/share/cec042893b6c476881e3b398f76f6fc4
-   ### After
https://www.loom.com/share/6f2c0e990acc403fae1203b38e061e6e

### DIFF
--- a/components/ModDashboardLeftNavbar.vue
+++ b/components/ModDashboardLeftNavbar.vue
@@ -17,7 +17,7 @@
                 <button
                     data-testid="mod-dashboard-leftnav-for-review"
                     class="flex flex-row items-center text-start p-1"
-                    @click="updateSubmissionListViewState(SelectedSubmissionListViewTab.forReview)"
+                    @click="updateSubmissionListViewState(SelectedSubmissionListViewTab.ForReview)"
                 >
                     {{ $t("modDashboardLeftNav.forReview") }} ({{ autofillStatusCount.forReviewCount }})
                 </button>
@@ -27,7 +27,7 @@
                 <button
                     data-testid="mod-dashboard-leftnav-approved"
                     class="flex flex-row items-center text-start p-1 "
-                    @click=" updateSubmissionListViewState(SelectedSubmissionListViewTab.approved)"
+                    @click=" updateSubmissionListViewState(SelectedSubmissionListViewTab.Approved)"
                 >
                     {{ $t("modDashboardLeftNav.approved") }} ({{ autofillStatusCount.approvedCount }})
                 </button>
@@ -37,7 +37,7 @@
                 <button
                     data-testid="mod-dashboard-leftnav-rejected"
                     class="flex flex-row items-center text-start p-1"
-                    @click="updateSubmissionListViewState(SelectedSubmissionListViewTab.rejected)"
+                    @click="updateSubmissionListViewState(SelectedSubmissionListViewTab.Rejected)"
                 >
                     {{ $t("modDashboardLeftNav.rejected") }} ({{ autofillStatusCount.rejectedCount }})
                 </button>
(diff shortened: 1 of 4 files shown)


stores/moderationSubmissionsStore.ts
```diff
@@ -4,17 +4,41 @@ import { ref, type Ref } from 'vue'
...
 
         async function getSubmissions() {
             const submissionsSearchResults = await querySubmissions()
             submissionsData.value = submissionsSearchResults
         }
 
-        return { getSubmissions, submissionsData }
+        function filterSubmissionsBySelectedTab(tabValue: SelectedSubmissionListViewTab) {
+            switch (tabValue) {
+                case 'forReview':
```
theyokohamalife, line 26: I think you can use `SelectedSubmissionListViewTab.forReview`, etc instead of the strings here.

theyokohamalife:
Nit: I'd suggest renaming your PR title to something like "feat: show filtered data in ModSubmissionList"

components/ModDashboardLeftNavbar.vue
```diff
@@ -79,6 +78,9 @@ function createCountsForSubmissionListView(submissions: Submission[]) {
         if (submission.isRejected) {
             statusCountObject.rejectedCount += 1
         }
+        if (!submission.isRejected && !submission.isApproved && !submission.isUnderReview) {
```
theyokohamalife, line 81: What do you think about saving these to a variable and just matching the condition on the variable here?

ermish: Nice call @theyokohamalife i see you picking up good habits 🙌🙌

components/ModDashboardLeftNavbar.vue
```diff
@@ -88,4 +90,10 @@ function createCountsForSubmissionListView(submissions: Submission[]) {
 }
 
 const autofillStatusCount = computed(() => createCountsForSubmissionListView(moderationSubmissionsStore.submissionsData))
+
+watchEffect(() => {
+    if (moderationSubmissionsStore.submissionsData.length > 0) {
```
ermish, line 95: Style option: you can remove the >0 part and most people can read this fine

```suggestion
    if (moderationSubmissionsStore.submissionsData.length) {
```

stores/moderationSubmissionsStore.ts
ermish: +1 let's avoid hardcoded strings and use Enums (looks like you already have one)

ermish (changes requested):
Nice work team!

stores/moderationSubmissionsStore.ts
```diff
@@ -4,17 +4,43 @@ import { ref, type Ref } from 'vue'
 import { gqlClient } from '../utils/graphql.js'
 import { type Submission } from '~/typedefs/gqlTypes.js'
 
+export enum SelectedSubmissionListViewTab {
+    ForReview = 'ForReview',
```
ermish, line 8: nit: the value doesn't match the casing as the others

ermish (approved):
nice work!

PR #613, opened 2024-06-29
Chore: upgraded packages

ermish:
Resolves #607 
Resolves #608 

This is a maintenance PR to upgrade our packages regularly

## 🔧 What changed
- As part of eslint 9.6.0 upgrade, the linter fails. I found the bug was how we were importing vue recommended eslint rules and fixed it 🐛 🔨  (you can see #607 is failing because of it)
- upgraded most libraries 
- tested to make sure the app still works


components/ModSubmissionsEditFormFacility.vue
```diff
@@ -286,20 +286,20 @@ const validateFields = () => {
     const isLongitudeValid: boolean = validateFloat(mapLongitude.value)
 
     return (
-        !isNameEnValid ||
```
ermish, line 289: the non-config file changes are just from running `yarn lint:fix` which changed as part of the package upgrades and a fix I had to do to eslint config

theyokohamalife (approved):

PR #614, opened 2024-06-30
feat: implement transition for fade-in images

NabbeunNabi:
## 🔧 What changed
This component can be used anywhere in the app. It allows for a smoother load of images no matter the size. Even after compressing the about us files they would sometimes load in a manner that was less than ideal. Now it should load with a fade in effect if there is a delay or load the whole image at once.

## 🧪 Testing instructions
Go to the branch and run:

```
yarn dev
yarn cypress open
```

## 📸 Screenshots

-   ### Before
https://www.loom.com/share/28505cfdc1d44891993686e793a64c29
-   ### After
https://www.loom.com/share/9fc53ffec9aa4122a7bf8f933379f096

### DIFF
--- a/pages/about.vue
+++ b/pages/about.vue
@@ -61,16 +61,17 @@
             >
                 <div
                     v-for="(member, index) in data.members"
-                    :key="index"
+                    :key="member.avatarImg"
                     data-testid="member"
                     class="members-list grid"
                 >
                     <MemberComponent
                         :avatar-img="member.avatarImg"
-                        :name="member.name"
-                        :title="member.title"
-                        :linked-in-url="member.linkedInUrl"
                         :github-url="member.githubUrl"
+                        :linked-in-url="member.linkedInUrl"
+                        :name="member.name"
+                        :data-test-id="index"
+                        :title="member.title"
                     />
                 </div>
             </div>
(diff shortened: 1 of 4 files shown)


pages/about.vue
```diff
@@ -61,16 +61,17 @@
...
+                    :key="member.avatarImg"
                     data-testid="member"
                     class="members-list grid"
                 >
                     <MemberComponent
                         :avatar-img="member.avatarImg"
+                        :github-url="member.githubUrl"
+                        :linked-in-url="member.linkedInUrl"
                         :name="member.name"
+                        :test-number="index"
```
ermish, line 73: I wouldn't know from reading this property what it does. maybe `data-testid` would be better

ermish (changes requested):
makes it look much better!

pages/about.vue
NabbeunNabi: I have a question though. Would that not be confusing since `data-testid` is a cypress thing. And this is just to pass the index down to be used later in the data-testid? It by itself is not the test Id but being used as an identifier in the test id. Would having an element attribute and a prop named the same thing not be more confusing? Is there another name you would suggest @ermish ?

ermish: `data-testid` is just the name I use for tests since it's generic and can be used by any testing library. 
[Cypress](https://docs.cypress.io/guides/references/best-practices#Selecting-Elements) usually prefers `data-cy`. It's fine to have a prop named that makes it clear what it's being used for. I would just call it `dataTestId` which would be easy to understand but is technically different.

ermish (approved):
Ship it!

PR #615, opened 2024-06-30
chore: remove stylelint

ermish:
## 🔧 What changed
Some cleanup work.
Since we are using tailwindcss, we don't really need stylelint as we don't have css to lint. 

## 🧪 Testing instructions
- CI, `yarn lint`, and `yarn dev` should all work as normal


package.json
```diff
@@ -7,10 +7,9 @@
         "dev": "yarn && nuxi dev",
         "dev:localserver": "NUXT_USE_LOCAL_API=true; yarn && nuxi dev",
         "generate": "graphql-codegen --config ./typesgeneratorconfig.ts",
-        "lint": "ESLINT_USE_FLAT_CONFIG=true eslint . -c eslint.config.js && stylelint **/*.{vue,css,scss} --ignore-path .gitignore",
+        "lint": "ESLINT_USE_FLAT_CONFIG=true eslint . -c eslint.config.js",
```
theyokohamalife, line 10: It looks like this was running on `*.vue` files as well. I don't know much about stylelint, so I'll leave this up to your judgment.

theyokohamalife (approved):

package.json
ermish: yup! it was running against "all our css" in our vue files which is 0 because of tailwind 😆

PR #625, opened 2024-07-02
refactor: implement new error logic in submission form

### LINKED ISSUE #623: BUG: Error messages displayed before filling in the submission
# Problem

In production when you got to `add a doctor`  the error messages are already displayed, this logic should happen after you entered wrong value or no value when filing in the form. 

Image

# Success
Error message only displayed when starting filling in the form and a value is missing or incorrect.
### Requirements

- [ ] Change the logic when error message is displayed

NabbeunNabi:
Resolves #623 

## 🔧 What changed
Before error messages were displayed before the user left the field and only showed when correct. The logic is now changed to reflect the same way we used in the Moderation Panel Form. We validate on blur for the first entry. Then we validate on change until it is correct. We also have an extra check on submit to validate.

## 🧪 Testing instructions
The tests should not have changed. They are the same as before with the submissions. Only error showing logic has changed.

## 📸 Screenshots

-   ### Before
image

-   ### After
https://www.loom.com/share/2d1500f2fb9b4f0b801911c389c41a06

### DIFF
--- a/components/SubmissionForm.vue
+++ b/components/SubmissionForm.vue
@@ -16,7 +16,7 @@
             class="mb-2 text-neutral-600 text-sm font-normal font-['Noto Sans JP']"
         >{{ $t('submitPage.googleMaps') }}</span>
         <p
-            v-show="!isValidGoogleMapsUrl"
+            v-show="!isValidInput.googleMapsUrl.value"
             class="text-currentColor text-xs font-['Noto Sans JP']"
         >
             {{ $t('submitPage.googleMapsValidation') }}
@@ -27,17 +27,31 @@
             type="text"
             required
             class="mb-5 px-3 py-3.5 w-[350px] h-[50px] bg-white rounded-lg border border-zinc-400
-          text-neutral-600 text-sm font-normal font-['Noto Sans JP']"
+            text-neutral-600 text-sm font-normal font-['Noto Sans JP']"
             :placeholder="$t('submitPage.location')"
+            @blur="initialValidationCheck(submissionStore.location, 'googleMaps')"
         >
         <span
             class="mb-2 text-neutral-600 text-sm font-normal font-['Noto Sans JP']"
         >{{ $t('submitPage.healthcareProfessionalName') }}</span>
-        <p
-            v-show="!isValidLastName || !isValidFirstName"
-            class="text-currentColor text-xs font-['Noto Sans JP']"
-        >
-            {{ $t('submitPage.lastNameValidation') }}
+        <p class="flex text-currentColor text-xs font-['Noto Sans JP']">
+            <span
+                :class="[
+                    !isValidInput.lastName.value ? 'visible' : 'invisible',
+                    'w-44',
+                    'mr-1',
+                ]"
+            >
+                {{ $t('submitPage.lastNameValidation') }}
+            </span>
+            <span
+                :class="[
+                    !isValidInput.firstName.value ? 'visible' : 'invisible',
+                    'w-44',
+                ]"
+            >
+                {{ $t('submitPage.firstNameValidation') }}
+            </span>
         </p>
         <div>
             <input
@@ -49,6 +63,7 @@
                 required
                 maxlength="30"
                 :placeholder="$t('submitPage.lastName')"
+                @blur="initialValidationCheck(submissionStore.lastName, 'lastName')"
             >
             <input
                 v-model="submissionStore.firstName"
@@ -58,13 +73,14 @@
                 type="text"
                 maxlength="30"
                 :placeholder="$t('submitPage.firstName')"
+                @blur="initialValidationCheck(submissionStore.firstName, 'firstName')"
             >
         </div>
         <span
             class="mb-2 text-neutral-600 text-sm font-normal font-['Noto Sans JP']"
         >{{ $t('submitPage.spokenLanguage1') }}</span>
         <p
-            v-show="!isValidPrimarySpokenLanguage"
+            v-show="!isValidInput.primarySpokenLangauge.value"
             class="text-currentColor text-xs font-['Noto Sans JP']"
         >
             {{ $t('submitPage.spokenLanguageValidation') }}
@@ -74,6 +90,7 @@
             data-testid="submit-select-language1"
             class="mb-5 px-3 py-3.5 w-[350px] h-[50px] bg-white rounded-lg border border-zinc-400 text-primary-text"
             :placeholder="$t('submitPage.selectLanguage1')"
+            @change="initialValidationCheck(submissionStore.selectLanguage1, 'primaryLanguage')"
         >
             <option
                 v-for="(locale, index) in localeStore.localeDisplayOptions"
@@ -88,7 +105,7 @@
             class="mb-2 text-neutral-600 text-sm font-normal font-['Noto Sans JP']"
         >{{ $t('submitPage.spokenLanguage2') }}</span>
         <p
-            v-show="!isValidSecondarySpokenLanguage"
+            v-show="!isValidInput.secondarySpokenLanguage.value"
             class="text-currentColor text-xs font-['Noto Sans JP']"
         >
             {{ $t('submitPage.invalidOption') }}
@@ -98,6 +115,7 @@
             data-testid="submit-select-language2"
             class="mb-5 px-3 py-3.5 w-[350px] h-[50px] bg-white rounded-lg border border-zinc-400 text-primary-text"
             :placeholder="$t('submitPage.selectLanguage2')"
+            @change="initialValidationCheck(submissionStore.selectLanguage1, 'secondaryLanguage')"
         >
             <option
                 v-for="(locale, index) in localeStore.localeDisplayOptions"
@@ -130,7 +148,7 @@
 </template>
 
 <script lang="ts" setup>
-import { computed, type ComputedRef } from 'vue'
+import { ref, watch, nextTick } from 'vue'
 import * as validations from '../utils/formValidations'
 import { useSubmissionStore } from '~/stores/submissionStore'
 import { useLocaleStore } from '~/stores/localeStore'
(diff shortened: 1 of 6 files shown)


components/SubmissionForm.vue
```diff
@@ -168,4 +188,46 @@ const validateFields = (e: Event) => {
...
+        case ValidationReference.FirstName:
+            return isValidFirstName
+        case ValidationReference.PrimarySpokenLanguage:
+            return isValidPrimarySpokenLanguage
+        case ValidationReference.SecondarySpokenLanguage:
+            return isValidSecondarySpokenLanguage
+    }
+}
+
+submissionInputFields.forEach(({ field, validate, validationReference }) => {
```
ermish, line 223:
```suggestion
submissionInputFields.forEach(({ field, validationFunction, validationReference }) => {
```

i18n/locales/en.json
```diff
@@ -59,7 +59,7 @@
         "healthcareProfessionalName": "Healthcare Professional Name",
         "invalidOption": "Invalid option",
         "lastName": "Last Name",
-        "lastNameValidation": "Enter a last name",
+        "nameValidation": "Enter a first and last name",
```
ermish, line 62: if you split the names, I would make a separate string here for `first` and `last`

components/SubmissionForm.vue
```diff
@@ -168,4 +188,46 @@ const validateFields = (e: Event) => {
...
+    { field: 'firstName', validate: validations.validateUserSubmittedFirstName, validationReference: isValidFirstName },
+    { field: 'selectLanguage1', validate: validations.validateFirstSpokenLanguage, validationReference: isValidPrimarySpokenLanguage },
+    { field: 'selectLanguage2', validate: validations.validateSecondSpokenLanguage, validationReference: isValidSecondarySpokenLanguage }
+]
+
+const initialValidationCheck =
+async (inputValue: string, inputValidationCheck: (inputValue: string) => boolean, field: ValidationReference) => {
+    validationCheckedPreviously.value = true
+    await nextTick()
+    const validationReference = useCorrectValidation(field)
```
ermish, line 204: after figuring out the clever goodness you're writing here, I think we can simplify this logic in one of two ways. 

### Approach 1
Why: A bit simpler, but still keeps your genericized approach

How: you could pass in the `isValid__` variable directly, then you don't need the `useCorrectValidation` method. 

### Approach 2
Why: Do we really need the dynamic complexity here? While cool, it's harder to read and just writing out 5 "dumb" validation methods is easier to read and not too tedius. 

How: Just write a separate validation method tied to each field + `isValidx` boolean


### My recommendation
I like approach 2 as it's the most simple and readable approach. But I totally get that approach 1 is 😎

ermish (changes requested):
some cool ideas here!

components/SubmissionForm.vue
```diff
@@ -168,4 +188,46 @@ const validateFields = (e: Event) => {
...
+    { field: 'lastName', validate: validations.validateUserSubmittedLastName, validationReference: isValidLastName },
+    { field: 'firstName', validate: validations.validateUserSubmittedFirstName, validationReference: isValidFirstName },
+    { field: 'selectLanguage1', validate: validations.validateFirstSpokenLanguage, validationReference: isValidPrimarySpokenLanguage },
+    { field: 'selectLanguage2', validate: validations.validateSecondSpokenLanguage, validationReference: isValidSecondarySpokenLanguage }
+]
+
+const initialValidationCheck =
+async (inputValue: string, inputValidationCheck: (inputValue: string) => boolean, field: ValidationReference) => {
+    validationCheckedPreviously.value = true
+    await nextTick()
```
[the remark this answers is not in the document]
NabbeunNabi: added!

components/SubmissionForm.vue
```diff
@@ -37,7 +39,7 @@
             v-show="!isValidLastName || !isValidFirstName"
             class="text-currentColor text-xs font-['Noto Sans JP']"
         >
-            {{ $t('submitPage.lastNameValidation') }}
+            {{ $t('submitPage.nameValidation') }}
```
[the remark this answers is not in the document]
NabbeunNabi: added and matched the styling of the inputs below in order to match the correct input widths. I think Anissa has a ticket to remove pixel widths so hopefully these two together will eventually fade out hard coded pixels

components/SubmissionForm.vue
NabbeunNabi: Makes sense. I refactored a lot to be like solution two. I added a comment to explain why I have two refs for the validation. The first is just to check if it was checked before. And if it has then we can watch it to still provide the interactivity. But the enum case was removed and the extra function. I went through it and with the comments and organization I think it is much more readable with the same interactivity

components/SubmissionForm.vue
NabbeunNabi: Not using this anymore. I have 5 watches since I think it is more readable. It isn't super dry. But like before. These fields are pretty static so the necessity for dynamic fields is maybe an overcomplicated approach.

i18n/locales/en.json
NabbeunNabi: I separated them to be different for each one. And added the values to all the languages `.json` files. The Portuguese and Japanese have the english and will need to be updated later. I think maybe we should have a way to have a bot to create a ticket in the CI/CD that anytime a language file is merged a ticket is made to check to see if the languages files have any untranslated updates. If the person who wrote the ticket translated files cuz they have the fluency we can close the issue. Otherwise it can be medium priority? @Anissa3005 @ermish @theyokohamalife ?

components/SubmissionForm.vue
```diff
@@ -142,37 +148,46 @@
...
 const localeStore = useLocaleStore()
 
-const validationCheckedPreviously: Ref<boolean> = ref(false)
-
+/*
+These values are used in order to make it more readable but also to add
+interactive for the user. Originally the checks are performed after the
+first submission or the first blur event. Then it checks on change so once
+the user enters a valid format the error goes away
+*/
```
theyokohamalife, line 164: What do you think about either rewording or removing this comment?

components/SubmissionForm.vue
```diff
@@ -142,37 +148,46 @@
...
+const lastNameValidationCheckedPreviously: Ref<boolean> = ref(false)
 const isValidLastName: Ref<boolean> = ref(true)
+
+const firstNameValidationCheckedPreviously: Ref<boolean> = ref(false)
 const isValidFirstName: Ref<boolean> = ref(true)
+
+const primarySpokenLanguageValidationCheckedPreviously: Ref<boolean> = ref(false)
 const isValidPrimarySpokenLanguage: Ref <boolean> = ref(true)
+
+const secondaryLanguageValidationCheckedPreviously: Ref<boolean> = ref(false)
```
theyokohamalife, line 177: I wonder if we could simplify this by adding the values to two separate maps: one for `previouslyChecked` and another for `isValid`. @ermish  thoughts?

components/SubmissionForm.vue
```diff
@@ -189,45 +204,64 @@ const validateFields = (e: Event) => {
...
-            return isValidLastName
-        case ValidationReference.FirstName:
-            return isValidFirstName
-        case ValidationReference.PrimarySpokenLanguage:
-            return isValidPrimarySpokenLanguage
-        case ValidationReference.SecondarySpokenLanguage:
-            return isValidSecondarySpokenLanguage
+        case 'googleMaps':
+            googleMapsValidationCheckedPreviously.value = true
+            // nextTick is used to ensure that the model value is updated before doing the validation check
```
theyokohamalife, line 211: What do you think about leaving this comment just once at the top of the switch/case?

components/SubmissionForm.vue
```diff
@@ -189,45 +204,64 @@ const validateFields = (e: Event) => {
...
+            await nextTick()
+            isValidGoogleMapsUrl.value = validations.validateGoogleMapsUrlInput(inputValue)
+            break
+        case 'lastName':
+            lastNameValidationCheckedPreviously.value = true
+            // nextTick is used to ensure that the model value is updated before doing the validation check
```
theyokohamalife, line 217: ♻️

components/SubmissionForm.vue
```diff
@@ -189,45 +204,64 @@ const validateFields = (e: Event) => {
...
+            await nextTick()
+            isValidLastName.value = validations.validateUserSubmittedLastName(inputValue)
+            break
+        case 'firstName':
+            firstNameValidationCheckedPreviously.value = true
+            // nextTick is used to ensure that the model value is updated before doing the validation check
```
theyokohamalife, line 223: ♻️

components/SubmissionForm.vue
```diff
@@ -189,45 +204,64 @@ const validateFields = (e: Event) => {
...
+            await nextTick()
+            isValidFirstName.value = validations.validateUserSubmittedFirstName(inputValue)
+            break
+        case 'primaryLanguage':
+            primarySpokenLanguageValidationCheckedPreviously.value = true
+            // nextTick is used to ensure that the model value is updated before doing the validation check
```
theyokohamalife, line 229: ♻️

components/SubmissionForm.vue
```diff
@@ -189,45 +204,64 @@ const validateFields = (e: Event) => {
...
+            await nextTick()
+            isValidPrimarySpokenLanguage.value = validations.validateFirstSpokenLanguage(inputValue)
+            break
+        case 'secondaryLanguage':
+            secondaryLanguageValidationCheckedPreviously.value = true
+            // nextTick is used to ensure that the model value is updated before doing the validation check
```
theyokohamalife, line 235: ♻️

components/SubmissionForm.vue
NabbeunNabi: Fine with me. I added it because in an earlier resolved conversation @ermish mentioned this

"Philosophy: Comments should explain "why" and not "what"

let's add a comment here to help the next reader understand why there's a next tick here"

So I thought just so others would understand what is happening they could see in the comment. I can just remove it and add it as a comment to this part of the PR

components/SubmissionForm.vue
NabbeunNabi: That sounds good

components/SubmissionForm.vue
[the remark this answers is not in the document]
NabbeunNabi: Not using this anymore due to refactoring

components/SubmissionForm.vue
NabbeunNabi: > after figuring out the clever goodness you're writing here, I think we can simplify this logic in one of two ways.
> 
> ### Approach 1
> Why: A bit simpler, but still keeps your genericized approach
> 
> How: you could pass in the `isValid__` variable directly, then you don't need the `useCorrectValidation` method.
> 
> ### Approach 2
> Why: Do we really need the dynamic complexity here? While cool, it's harder to read and just writing out 5 "dumb" validation methods is easier to read and not too tedius.
> 
> How: Just write a separate validation method tied to each field + `isValidx` boolean
> 
> ### My recommendation
> I like approach 2 as it's the most simple and readable approach. But I totally get that approach 1 is 😎

@theyokohamalife. This is what Phil mentioned when I was using dynamic complexity

components/SubmissionForm.vue
NabbeunNabi: Originally I had an array of objects. If I understood correctly @ermish commented earlier that by doing certain things that way was maybe overcomplicating it. I think based on the same philosophy of hardcoding the html if it doesn't need to be dynamic for the readability. But let me know if I misunderstood. I quoted his comment from before.

components/SubmissionForm.vue
```diff
@@ -27,17 +27,31 @@
...
-        <p
-            v-show="!isValidLastName || !isValidFirstName"
-            class="text-currentColor text-xs font-['Noto Sans JP']"
-        >
-            {{ $t('submitPage.lastNameValidation') }}
+        <p class="flex text-currentColor text-xs font-['Noto Sans JP']">
+            <span
+                :class="[
+                    !isValidLastName ? 'visible' : 'invisible',
+                    'w-[170px]',
```
ermish, line 41: try to avoid fixed pixel size widths. You can use something like `w-48`

components/SubmissionForm.vue
```diff
@@ -27,17 +27,31 @@
...
         >{{ $t('submitPage.healthcareProfessionalName') }}</span>
-        <p
-            v-show="!isValidLastName || !isValidFirstName"
-            class="text-currentColor text-xs font-['Noto Sans JP']"
-        >
-            {{ $t('submitPage.lastNameValidation') }}
+        <p class="flex text-currentColor text-xs font-['Noto Sans JP']">
+            <span
+                :class="[
+                    !isValidLastName ? 'visible' : 'invisible',
```
ermish, line 40: why use `visible` vs `v-show`?

ermish (approved):
looks good! double check the comments before you merge

components/SubmissionForm.vue
theyokohamalife: @NabbeunNabi I was referring to something different like just a simple map to store the booleans. For example:

```
const checkedPreviously = {
"googleMaps": Ref<boolean> = ref(false), 
"lastName" : Ref<boolean> = ref(false)
}

const isValid = {
"googleMapsUrl": Ref<boolean> = ref(true),
"LastName": Ref<boolean> = ref(true)
}
```

I haven't personally tried using a ref instead of a map so I'm not sure if it would work. It would help to improve the readability though.

cypress/e2e/submit.cy.ts
```diff
@@ -84,7 +84,8 @@ describe('Submit page', () => {
         it('requires a last name of 30 characters or less', () => {
             cy.get('[data-testid="submit-input-lastname"]').type(' ')
             cy.contains(enUS.submitPage.lastNameValidation).should('be.visible')
-            cy.get('[data-testid="submit-input-lastname"]').type('some last name under 30 characters')
+            cy.get('[data-testid="submit-input-lastname"]').type('The Frog')
+            cy.get('[data-testid="submit-input-firstname"]').type('Kermy')
```
theyokohamalife, line 88: 🐸

theyokohamalife:
I could be wrong, but I think that the first name isn't required for a user submission, and that's why it was initially left out of the validation. I would check the backend to confirm.

NabbeunNabi:
> I could be wrong, but I think that the first name isn't required for a user submission, and that's why it was initially left out of the validation. I would check the backend to confirm.

Ahh ok. I'll give it a look. I didn't add any verification methods to the file so I assumed cuz I was just looking at what was already provided

NabbeunNabi:
> I could be wrong, but I think that the first name isn't required for a user submission, and that's why it was initially left out of the validation. I would check the backend to confirm.

```
type LocalizedName {
  firstName: String!
  middleName: String
  lastName: String!
  locale: Locale!
}
```

This from the backend so it seems required. It was being checked originally just did not have a name being shown

components/SubmissionForm.vue
NabbeunNabi: Great Idea. It should work!

components/SubmissionForm.vue
NabbeunNabi: I use this because v-show uses hidden which does not keep the spacing. I wanted to keep the spacing so the span for the first name would appear above the First name input.

If I use v-show and the last name is valid but the first name isn't the error will appear above the last name.

PR #630, opened 2024-07-03
FIX: Remove the extra scrollbar on the left navbar

### LINKED ISSUE #624: BUG: Double scrollbars in the search results
# Problem
In the search results there are double scrollbars. It seems a better idea to have one scroll bar for the search results. I think it's better to have only a scrollbar to scroll over the doctors, so the filters tab keeps visible.  

Image

# Success
Having one scrollbar
### Requirements

- [ ] Remove one of the scrollbar

Anissa3005:
Resolves #624 

## 🔧 What changed

The left navbar had a double scrollbars. I have removed the outer scrollbar, and kept the inner scrollbar so the Filters are always visible while scrolling through the doctors.

1. Removed unnecessary overflows
2. Removed unnecessary hover effect

## 📸 Screenshots

-   ### Before
image

-   ### After
image


components/LeftNavbar.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-    <div class="border-r border-secondary-bg/40 hover:shadow-inner hover:shadow-secondary-bg/50 transition-shadow">
+    <div>
```
Anissa3005, line 2: I removed this class since it seemed unnecessary? 
We have all the hover effects applied on the individual doctors items.
After I removed it I tested if it made visual a difference on different sizes and I could not see any difference with and without the class. 
Please let me know if there was actually a need for this class.

ermish: It looks very different on my screen now as the sidebar is bleeding into the tapnav. You'll notice it even more in a dark theme too. 
You could make the shadows stick instead of transitioning if you want.  

I would keep the border alone shadows as they help distinguish the layers between component me which are inspired after googles material design standards for layering.

ermish (changes requested):
Let's put back the styling for now and this is good to go! 
Really happy to see the scrollbar fixed!!

theyokohamalife:
@Anissa3005 I don't see the double scrollbars that you're showing in your screenshot. 🤔  When I visit the site in production, this is what I see: 

Screenshot 2024-07-03 at 11 56 03 PM

Anissa3005:
@theyokohamalife I use chrome, on my laptop screen. In production there are two scroll bars. There were three overflows applied so I think that was anyway unnecessary.

ermish:
> I'd like to confirm if other people are experiencing this issue in production or not before making these changes.

I've also seen in prod

NabbeunNabi:
> I'd like to confirm if other people are experiencing this issue in production or not before making these changes.
I am experiencing these issues. It could be an OS thing with how scrollbars are applied. Like the discussion I had with Natalja one time about how windows and macs apply scrollbars different

image

ermish (approved):
ship it! :shipit:

ermish:
@theyokohamalife  can you re-review?

PR #631, opened 2024-07-03
feat: implement pagination for moderation

### LINKED ISSUE #629: Implement pagination for submission ID
# Problem
Currently we are conditionally rendering the submission form. This will update the logic to use pagination so we can use the browsers native back and forward navigation

# Success
Implements the id of a submission in the route path
### Requirements

- [ ] Implement pagination
- [ ] Can use the browser's native navigation for forward and back 

# Implementation notes

NabbeunNabi:
Resolves #629 

## 🔧 What changed
Before we were using a moderation store to conditionally render certain commits. Now we use the params and paginations in order to use the native browser buttons.

## 🧪 Testing instructions
No tests were added. Since it was refactoring the other tests were still needed to pass.

## 📸 Screenshots

-   ### After
https://www.loom.com/share/5fd7daff22d8412da736d865cbb0ae0a

### DIFF
--- a/components/ModEditSubmissionTopbar.vue
+++ b/components/ModEditSubmissionTopbar.vue
@@ -47,7 +47,7 @@
 </template>
 
 <script setup lang="ts">
-import { ref, watch, type Ref } from 'vue'
+import { ref, type Ref } from 'vue'
 import SVGCopyContent from '~/assets/icons/content-copy.svg'
 import SVGSuccessCheckMark from '~/assets/icons/checkmark-square.svg'
 import { useModerationSubmissionsStore } from '~/stores/moderationSubmissionsStore'
@@ -76,8 +76,4 @@ const acceptSubmission = () => {
 const rejectSubmission = () => {
     console.log('You are rejected')
 }
-
-watch(selectedSubmissionId, newSelectedId => {
-    selectedSubmissionId.value = newSelectedId
-})
 </script>
--- a/components/ModEditSubmissionTopbar.vue
+++ b/components/ModEditSubmissionTopbar.vue
@@ -53,7 +53,6 @@ import SVGSuccessCheckMark from '~/assets/icons/checkmark-square.svg'
 import { useModerationSubmissionsStore } from '~/stores/moderationSubmissionsStore'
 
 const modScreenStore = useModerationSubmissionsStore()
-
 const selectedSubmissionId = modScreenStore.selectedSubmissionId
 
 const showCopySuccessIcon: Ref<boolean> = ref(false)
--- a/components/ModEditSubmissionTopbar.vue
+++ b/components/ModEditSubmissionTopbar.vue
@@ -47,25 +47,25 @@
 </template>
 
 <script setup lang="ts">
-import { ref, type Ref } from 'vue'
+import { ref, watch, type Ref } from 'vue'
 import SVGCopyContent from '~/assets/icons/content-copy.svg'
 import SVGSuccessCheckMark from '~/assets/icons/checkmark-square.svg'
 import { useModerationSubmissionsStore } from '~/stores/moderationSubmissionsStore'
 
-const modScreenStore = useModerationSubmissionsStore()
-const selectedSubmissionId = modScreenStore.selectedSubmissionId
+const modSubmissionStore = useModerationSubmissionsStore()
+const selectedSubmissionId: Ref<string> = ref(modSubmissionStore.selectedSubmissionId)
 
 const showCopySuccessIcon: Ref<boolean> = ref(false)
 
 const copySubmissionId = async () => {
     try {
-        await navigator.clipboard.writeText(selectedSubmissionId)
+        await navigator.clipboard.writeText(selectedSubmissionId.value)
         showCopySuccessIcon.value = true
         setTimeout(() => {
             showCopySuccessIcon.value = false
         }, 2000)
     } catch (err: unknown) {
-        console.error(`Failed to copy submission ID ${selectedSubmissionId}: ${err}`)
+        console.error(`Failed to copy submission ID ${selectedSubmissionId.value}: ${err}`)
     }
 }
 
@@ -76,4 +76,8 @@ const acceptSubmission = () => {
 const rejectSubmission = () => {
     console.log('You are rejected')
 }
+
+watch(selectedSubmissionId, newSelectedId => {
+    selectedSubmissionId.value = newSelectedId
+})
 </script>
--- a/components/ModMainContent.vue
+++ b/components/ModMainContent.vue
@@ -1,10 +1,10 @@
 <template>
     <div class="h-full">
-        <div v-if="store.activeScreen === ModerationScreen.dashboard">
+        <div v-if="screenStore.activeScreen === ModerationScreen.dashboard">
             <ModSubmissionListContainer />
         </div>
         <div
-            v-else-if="store.activeScreen === ModerationScreen.editSubmission"
+            v-else-if="screenStore.activeScreen === ModerationScreen.editSubmission"
             class="h-full"
         >
             <ModSubmissionsEditFormFacility />
@@ -13,7 +13,32 @@
 </template>
 
 <script setup lang="ts">
+import { useRoute } from 'vue-router'
+import { computed, onMounted, watch } from 'vue'
 import { useModerationScreenStore, ModerationScreen } from '~/stores/moderationScreenStore'
+import { useModerationSubmissionsStore } from '~/stores/moderationSubmissionsStore'
 
-const store = useModerationScreenStore()
+const route = useRoute()
+
+const screenStore = useModerationScreenStore()
+const moderationSubmissionsStore = useModerationSubmissionsStore()
+
+const submissionIdInParams = computed(() => route.params.id as string)
+
+watch(submissionIdInParams, newSelectedSubmissionId => {
+    if (newSelectedSubmissionId) {
+        screenStore.setActiveScreen(ModerationScreen.editSubmission)
+        moderationSubmissionsStore.selectedSubmissionId = newSelectedSubmissionId
+    } else {
+        screenStore.setActiveScreen(ModerationScreen.dashboard)
+        moderationSubmissionsStore.selectedSubmissionId = ''
+    }
+})
+
+onMounted(() => {
+    if (submissionIdInParams.value) {
+        screenStore.setActiveScreen(ModerationScreen.editSubmission)
+        moderationSubmissionsStore.selectedSubmissionId = submissionIdInParams.value
+    }
+})
 </script>
--- a/components/ModSubmissionListContainer.vue
+++ b/components/ModSubmissionListContainer.vue
@@ -32,11 +32,16 @@
                     class="grid grid-cols-subgrid col-span-5 bg-tertiary-bg cursor-pointer hover:bg-primary"
                     @click="handleClickToSubmissionForm(submission.id)"
                 >
-                    <span class="text-start">{{ index + 1 }}</span>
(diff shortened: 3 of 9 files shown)


components/ModEditSubmissionTopbar.vue
```diff
@@ -76,4 +74,6 @@ const acceptSubmission = () => {
 const rejectSubmission = () => {
     console.log('You are rejected')
 }
+
+const submissionId = computed(() => route.params.id as string)
```
ermish, line 78: This is looking good and happy to see some vue router action! I think there's a few things that we can adjust here so I'll just put it in one comment

## feature: adding support for refreshing page or coming from a shared url. 

You could do this is the PR or separate. I'd say add it here as it will require a few changes to the code to make it work. 

#### what needs to change 
- since we could view or edit submissions, you should specify that this is `/editsubmission/id` 
  - this will also help when you refresh to know which page you're going to 
- You need to make the pages based on the screen again as the submission id could be for more than one screen
- I would recommend setting back to `selectedsubmissionId` again as it's a little more clear which submission id you're referring to. (For example as opposed to `submissionIdToDelete`)
- I would probably set the selectedsubmissionId in the store as it's shared and part of the route
- you can set the screen based on the route now if you have the `editSubmission/` route. 
- You need to add in logic for when the mod dashboard is first loaded, it should check the route and go to the edit page immediately. This could be tricky to not have the page flash quickly. Let me know if you want to know how or you want to try and figure it out 😎

ermish (changes requested):
Check out the design comment

components/ModEditSubmissionTopbar.vue
NabbeunNabi: "You need to make the pages based on the screen again as the submission id could be for more than one screen"


I do not understand this. When would the id be used for a different screen. It was conditional based off the selected submissionId before. We are going to add more screens? Wouldn't we use the pagination to continue to conditionally render everywhere? Is it not adding complexity to use router parameters and a store where they can do the same thing? Or does it just give way more control?

Since you suggested `editSubmission/submissionId` couldn't we use the router in place of the store? Is this bad practice? Less reliable?

ermish: Let's huddle tomorrow  and we can talk through it 👌

NabbeunNabi: I think I understand now. If we have another variable we set in the store like `isSubmissionToDelete` we could show a different screen or a modal or something based on the screen. E.G. `ModerationScreen.ConfrimDelete`.

NabbeunNabi: > This is looking good and happy to see some vue router action! I think there's a few things that we can adjust here so I'll just put it in one comment
> 
> ## feature: adding support for refreshing page or coming from a shared url.
> You could do this is the PR or separate. I'd say add it here as it will require a few changes to the code to make it work.
> 
> #### what needs to change
> * since we could view or edit submissions, you should specify that this is `/editsubmission/id`
>   
>   * this will also help when you refresh to know which page you're going to
> * You need to make the pages based on the screen again as the submission id could be for more than one screen
> * I would recommend setting back to `selectedsubmissionId` again as it's a little more clear which submission id you're referring to. (For example as opposed to `submissionIdToDelete`)
> * I would probably set the selectedsubmissionId in the store as it's shared and part of the route
> * you can set the screen based on the route now if you have the `editSubmission/` route.
> * You need to add in logic for when the mod dashboard is first loaded, it should check the route and go to the edit page immediately. This could be tricky to not have the page flash quickly. Let me know if you want to know how or you want to try and figure it out 😎

Updated based on feedback plus reasoning.
-The path now is `/editsubmission/id` for the form.
- The pages are reverted to the screen again and my understanding is show as above.
- The selectedSubmissionId is set in the store if we need to use it other places and watched in order to update everywhere.
-I added a ref that is a boolean to check the path. and added different conditions to show the different page so the one that shows the link to the form isn't shown unless auth fails. Otherwise the user is shown 'Checking Authentication'. I left that in for all the paths cuz it is happening and I want the user to see that it is still a safe and protected path.

components/ModEditSubmissionTopbar.vue
```diff
@@ -76,4 +75,8 @@ const acceptSubmission = () => {
 const rejectSubmission = () => {
     console.log('You are rejected')
 }
+
+watch(selectedSubmissionId, newSelectedId => {
```
ermish, line 79: uhhh....it's watching and setting itself 😆

components/ModMainContent.vue
```diff
@@ -13,7 +13,32 @@
...
 
-const store = useModerationScreenStore()
+const route = useRoute()
+
+const screenStore = useModerationScreenStore()
+const moderationSubmissionsStore = useModerationSubmissionsStore()
+
+const submissionIdInParams = computed(() => route.params.id as string)
+
+watch(submissionIdInParams, newSelectedSubmissionId => {
```
ermish, line 28: nice!

components/ModSubmissionListContainer.vue
```diff
@@ -32,11 +32,16 @@
...
                     @click="handleClickToSubmissionForm(submission.id)"
                 >
-                    <span class="text-start">{{ index + 1 }}</span>
-                    <span class="text-start">{{ submission.healthcareProfessionalName }}</span>
-                    <span class="text-start">{{ getSubmissionStatus(submission) }}</span>
-                    <span class="text-start">{{ convertDateToLocalTime(submission.updatedDate) }}</span>
-                    <span class="text-start">{{ convertDateToLocalTime(submission.createdDate) }}</span>
+                    <nuxt-link
+                        :to="`/moderation/editsubmission/${submission.id}`"
+                        class="grid grid-cols-subgrid col-span-5 bg-gray-200 p-1 hover:bg-primary"
```
ermish, line 37: let's use our theme colors

```suggestion
                        class="grid grid-cols-subgrid col-span-5 bg-primary-text-muted p-1 hover:bg-primary"
```

cypress.config.ts
```diff
@@ -5,6 +5,7 @@ export default defineConfig({
     e2e: {
         setupNodeEvents() { },
         baseUrl: 'http://localhost:3000',
-        testIsolation: false
+        testIsolation: false,
+        experimentalRunAllSpecs: true
```
ermish, line 9: 💯

pages/moderation.vue
```diff
@@ -71,13 +71,16 @@ const authStore = useAuthStore()
 const route = useRoute()
 const router = useRouter()
 
+const doesTheUserHaveAccess: Ref<boolean> = ref(true)
```
ermish, line 74: nice variable name

pages/moderation.vue
```diff
@@ -71,13 +71,16 @@ const authStore = useAuthStore()
...
+
 const checkIfUserIsLoggedIn = async () => {
     // This promise is here to make the Suspense component work.
     // It doesn't do anything, but <Suspense> requires an awaited setup method
     await new Promise(resolve => {
         //ignore if route change is unrelated to moderation
-        if (route.path === '/moderation' && !authStore.isLoggedIn && !authStore.isLoadingAuth) {
+        if (route.path.startsWith('/moderation') && !authStore.isLoggedIn && !authStore.isLoadingAuth) {
             // give the user a bit of time to read the message before redirecting
+            doesTheUserHaveAccess.value = false
```
ermish, line 83: you can make this a bit clearer like this:
```
doesTheUserHaveAccess.value = route.path.startsWith('/moderation') && !authStore.isLoggedIn && !authStore.isLoadingAuth
if(doesTheUserHaveAccess.value) 
....
```

pages/moderation.vue
```diff
@@ -90,7 +93,12 @@ const checkIfUserIsLoggedIn = async () => {
 
 watch(route, async () => {
     await checkIfUserIsLoggedIn()
+    console.log('2', doesTheUserHaveAccess.value)
```
ermish, line 96: 💣

pages/moderation.vue
```diff
@@ -90,7 +93,12 @@ const checkIfUserIsLoggedIn = async () => {
...
 })
 
-await checkIfUserIsLoggedIn()
+onMounted(async () => {
+    console.log('1', doesTheUserHaveAccess.value)
```
ermish, line 100: 💣

pages/moderation.vue
```diff
@@ -90,7 +93,12 @@ const checkIfUserIsLoggedIn = async () => {
...
+    await checkIfUserIsLoggedIn()
+    console.log('2', doesTheUserHaveAccess.value)
```
ermish, line 102: 💣

pages/moderation/editsubmission/index.vue
```diff
@@ -0,0 +1,3 @@
+<template>
```
ermish, line 1: do you still need these 3 pages?

ermish (changes requested):
great job re-designing this!!

components/ModEditSubmissionTopbar.vue
NabbeunNabi: Whoops

components/ModSubmissionListContainer.vue
NabbeunNabi: Will do! Sometimes I don't notice the tailwind when just changing the functionality. I'll try to look more at the files I am working on in order to get rid of the old styling

pages/moderation.vue
NabbeunNabi: Whoops. Forgot to remove these

pages/moderation.vue
NabbeunNabi: ♻️

pages/moderation.vue
NabbeunNabi: ♻️

pages/moderation/editsubmission/index.vue
NabbeunNabi: I believe so. I tried removing different ones. Seems seems like every folder that is a page needs its on index to create the `/param` and if I want to add an id as the param it needs the file with the name of the `[param]`

ermish: hmmmmmmmmmmmmmmmmmm.,..... I'm wondering if we actually need it. I'm guessing you normally would for building out static pages through nuxt+vue router. However, since these pages aren't indexed or publicly visible, I'm wondering if there's a programmatic way to a subscribe them in vue router without requiring pages.

NabbeunNabi: https://nuxt.com/docs/getting-started/routing

Seems like you are right and I need to refactor and see how to get it working 🙇

NabbeunNabi: Done!

ermish (approved):
looks great! nice rework of this 👍 
Do we still need the `/[id].vue` page? or is there a vue router programmatic way to add/register this route instead of having a page? That would be better since 1) this page isn't publicly exposed 2) I don't think it needs to be a full page with a page refresh. The edit page is more of a component than a full page to me.
I'll approve and merge, but please investigate if it's possible. We can do in a separate PR/issue

NabbeunNabi:
> looks great! nice rework of this 👍 Do we still need the `/[id].vue` page? or is there a vue router programmatic way to add/register this route instead of having a page? That would be better since 1) this page isn't publicly exposed 2) I don't think it needs to be a full page with a page refresh. The edit page is more of a component than a full page to me. I'll approve and merge, but please investigate if it's possible. We can do in a separate PR/issue

This is the way that Nuxt and vue use routing as seen in the documentation as a NuxtPage in this documentation here:
https://nuxt.com/docs/getting-started/routing

I cannot seem to see if there is another way. I looked before implementing this. Should I see if there is a different workaround to how Nuxt Routes?

PR #632, opened 2024-07-03
Toggle location submissions and healthcareProfessionals submissions

### LINKED ISSUE #601: FEAT: Toggle between facility submissions and healthcare professional submissions
# Problem
At the moment there is no functionality switching between the facility and healthcare professional submissions list. In this issue we want to add the functionality if healthcare professional is selected we need to display the healthcare professional submissions.

# Success
Going from this:
Image

To thiss:
Image

### Requirements

- [x] Add logic when healthcare professionals is selected display the healthcare professional submissions

Diego-Bittencourt:
Resolves #601

## 🔧 What changed
This PR contains a big rework of logic based off of @Anissa3005 and @ermish's discussion of the flow with the backend. It seems like a lot of files changes. But a lot of it is just variable name changes that were affected in other files due to the logic changing and me finding it better to have variable names and a store name more closely to what it was doing. 

We now have three possible screens set in the `ModSubmissionListComponent` The name didn't change. But it displays either submissions, facilities, or healtchare professionals. This allows us to edit all of the database after the fact. Whether accepted or not. It defaults to submissions with a sidebar as previously shown in the Figma. The other two just show the select based on what screen and selection there is. We can confirm later what we would like the `LeftNav` to have if any.

There are also tests to show that the correct lists are showing. And paths for each form that will be necessary in the future with a set ID for each one.

## 🧪 Testing instructions
You can run the tests locally with

```
yarn dev
```
```
yarn cypress open
```

### Before
image

### After
image
image
image

This is the path that it will go to for others. The logic and Figma might still need to be implemented. The TopBar probably could be the same. But since they already exist it might be different?

image


components/ModSubmissionListContainer.vue
```diff
@@ -33,7 +33,10 @@
                     @click="handleClickToSubmissionForm(submission.id)"
                 >
                     <span class="text-start">{{ index + 1 }}</span>
-                    <span class="text-start">{{ submission.healthcareProfessionalName }}</span>
+                    <span
+                        :data-testid="`mod-submission-name-${submission.healthcareProfessionalName}`"
```
theyokohamalife, line 37: What do you think about using either the index or an ID here instead of the `heathcareProfessionalName`? An `id` would be more uniquely identifiable and probably easier to test for. It's possible for multiple healthcare professionals to have the same name.

Perhaps we could do something like: 
```
:data-testid="`mod-submission-healthcareprofessional-${submission.id}`"
```

theyokohamalife:
You're off to a really good start! ✨  Left one small suggestion.

components/ModDashboardLeftNavbar.vue
```diff
@@ -65,6 +66,11 @@ const updateSubmissionListViewState = (submissionListViewOptionValue: SelectedSu
 const updateSelectedDashboardView = (event: Event) => {
     const selectedValue = (event.target as HTMLSelectElement).value
     selectedDashboardView.value = selectedValue
+    if (selectedDashboardView.value === 'facilities') {
+        moderationSubmissionsStore.getSubmissions()
+    } else if (selectedDashboardView.value === 'healthcareProfessionals') {
+        moderationSubmissionsStore.getProfessionals()
+    }
```
Anissa3005, line 73: This part would be perfect if we need to fetch different data. But in this case the healthcare professionals submissions are existing in the `getSubmissions()` so we just need to display different data from that submission query. 

If you go to the `modSubmissionsStore` in the `getSubmissionsGqlQuery` you will see the `healthcareProfessionals` field this is what we want to display if we toggle Healthcare Professionals. At the moment that is still empty so I would use fake data for now as you did.

stores/moderationSubmissionsStore.ts
```diff
@@ -23,6 +23,11 @@ export const useModerationSubmissionsStore = defineStore(
             submissionsData.value = submissionsSearchResults
         }
 
+        async function getProfessionals() {
```
Anissa3005, line 26: Not needed since we can use the `submissionsData` healthcareProfessionsal field

Anissa3005 (commented):
Looks great  ✨
Some logic needs to be changed and than it's good to go

components/ModDashboardLeftNavbar.vue
```diff
@@ -65,6 +66,11 @@ const updateSubmissionListViewState = (submissionListViewOptionValue: SelectedSu
 const updateSelectedDashboardView = (event: Event) => {
     const selectedValue = (event.target as HTMLSelectElement).value
     selectedDashboardView.value = selectedValue
+    if (selectedDashboardView.value === 'facilities') {
+        moderationSubmissionsStore.getSubmissions()
+    } else if (selectedDashboardView.value === 'healthcareProfessionals') {
+        moderationSubmissionsStore.getProfessionals()
```
ermish, line 72: these get calls should be `await`ed

stores/moderationSubmissionsStore.ts
ermish: ^ could we keep the method, but it just pulls out the professionals from the submission objects?

Anissa3005: @ermish In my understanding the submissions get fetched when the moderation panel is accessed. So all the data is there. 
This is the data that is been fetched:
```
type Submission {
  id: ID!
  googleMapsUrl: String!
  healthcareProfessionalName: String!
  spokenLanguages: [Locale!]!
  facility: FacilitySubmission
  healthcareProfessionals: [HealthcareProfessionalSubmission!]
  isUnderReview: Boolean!
  isApproved: Boolean!
  isRejected: Boolean!
  createdDate: String!
  updatedDate: String!
  notes: String
}
```

So all we need to do is display the `facility` and `healthcareProfessionals` fields.

components/ModDashboardLeftNavbar.vue
Anissa3005: @Diego-Bittencourt I have added seeding data for this to make it easier, so you should be able to use that data instead of creating your own fake data! So the facility and healthcareProfessional field in submissions is now seeded and not empty anymore.

stores/moderationSubmissionsStore.ts
NabbeunNabi: I kept it but just made a new object with the filtered healthcareprofessionals if we want to use that is any way. I also made a different view. I don't know what kind of link we might want to add. Right now it links you to the Facility Form. But @Anissa3005 How are you going to display the healthcare professional form? Should it be based on a different page? This might be easiest with `editsubmission/healthcare-professional` or something like that

ermish:
@Anissa3005 I started looking at this, but it might be conflicting with the design you and I talked about. Can you and @Diego-Bittencourt  sync on this?

NabbeunNabi:
> @Anissa3005 I started looking at this, but it might be conflicting with the design you and I talked about. Can you and @Diego-Bittencourt  sync on this?

@ermish I took over this ticket and talked to @Anissa3005 after.

I will update it based on what you both talked about on Monday. And then post it again in the code review channel

stores/moderationSubmissionsStore.ts
```diff
@@ -2,18 +2,25 @@ import { gql } from 'graphql-request'
...
     ForReview = 'FOR_REVIEW',
     Approved = 'APPROVED',
     Rejected = 'REJECTED'
 }
 
+export enum SelectedSubmissionType {
+    Facilities = 'FACILITIES',
+    HealthcareProfessionals = 'HEALTHCARE_PROFESSIONALS'
+}
+
```
Anissa3005, line 17: Not clear by the name of the enum. But we want to toggle between 3 things; healthcare professionals, facilities and submissions.

- Healthcare professionals are the hp's existing in our database (so not the healthcare professionals submissions) 
- Facilities are the existing facilities (so not the facility submissions) 
- Submissions the existing submissions

stores/moderationSubmissionsStore.ts
```diff
@@ -2,18 +2,25 @@ import { gql } from 'graphql-request'
...
 export const useModerationSubmissionsStore = defineStore(
     'submissionsStore',
     () => {
         const submissionsData: Ref<Submission[]> = ref([])
+        const submissionHealtchareProfessionalsData: Ref< (Maybe<HealthcareProfessionalSubmission[]> | undefined)[]>
+        = ref([])
```
Anissa3005, line 23: We want to fetch the healthcare professionals instead of the healthcare professional submissions. Same for the Facilities

stores/moderationSubmissionsListStore.ts
```diff
@@ -0,0 +1,249 @@
+import { gql } from 'graphql-request'
```
ermish, line 1: should rename this to just moderationSubmissionsStore. The list is the UI presentation part which should be abstracted from a store

stores/moderationSubmissionsListStore.ts
```diff
@@ -0,0 +1,249 @@
...
+
+        function setSelectedModerationListViewChosen(selectedOption: SelectedModerationListView) {
+            selectedModerationListViewChosen.value = selectedOption
+        }
+
+        function filterSelectedSubmission(submissionId: string | undefined) {
+            selectedSubmissionData.value = submissionsData.value.find(submission => submission.id === submissionId)
+        }
+
+        function filterSubmissionsBySelectedTab(tabValue: SelectedSubmissionListViewTab) {
```
ermish, line 56: should keep this store data separate from the UI tab implementation. You can change this to filter by status and use the status enum (approved, rejected, inreview)

stores/moderationSubmissionsListStore.ts
```diff
@@ -0,0 +1,249 @@
...
+            submissionsFilters
+        )
+        return result?.submissions ?? []
+    } catch (error) {
+        console.log(`Error querying the submissions: ${JSON.stringify(error)}`)
+        return []
+    }
+}
+
+async function queryFacilities() {
```
ermish, line 112: I think it would be good to move the queryFacilities and queryProfessionals to their own respective state stores and then call them from here.

stores/moderationSubmissionsListStore.ts
```diff
@@ -0,0 +1,249 @@
...
+            selectedSubmissionData.value = submissionsData.value.find(submission => submission.id === submissionId)
+        }
+
+        function filterSubmissionsBySelectedTab(tabValue: SelectedSubmissionListViewTab) {
+            switch (tabValue) {
+                case SelectedSubmissionListViewTab.ForReview:
+                    filteredSubmissionDataForListComponent.value = submissionsData.value
+                        .filter((submission: Submission) => {
+                            const isNewSubmission = !submission.isRejected && !submission.isApproved && !submission.isUnderReview
+                            return submission.isUnderReview || isNewSubmission
```
ermish, line 62: checking `isUnderReview` twice

ermish (changes requested):
looking really good! just a few changes and it's ready!

stores/moderationSubmissionsListStore.ts
NabbeunNabi: This is written this way to include if it is under review or if it is completely new and has never even been updated. By default they are all `false` because they haven't been checked at all. But I want both to be filtered into the array. Whether it is new or has been updated but not approved or rejected

ermish (approved):
beautiful! good breakdown of stores, components, and healthy flow of properties and state management

PR #634, opened 2024-07-05
feat: add library that helps with viewport loads

NabbeunNabi:
## 🔧 What changed
I added a library that is `VueUse`. This can use screen orientation to set variables and then improve the responsiveness of the app on mobile.

## 🧪 Testing instructions
The tests should still pass since it didn't change functionality.

## 📸 Screenshots

-   ### Before

https://github.com/ourjapanlife/findadoc-web/assets/124335161/b143baa3-c5f8-4dcd-a8a5-99ee2bfae1aa

-   ### After

https://github.com/ourjapanlife/findadoc-web/assets/124335161/d0eff02e-d1e1-4fa8-b51c-301355320df4

### DIFF
--- a/pages/index.vue
+++ b/pages/index.vue
@@ -4,7 +4,7 @@
         class="h-full w-full overflow-hidden"
     >
         <div
-            v-if="$viewport.isGreaterThan('tablet')"
+            v-if="screenOrientation?.includes('landscape')"
             class="flex h-full overflow-clip"
         >
             <LeftNavbar class="bg-primary-bg w-[358px] overflow-y-auto" />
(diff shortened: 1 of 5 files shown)


pages/index.vue
```diff
@@ -20,9 +20,19 @@
 </template>
 
 <script setup lang="ts">
-import { useNuxtApp } from '#app'
+import { useScreenOrientation } from '@vueuse/core'
```
ermish, line 23: nice! this was something I was looking to fix and you beat me to it 😆 here's some thoughts

## design improvements
- removing library -  I'm fine to use this library, but if there's a snippet option that doesn't require us to add a dependency (each one adds risk/maintenance)
- resuability - This orientation check will likely be used throughout the app, so let's make this a [composable](https://vuejs.org/guide/reusability/composables.html#mouse-tracker-example) that can be reused easily anywhere in the app. This should also prevent other devs from needlessly creating additional `watch`s for orientation.

ermish (changes requested):
see comment

pages/index.vue
NabbeunNabi: I thought about a way I can try to refactor without the library. I added it mostly due to it changing exactly the way we wanted. But I have another idea. And I'll work on making a composable...Monday

theyokohamalife:
I highly recommend testing it on an actual mobile phone. Sometimes they render differently than the browser preview.

NabbeunNabi:
@theyokohamalife The videos I included are two netlify previews of before and after showing the rendering on my mobile phone using the deployment preview with the QR code.

NabbeunNabi:
Changing this to not use a library so remaking branch and PR

PR #635, opened 2024-07-08
feat: create composable to check viewport

NabbeunNabi:
Resolves issue #640

## 🔧 What changed
I made a composable. This can use screen orientation to set variables and then improve the responsiveness of the app on mobile.

##Details
As you can see I used my phone to record the video of the deployed preview from a ticket without this change and with this change and the layout is correct instantly on the after video

## 🧪 Testing instructions
The tests should still pass since it didn't change functionality.

## 📸 Screenshots

-   ### Before

https://github.com/ourjapanlife/findadoc-web/assets/124335161/b1a664bc-9203-400a-a6ed-5018bf76db4e

-   ### After

https://github.com/ourjapanlife/findadoc-web/assets/124335161/9af45665-a7e0-4927-8ab5-0c7f19011b82

### DIFF
--- a/pages/index.vue
+++ b/pages/index.vue
@@ -4,7 +4,7 @@
         class="h-full w-full overflow-hidden"
     >
         <div
-            v-if="$viewport.isGreaterThan('tablet')"
+            v-if="!isPortrait"
             class="flex h-full overflow-clip"
         >
             <LeftNavbar class="bg-primary-bg w-96 overflow-y-auto" />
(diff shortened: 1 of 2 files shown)


composables/useScreenOrientation.ts
```diff
@@ -0,0 +1,17 @@
+import { ref, onMounted } from 'vue'
```
ermish, line 1: I think we can actually move this to the `/utils` folder since it's just an esmodule which is great! 🙌  nice simplification

ermish (changes requested):
nice

composables/useScreenOrientation.ts
NabbeunNabi: @ermish should I keep it in its own folder or just in the utils straight away. Is it worth having a composable folder in case we make some later?

ermish (approved):
Ship it!

PR #636, opened 2024-07-08
feat: implement submission update in form

### LINKED ISSUE #616: Send graphql mutation for saving updatedFacilitySubmissions
# Problem
Now that the form autofills and can show submissions that have been fetched, we need to create a function that allows the form to be submitted and saved with any updated changes.

# Success
Send successful mutations from updated forms.

### Requirements

- [x] Send successful mutations from updated facility forms.

# Implementation notes

NabbeunNabi:
Resolves #616 #637 

## 🔧 What changed
This now implements the graphQL retry from #659 into the update Submission. I have written another ticket to show the UI that it is retrying in the top nav. For now this just adds the functionality where you can mutate the submission. I wanted it to be pulled out as the functionality is more important as a start to get the Mod Panel up and running instead of a countdown message that is fun, but not necessary for `modPanel` MVP

#693 is the follow up ticket that is asking for updating the UI to show the retry countdown similar to the figma: 

## 🧪 Testing instructions
Tested manually by running the backend with (yes yarn seems to work)
```
yarn dev:startlocaldb
yarn dev
```

Then running the frontend with 
```
yarn dev:localserver
```


components/ModEditSubmissionTopbar.vue
```diff
@@ -22,11 +22,30 @@
             </button>
         </div>
         <div class="flex flex-row justify p-2 font-bold ">
+            <p
```
NabbeunNabi, line 25: This retry message appears 3 times if the request fails. To show it is retrying the mutation in 60s with a countdown timer

components/ModEditSubmissionTopbar.vue
```diff
@@ -22,11 +22,30 @@
...
             <button
                 class="flex justify-center items-center rounded-full bg-secondary-bg border-primary-text-muted
-                border-2 w-28 text-sm mr-2 "
+                border-2 w-28 text-sm mr-2"
+                data-testid="submission-topNav-saveAndExit"
+                @click="saveAndExit"
             >
-                {{ $t('modEditSubmissionTopNav.saveAndExit') }}
+                <span
+                    v-show="updatedFormBeingSentForMutation"
```
NabbeunNabi, line 39: This shows saving text. Very briefly but follows the decision from Figma. If the request takes a while it will show. Depends on speed of server response.

components/ModEditSubmissionTopbar.vue
```diff
@@ -22,11 +22,30 @@
...
+                v-show="showRetryMessage"
+                class="flex flex-row justify-center items-center whitespace-nowrap text-error text-xs mr-2"
+                data-testid="submission-topNav-retryMessage"
+            >
+                <SVGCautionSign class="w-3 h-3 mr-1" /> {{ retryMessage }}
+            </p>
             <button
                 class="flex justify-center items-center rounded-full bg-secondary-bg border-primary-text-muted
-                border-2 w-28 text-sm mr-2 "
+                border-2 w-28 text-sm mr-2"
```
NabbeunNabi, line 34: sets a state in the `ModSubmissionsStore` that calls the submit function on the `ModEditFacilitySubmissionsStore`

components/ModEditSubmissionTopbar.vue
```diff
@@ -47,13 +66,21 @@
...
 
 <script setup lang="ts">
-import { ref, type Ref } from 'vue'
+import { ref, type Ref, watch, computed, onUnmounted } from 'vue'
+import { useI18n } from 'vue-i18n'
 import SVGCopyContent from '~/assets/icons/content-copy.svg'
 import SVGSuccessCheckMark from '~/assets/icons/checkmark-square.svg'
+import SVGCautionSign from '~/assets/icons/caution-sign.svg'
 import { useModerationSubmissionsStore } from '~/stores/moderationSubmissionsStore'
+import { startCountdownTimer, type CountdownCallback } from '~/utils/countdownTimer'
```
NabbeunNabi, line 75: uses a created timer that is in a separate file in case we need timers in the future that can use a callback to set other variables in case we need to set states. This one will use a callback that will change a local state of number of seconds and if the retry message should be shown. It is translatable

components/ModEditSubmissionTopbar.vue
```diff
@@ -69,11 +96,38 @@ const copySubmissionId = async () => {
...
 }
+
+watch(moderationSubmissionStore, newValue => {
+    updatedFormBeingSentForMutation.value = newValue.isSendingUpdatedForm
+    if (newValue.didMutationFail) {
+        stopCountdownTimer = startCountdownTimer(60, 3, countdownCallback)
+    }
+})
+
+onUnmounted(() => {
```
NabbeunNabi, line 128: clears the interval on unmounted

components/ModSubmissionsEditFormFacility.vue
```diff
@@ -341,7 +346,121 @@ function submitForm(e: Event) {
...
+    }
+
+    const maxAttempts = 3
+    let mutationAttempts = 0
+    let mutationSuccess = false
+
+    function delay(milliseconds: number) {
+        return new Promise(resolve => setTimeout(resolve, milliseconds))
+    }
+
```
NabbeunNabi, line 394: this loop handles the retries and state management based on where the file is saved.

components/ModSubmissionsEditFormFacility.vue
```diff
@@ -341,7 +346,121 @@ function submitForm(e: Event) {
...
+            }
+        }
+        if (mutationAttempts === maxAttempts) {
+            moderationSubmissionStore.handleUpdateIsSendingUpdatedForm(false)
+            moderationSubmissionStore.handleUpdateSubmissionFromTopNav(false)
+            moderationSubmissionStore.handleDidMutationFail(false)
+        }
+    }
 }
+
```
NabbeunNabi, line 421: this sets up a synthetic event to mock the submit button event when the state changes on click from the Top Nav to save and exit

stores/moderationSubmissionsStore.ts
```diff
@@ -17,6 +17,22 @@ export const useModerationSubmissionsStore = defineStore(
         const selectedSubmissionId: Ref<string> = ref('')
         const selectedSubmissionData: Ref<Submission | undefined> = ref()
         const filteredSubmissionDataForListComponent: Ref<Submission[]> = ref([])
+        const updateSubmissionFromTopNav: Ref<boolean> = ref(false)
```
NabbeunNabi, line 20: adds states for manipulation between the ModEdit TopNav and the Form to submit the form

cypress/e2e/moderationDashboard.cy.ts
```diff
@@ -387,5 +387,27 @@ describe('Moderation Facility Submission Form', () => {
             cy.get('[data-testid="submission-form-mapLongitude"]').find('input').clear().type('Not Number Longitude').realPress('Tab')
             cy.get('[data-testid="submission-form-mapLongitude"]').find('p', { timeout: 10000 }).should('exist').contains('Invalid Longitude')
         })
+
```
NabbeunNabi, line 390: Test to make sure the error message for retry is shown on server error sent back

utils/countdownTimer.ts
```diff
@@ -0,0 +1,50 @@
+export type CountdownCallback = (
```
NabbeunNabi, line 1: Re-useable timer function

components/ModEditSubmissionTopbar.vue
```diff
@@ -47,13 +64,21 @@
...
 import { useModerationSubmissionsStore } from '~/stores/moderationSubmissionsStore'
+import { startCountdownTimer, type CountdownCallback } from '~/utils/countdownTimer'
+
+const { t } = useI18n()
 
-const modSubmissionStore = useModerationSubmissionsStore()
-const selectedSubmissionId: Ref<string> = ref(modSubmissionStore.selectedSubmissionId)
+const moderationSubmissionStore = useModerationSubmissionsStore()
+const selectedSubmissionId: Ref<string> = ref(moderationSubmissionStore.selectedSubmissionId)
+const updatedFormBeingSentForMutation: Ref<boolean> = ref(moderationSubmissionStore.isSendingUpdatedForm)
```
theyokohamalife, line 81: Nit: You could simplify the name to `updatedFormBeingSent`

components/ModEditSubmissionTopbar.vue
```diff
@@ -47,13 +64,21 @@
...
+const showRetryMessage: Ref<boolean> = ref(false)
+const retryMutationCountdown: Ref<number> = ref(0)
```
theyokohamalife, line 83: Nit: ♻️  `retryCountdown`

utils/countdownTimer.ts
```diff
@@ -0,0 +1,50 @@
...
+    isFinished: boolean
+) => void
+
+export const startCountdownTimer = (
+    seconds: number,
+    chosenAmountOfTimesToRepeat: number,
+    callback: CountdownCallback
+) => {
+    let timerRepeatCount = 0
+    let counterInterval:
```
theyokohamalife, line 12: What do you think about initializing `counterInterval` as `0`? That could simplify the type and the conditional could check if it's above `0`.

theyokohamalife (approved):
Looks good! I left a couple non-blocking comments. You might want to wait for feedback from @ermish as well before merging.

i18n/locales/en.json
```diff
@@ -125,6 +125,8 @@
     },
     "modEditSubmissionTopNav": {
         "saveAndExit": "Save & Exit",
+        "saving": "Saving...",
+        "retryMessage": "Something’s not right. Retrying in {seconds}s...",
```
ermish, line 129: I'm a little nervous about the variable being in the translation string here vs just splitting the string into two parts.  Personally, I would change it, but I'll allow it if you prefer this way

ermish: reason: separate strings prevents any mistakes where a future dev might not be aware there's a variable, or forget to put in in one of the translations which would break the code. Explicit sentence parts makes it apparent what is being built.

stores/moderationSubmissionsStore.ts
ermish: when using a state store, you should keep your data abstracted from the usage of it. 
ex. instead of calling data `firstTabData` `secondTabData`, it should be `facilitiesData` `profData`. 

So in this case, we should remove TopNav and  make it clear what it's doing.

stores/moderationSubmissionsStore.ts
```diff
@@ -17,6 +17,22 @@ export const useModerationSubmissionsStore = defineStore(
...
+        const isSendingUpdatedForm: Ref<boolean> = ref(false)
+        const didMutationFail: Ref<boolean> = ref(false)
+        const mutationAttempts: Ref<number> = ref(0)
+
+        function handleUpdateSubmissionFromTopNav(newValue: boolean) {
+            updateSubmissionFromTopNav.value = newValue
+        }
+
+        function handleUpdateIsSendingUpdatedForm(newValue: boolean) {
```
ermish, line 29: these handles aren't really clear what they're handling. Usually `handle` is used for handling an event. In this case, these just look like `setX` functions

stores/moderationSubmissionsStore.ts
```diff
@@ -17,6 +17,22 @@ export const useModerationSubmissionsStore = defineStore(
...
+            isSendingUpdatedForm.value = newValue
+        }
+
+        function handleDidMutationFail(newValue: boolean) {
+            didMutationFail.value = newValue
```
ermish, line 34: maybe these could be combine into a `formSubmissionStatus` with `sending`, `failed`, etc.

ermish (changes requested):
dropped a few comments

i18n/locales/en.json
NabbeunNabi: When it comes to splitting it would it still easily be clear with seconds being separated from the from part. It is just for the s or 秒

stores/moderationSubmissionsStore.ts
NabbeunNabi: @ermish I will make this a draft again since I am going to wait for the retry logic from the rejection usable function to be implemented

components/ModSubmissionsEditFormFacility.vue
```diff
@@ -353,7 +358,99 @@ function submitForm(e: Event) {
...
-    //submitthedatahere(googlemapsURL, mapLatitude, mapLongitude)
-    console.log('Form submitted!!! woohoo')
+    const updateSubmissionId = formSubmissionId || ''
+
+    if (!updateSubmissionId) {
+        console.error('Facility ID is required for updating the facility')
+        return
+    }
+
+    const facilityInputVariables = {
```
ermish, line 368: this should definitely be a `type`

ermish (changes requested):
looking muuuch cleaner after the retry logic updates

components/ModSubmissionsEditFormFacility.vue
NabbeunNabi: yea you right. Unfortunately since it is a type with the property `id` I cannot use a more descriptive variable name such as `spoon`

ermish (approved):
ship it!

PR #649, opened 2024-07-12
FIX(UI): overlapping modal with footer for desktop users

### LINKED ISSUE #213: SearchResultDetails -  consistent modal sizing across screen sizes
# Context

The modal is overlapping/out of the window when the height is too narrow! 

# Requirements
- [x] The model should fit all the content
- [x] The results should be clickable while the modal is open (might require making the results list height reduced when the modal is open
- [ ] (optional) you can see the map pin 📍 when the modal is open to help the user associate the results with the location. I think this will take more work though and probably a separate issue

# Dependencies

dominiksakic:
Resolves #213 

## 🔧 What changed
The Modal (bigger than tablet mode)
a)is not overlapping with the footer and 
b)scrollable IF the winodw height is too narrow

## 🧪 Testing instructions
1. Open browser
2. open dev tools
3. set the website to a screensize bigger than tablet
4. narrow the height 

## 📸 Screenshots

-   ### Before
スクリーンショット 2024-07-11 14 57 19

-   ### After

https://github.com/user-attachments/assets/600e7745-d98a-44b9-b414-f301c97f7b3c

### DIFF
--- a/components/MainContentContainer.vue
+++ b/components/MainContentContainer.vue
@@ -3,12 +3,12 @@
         <div
             v-if="$viewport.isGreaterThan('tablet')"
             id="landscape-content"
-            class="bg-secondary-bg/20 hover:shadow-inner hover:shadow-secondary-bg/90 h-full w-full"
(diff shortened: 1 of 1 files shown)


ermish (approved):
nice! ship it!

PR #657, opened 2024-07-16
feat(lang): implement key check for locale files on push

### LINKED ISSUE #655: Implement script to check locales keys exist in all `json` files for language
# Problem
Whenever we add internationalization to a file. We want to make sure the keys exist and match within all the files. This will prevent us from having to manually check in the future.

# Success
We are not allowed to push if any keys mismatch
### Requirements

- [ ] Implement function that checks to see if the json keys are the same in locale files
- [ ] Add to the pre-commit hook to check and fail if the  keys aren't the same

NabbeunNabi:
Resolves #655

## 🔧 What changed
Before it was possible to add translations in the English file, but not add it to other files. Now there is a script that will run on `prepush`. This script checks to make sure all keys are present in the other files in locales as in `en.json`. It auto adds them and will log what it added.

## 🧪 Testing instructions
To test I would go to this branch and delete random keys from non English files and run the following command. **DO NOT COMMIT THE CHANGED FILES AS THEY WON'T BE THE ORIGINAL TRANSLATIONS**
```
yarn lint:locales
```
 See that it adds those missing keys.
You can then push those added keys to your test branch

## 📸 Screenshots

-   ### Before
N/A just no errors on push

-   ### After
image


i18n/checkLocaleKeys.js
```diff
@@ -0,0 +1,73 @@
...
+    '/locales'
+)
+
+// Read all JSON files from the locales directory
+let localeFiles = fs
+    .readdirSync(localesI18nDirectory)
+    .filter(file => file.endsWith('.json'))
+
+// Ensure en.json is the first file
+const enFile = 'en.json'
```
ermish, line 21:
```suggestion
const enFileName = 'en.json'
```

i18n/checkLocaleKeys.js
```diff
@@ -0,0 +1,73 @@
...
+if (!localeFiles.includes(enFile)) {
+    console.error('en.json file not found in the locales directory')
+    process.exit(1)
+}
+localeFiles = localeFiles.filter(file => file !== enFile)
+localeFiles.unshift(enFile)
```
ermish, line 27: instead of unshifting, let's just keep the `enFile` as a separate variable to make it more readable

i18n/checkLocaleKeys.js
```diff
@@ -0,0 +1,73 @@
...
+
+// This function extracts the keys from the JSON object
+const getKeys = (obj, prefix = '') =>
```
ermish, line 30:
```suggestion
const getAllKeys = (obj, prefix = '') =>
```

i18n/checkLocaleKeys.js
```diff
@@ -0,0 +1,73 @@
...
+// Ensure en.json is the first file
+const enFile = 'en.json'
+if (!localeFiles.includes(enFile)) {
+    console.error('en.json file not found in the locales directory')
+    process.exit(1)
+}
+localeFiles = localeFiles.filter(file => file !== enFile)
+localeFiles.unshift(enFile)
+
+// This function extracts the keys from the JSON object
```
ermish, line 29: let's enhance this comment to clarify this function gets all keys, including nested keys

i18n/checkLocaleKeys.js
```diff
@@ -0,0 +1,73 @@
...
+        if (typeof obj[key] === 'object' && obj[key] !== null) {
+            keys.push(...getKeys(obj[key], fullKey))
+        } else {
+            keys.push(fullKey)
+        }
+        return keys
+    }, [])
+
+// This reads and parses the JSON files
+const localeKeys = localeFiles.reduce((acc, file) => {
```
ermish, line 42: optimization time! since we're iterating over the localfiles + the keys in each file, we have O(n+m)^2 (I think). Instead, we can make this just O(n+m).
Basically, you're looping over the localFiles + the keys multiple times. Instead, just loop over the localeFiles and keys in one big reduce function. When you pull each key, check it, then add it to the `missingKeys` object if it's missing. 

```
.reduce((missingKeys, file) => { 
...
missingKeys[fileName].push(key)
...
  })
```

i18n/insertMissingLocaleKeys.js
```diff
@@ -0,0 +1,94 @@
...
+            }
+        }
+    })
+
+    // Insert missing keys
+    const insertMissingKeys = (jsonContent, missingKeys, enContent) => {
+        missingKeys.forEach(key => {
+            const value = key
+                .split('.')
+                .reduce((obj, k) => obj?.[k], enContent)
```
ermish, line 61: what's `k`?

i18n/insertMissingLocaleKeys.js
```diff
@@ -0,0 +1,94 @@
...
+        .readdirSync(localesI18nDirectory)
+        .filter(file => file.endsWith('.json'))
+
+    const enFile = 'en.json'
+    const enFilePath = path.join(localesI18nDirectory, enFile)
+    const enContent = JSON.parse(fs.readFileSync(enFilePath, 'utf-8'))
+
+    // Extract keys and find missing ones
+    let missingKeys = {}
+    localeFiles.forEach(file => {
```
ermish, line 27: you could do this in a separate PR: 
I think this could be written a little bit simpler. 
1) iterate over the en file keys. 
2) forEach the files and traverse to the key. If it's missing, add the key/value. 
3) done. so easy. 😆  (It's not that easy)

package.json
```diff
@@ -4,6 +4,8 @@
     "private": true,
     "scripts": {
         "prepare": "husky",
+        "check-locales": "node i18n/checkLocaleKeys.js",
```
ermish, line 7: should group these under something like `lint:locales`

ermish (changes requested):
this is awesome! I think some of the code is pretty tough to read, but the functionality is amazing!

i18n/checkLocaleKeys.js
```diff
@@ -0,0 +1,90 @@
...
+        })
+        return keys
+    }
+
+    // Function to insert a key-value pair into a nested object
+    const insertMissingKey = (targetObject, keyPath, value) => {
+        const keyParts = keyPath.split('.')
+        let currentLevel = targetObject
+        keyParts.forEach((part, index) => {
+            if (index === keyParts.length - 1) {
```
ermish, line 47: nit: split into variable

i18n/checkLocaleKeys.js
```diff
@@ -0,0 +1,90 @@
...
+    const localeFilesWithoutEnFileName = fs
+        .readdirSync(localesI18nDirectory)
+        .filter(localeFile => localeFile.endsWith('.json') && localeFile !== 'en.json')
+
+    const enFileName = 'en.json'
+    const enFilePath = path.join(localesI18nDirectory, enFileName)
+    const enFileContent = JSON.parse(fs.readFileSync(enFilePath, 'utf-8'))
+
+    // Function to create a full key path
+    const createFullKey = (prefix, key) => prefix ? `${prefix}.${key}` : key
```
ermish, line 23: personal preference: I would move these functions out so they're not nested. Since there's so many functions being defined and running inside `checkLocaleKeys` it makes it harder to track the function scope that's running

i18n/checkLocaleKeys.js
```diff
@@ -0,0 +1,90 @@
...
+    // Function to check if value is an object
+    const isNestedObject = value => typeof value === 'object' && value !== null
+
+    // Extract all keys from a locale file
+    const extractAllKeys = (jsonObject, prefix = '') => {
+        const keys = []
+        Object.keys(jsonObject).forEach(key => {
+            const fullKey = createFullKey(prefix, key)
+            if (isNestedObject(jsonObject[key])) {
+                keys.push(...extractAllKeys(jsonObject[key], fullKey))
```
ermish, line 34: nice!

i18n/checkLocaleKeys.js
```diff
@@ -0,0 +1,90 @@
...
+            } else {
+                keys.push(fullKey)
+            }
+        })
+        return keys
+    }
+
+    // Function to insert a key-value pair into a nested object
+    const insertMissingKey = (targetObject, keyPath, value) => {
+        const keyParts = keyPath.split('.')
```
ermish, line 44:
```suggestion
        const keyPathParts = keyPath.split('.')
```

ermish: ^ I mistook this for key value pair and  then realize they were parts of the path, so would be good to clarify

i18n/checkLocaleKeys.js
```diff
@@ -0,0 +1,90 @@
...
+                    currentLevel[part] = {}
+                }
+                currentLevel = currentLevel[part]
+            }
+        })
+    }
+
+    const referenceKeys = extractAllKeys(enFileContent)
+
+    localeFilesWithoutEnFileName.forEach(file => {
```
ermish, line 62: should functionize it for readability to explain what this foreach is doing

i18n/checkLocaleKeys.js
```diff
@@ -0,0 +1,90 @@
...
+                keys.push(...extractAllKeys(jsonObject[key], fullKey))
+            } else {
+                keys.push(fullKey)
+            }
+        })
+        return keys
+    }
+
+    // Function to insert a key-value pair into a nested object
+    const insertMissingKey = (targetObject, keyPath, value) => {
```
ermish, line 43:
```suggestion
    const insertMissingKeysAndValues = (valueToSet, keyPath, value) => {
```

This is also setting the values, so we should clarify the name. Also targetObject was a bit confusing

i18n/checkLocaleKeys.js
```diff
@@ -0,0 +1,90 @@
...
+
+        const existingKeys = new Set(extractAllKeys(jsonContent))
+        let fileUpdated = false
+        const missingKeys = []
+
+        referenceKeys.forEach(key => {
+            if (!existingKeys.has(key)) {
+                const value = key.split('.').reduce((obj, part) => obj?.[part], enFileContent)
+                if (value !== undefined) {
+                    insertMissingKey(jsonContent, key, value)
```
ermish, line 74: should add either some comments here and function names around the fact that we're setting the other language files missing values to english values.

ermish (changes requested):
muuuuch better! Still got a few readability improvements, but the performance looks good now 🏁

i18n/checkLocaleKeys.js
NabbeunNabi: Pulled out a bunch of nested functions and added JSDoc for the difficult one

ermish (approved):
passed! This looks really clean now 💯  
I was able to read through the whole file in less than 2 mins. you've taken complex code and made it readable, well done

ermish:
@NabbeunNabi since it's not guaranteed that someone runs husky hooks before a PR, we should make a CI step that runs your new lint command. Let's make an issue for this 🗒️

PR #659, opened 2024-07-18
feat: implement mutation rejection for facility submission

### LINKED ISSUE #639: Implement logic to reject a facility submission
BLOCKED BY #637 

# Problem
We need to be able to reject a submission

# Success
This should make us able to reject a facility submission.

### Requirements

- [ ] Implement logic to reject a facility submission

NabbeunNabi:
Resolves #639 
Related to #636

## 🔧 What changed
This implements the functionality of the reject button to reject a facility submission. Before it just logged. Now if I click, the facility will be rejected and moved in tabs.

It also contains a custom function that will allow us to pass `GQL requests` and retry recursively.

## 🧪 Testing instructions

1.  Run the local backend 

```
yarn dev:startlocaldb
```
```
yarn dev
```

2. Then start the frontend 
```
yarn dev:localserver
```

3. Reject a submission and you can see how it changes as shown in the After screenshot

## 📸 Screenshots

-   ### Before

-   ### After

This is the submission list before I reject Stephanie Mills

image

and after with the rejection seen in the separate tab
image

image

### DIFF
--- a/components/ModEditSubmissionTopbar.vue
+++ b/components/ModEditSubmissionTopbar.vue
@@ -48,12 +48,14 @@
 
 <script setup lang="ts">
 import { ref, type Ref } from 'vue'
+import { gql } from 'graphql-request'
 import SVGCopyContent from '~/assets/icons/content-copy.svg'
 import SVGSuccessCheckMark from '~/assets/icons/checkmark-square.svg'
 import { useModerationSubmissionsStore } from '~/stores/moderationSubmissionsStore'
+import { gqlClient, graphQLClientRequestWithRetry } from '~/utils/graphql'
 
-const modSubmissionStore = useModerationSubmissionsStore()
-const selectedSubmissionId: Ref<string> = ref(modSubmissionStore.selectedSubmissionId)
+const moderationSubmissionStore = useModerationSubmissionsStore()
+const selectedSubmissionId: Ref<string> = ref(moderationSubmissionStore.selectedSubmissionId)
 
 const showCopySuccessIcon: Ref<boolean> = ref(false)
 
@@ -73,7 +75,31 @@ const acceptSubmission = () => {
     console.log('You are accepted')
 }
 
-const rejectSubmission = () => {
-    console.log('You are rejected')
+const rejectSubmission = async () => {
+    const facilityInputVariables = {
+        updateSubmissionId: selectedSubmissionId.value,
+        input: {
+            isRejected: true
+        }
+    }
+
+    try {
+        await graphQLClientRequestWithRetry(
+            gqlClient.request.bind(gqlClient),
+            rejectFacilitySubmissionGqlMutation,
+            facilityInputVariables
+        )
+        moderationSubmissionStore.setDidMutationFail(false)
+    } catch (error) {
+        console.error('Failed to reject submission:', error)
+        moderationSubmissionStore.setDidMutationFail(true)
+    }
 }
+
+const rejectFacilitySubmissionGqlMutation = gql`
+mutation Mutation($updateSubmissionId: ID!, $input: UpdateSubmissionInput!) {
+  updateSubmission(id: $updateSubmissionId, input: $input) {
+    isRejected
+  }
+}`
 </script>
--- a/stores/moderationSubmissionsStore.ts
+++ b/stores/moderationSubmissionsStore.ts
@@ -32,6 +32,11 @@ export const useModerationSubmissionsStore = defineStore(
         const selectedHealthcareProfessionalId: Ref<string> = ref('')
         const selectedSubmissionData: Ref<Submission | undefined> = ref()
         const filteredSubmissionDataForListComponent: Ref<Submission[]> = ref([])
+        const didMutationFail: Ref<boolean> = ref(false)
+
+        function setDidMutationFail(newValue: boolean) {
+            didMutationFail.value = newValue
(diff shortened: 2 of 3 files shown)


components/ModEditSubmissionTopbar.vue
```diff
@@ -73,7 +75,41 @@ const acceptSubmission = () => {
...
+
+    const maxAttempts = 3
+    let mutationAttempts = 0
+    let mutationSuccess = false
+
+    function delay(milliseconds: number) {
+        return new Promise(resolve => setTimeout(resolve, milliseconds))
+    }
+
+    while (mutationAttempts < maxAttempts && !mutationSuccess) {
```
ermish, line 94: love the retries! Since retries are a common pattern, usually most query libraries already have them built in. Let's see if our gql client has them.
We should also move the retry logic to the gqlclient so every request has the retry logic and can overridden through optional parameters. Otherwise, we're at the goodwill of other devs to write the same retry logic you have here

components/ModEditSubmissionTopbar.vue
```diff
@@ -48,12 +48,14 @@
 
 <script setup lang="ts">
 import { ref, type Ref } from 'vue'
+import { gql } from 'graphql-request'
 import SVGCopyContent from '~/assets/icons/content-copy.svg'
 import SVGSuccessCheckMark from '~/assets/icons/checkmark-square.svg'
 import { useModerationSubmissionsStore } from '~/stores/moderationSubmissionsStore'
+import { gqlClient } from '#imports'
```
ermish, line 55: let's use an explicit path so we know where it's coming from. It will be easier for devs reading to understand where it comes from

ermish (changes requested):
left some comments!

components/ModEditSubmissionTopbar.vue
NabbeunNabi: Now we will have our own built in

components/ModEditSubmissionTopbar.vue
```diff
@@ -73,7 +75,37 @@ const acceptSubmission = () => {
...
+        }
+    }
+
+    const rejectSubmission = async () => {
+        try {
+            await graphQLClientRequestWithRetry(
+                gqlClient.request.bind(gqlClient),
+                rejectFacilitySubmissionGqlMutation,
+                facilityInputVariables,
+                60000,
```
ermish, line 92: let's make an `options` object to make these variables more readable. We can make them be optional and have a default value.

utils/graphql.ts
```diff
@@ -22,6 +22,31 @@ export const initializeGqlClient = () => {
...
+
+    const executeGQLClientRequest = async (): Promise<T> => {
+        try {
+            return await gqlClientRequestFunction(queryOrMutation, variables)
+        } catch (error) {
+            if (attempts < retryAmount) {
+                attempts++
+                await new Promise(resolve =>
+                    setTimeout(resolve, requestTimeoutInMilliseconds))
+                return executeGQLClientRequest()
```
ermish, line 42: heyyy a clear, readable example of recursion! nice work :)

stores/moderationSubmissionsStore.ts
```diff
@@ -17,6 +17,11 @@ export const useModerationSubmissionsStore = defineStore(
         const selectedSubmissionId: Ref<string> = ref('')
         const selectedSubmissionData: Ref<Submission | undefined> = ref()
         const filteredSubmissionDataForListComponent: Ref<Submission[]> = ref([])
+        const didMutationFail: Ref<boolean> = ref(false)
+
+        function handleDidMutationFail(newValue: boolean) {
```
ermish, line 22: `handle` is a standard convention for a function handling events. Since this is just setting the value, let's rename it
```suggestion
        function setDidMutationFail(newValue: boolean) {
```

ermish (changes requested):
woo recursion! much better now that it's extracted out and reusable 👍

NabbeunNabi:
Ok @ermish. Added all of your feedback! Hopefully it is good to go now

components/ModEditSubmissionTopbar.vue
```diff
@@ -73,7 +75,39 @@ const acceptSubmission = () => {
...
+const rejectSubmission = async () => {
+    const facilityInputVariables = {
+        updateSubmissionId: selectedSubmissionId.value,
+        input: {
+            isRejected: true
+        }
+    }
+
+    const rejectSubmissionRetryOptions: graphQLClientRequestWithRetryOptions = {
+        requestTimeoutInMilliseconds: 60000,
```
ermish, line 87: any reason why we're not just using the default values here that you set? If there is, you should add a comment to explain

ermish (changes requested):
almost there!

components/ModEditSubmissionTopbar.vue
NabbeunNabi: I am not using the default values I set because I was following the logic shown on the Figma. I really am ok with either. But the retry shown in the Figma started at 60 seconds for the mod panel. I thought in other places  5 and 3 is normal But. Honestly 60 seconds is also too long on the `modPanel`. Mods don't want to wait that long either. I will use the default values and remove this. We can always go back and change based on user needs.

ermish (approved):
after a few retries, this looks great!

PR #663, opened 2024-07-19
test: improve test speed for ci/cd pipeline

### LINKED ISSUE #662: TEST: Speed up testing for CI/CD
# Problem
CI/CD takes a long time due to the use of `cy.wait`. We want to implement a way to use better methods to speed up the testing.

# Success
CI/CD is sped up

### Requirements

- [x] Speed up cypress tests for CI/CD
- [x] Use build in viewports from Cypress

NabbeunNabi:
Resolves #662 

## 🔧 What changed
The tests were using `cy.wait` to make sure pages were visited and components had time to render. Now the build in timeout method is used. This is designed by Cypress in order to move on right when the test passes. `cy.wait` Takes the whole time.

## 🧪 Testing instructions
Build the production locally and run the tests on this branch. Check the timing. Check that also on another branch and compare the times.

## 📸 Screenshots

-   ### Before
This is from another branch I recently re-ran
image

-   ### After
This if from the new CI/CD

image

### DIFF
--- a/cypress/e2e/about.cy.ts
+++ b/cypress/e2e/about.cy.ts
@@ -3,16 +3,13 @@ import enUS from '../../i18n/locales/en.json'
 describe('About page', () => {
     context('Desktop resolution', () => {
         before(() => {
-            cy.visit('/about')
-            // This wait time is to give the page time to load from Prod when ran in CI.
-            cy.wait(3000)
+            cy.visit('/about', { timeout: 10000 })
         })
 
         beforeEach(() => {
             // The resolution is in the beforeEach() instead of before() to
             // prevent Cypress from defaulting to other screen sizes between tests.
-            cy.viewport(1920, 1080)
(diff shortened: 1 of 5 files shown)


cypress/e2e/about.cy.ts
```diff
@@ -3,16 +3,13 @@ import enUS from '../../i18n/locales/en.json'
 describe('About page', () => {
     context('Desktop resolution', () => {
         before(() => {
-            cy.visit('/about')
-            // This wait time is to give the page time to load from Prod when ran in CI.
-            cy.wait(3000)
+            cy.visit('/about', { timeout: 10000 })
```
ermish, line 6: 👏 soooo happy to see a timeout instead! this is much better than artifical wait times! nice find

ermish (approved):
🚀  to the mooon

PR #665, opened 2024-07-21
fix: change intercept order to confirm route registration

### LINKED ISSUE #664: BUG: Tests intercept not registering properly
# Problem
In fixing some of the tests and the speed the intercept isn't registering properly on a consistent basis. Fix this

# Success
Intercept registers in a more consistent manner
### Requirements

- [x] Intercept registers in a more consistent manner

NabbeunNabi:
Resolves #664

## 🔧 What changed
In the tests the intercept was being registered on the return from the auth login. But this could possible happen before or after the website in production had return to `localhost:3000`. Now the url is checked with a timeout to ensure that the intercept can register properly on the frontend.

## 🧪 Testing instructions
Tests run successfully

## 📸 Screenshots
N/A

### DIFF
--- a/cypress/e2e/moderationDashboard.cy.ts
+++ b/cypress/e2e/moderationDashboard.cy.ts
@@ -133,6 +133,8 @@ describe(
                     cy.get('[data-action-button-primary]').should('be.visible').click()
                 })
 
+                cy.url({ timeout: 10000 }).should('equal', 'http://localhost:3000/')
+
                 cy.intercept('POST', '**/', req => {
                     req.continue(res => {
                         if (req.body.query && req.body.query.includes('query Submissions')) {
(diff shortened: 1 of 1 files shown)


cypress/e2e/moderationDashboard.cy.ts
```diff
@@ -241,6 +243,8 @@ describe('Moderation Facility Submission Form', () => {
                 cy.get('[data-action-button-primary]').should('be.visible').click()
             })
 
+            cy.url({ timeout: 10000 }).should('equal', 'http://localhost:3000/')
```
ermish, line 246: This URL needs to be adapted for CI as well

ermish: N/m this works! Forgot CI was localhost too

ermish (approved):

Issue #681, opened 2024-08-01
[FIX] Implement fix for flaky submission test on error message showing for CI/CD


NabbeunNabi:
# Problem
Sometimes on CI/CD the submission error message showing fails. It needs to be updated or changed.

# Success
Fix the `submit.cy.ts` file to have the flaky test with the error message sometimes still showing to not fail.

### Requirements

- [ ] Fix test

PR #682, opened 2024-08-01
fix: refactor tests to lessen ci/cd failures

### LINKED ISSUE #674: FIX: Cypress intercept not always registering
# Problem
Tests will occasionally fail due to the intercept within the `moderationDashboard.cy.ts` does not always register on the redirect

# Success
On CI/CD you can run the test 4 times with cypress and it doesn't fail to register
### Requirements

- [ ] Fix cypress test not always registering the intercept

NabbeunNabi:
Resolves #674 and #681 

## 🔧 What changed
Timeouts were added to different tests as well as checks and location of the intercept for registering it in cypress.
Along with this there was a change in the `yml` file structure. It now runs the build and installation together. Then the start and waits on the server so the timeout doesn't happen connecting to the server.

## 🧪 Testing instructions
You can run the CI/CD on this branch as many times as you want. It takes a little longer by about 40 -55 seconds, but this keeps the tests from being flaky and failing which requires rerunning of tests for 3 different possible failings.

## Context
I ran the test in CI/CD 10 times and it did not fail


cypress/e2e/moderationDashboard.cy.ts
```diff
@@ -157,6 +157,17 @@ describe(
                 cy.visit('/login', { timeout: 10000 })
                 Cypress.session.clearCurrentSessionData()
 
+                cy.intercept('POST', '**/', req => {
```
ermish, line 160: I'm a newbie dev! help! I don't understand what's happening without any comments 😱

cypress/e2e/moderationDashboard.cy.ts
```diff
@@ -288,6 +288,17 @@ describe('Moderation Facility Submission Form', () => {
             cy.visit('/login', { timeout: 10000 })
             Cypress.session.clearCurrentSessionData()
 
+            cy.intercept('POST', '**/', req => {
```
ermish, line 291: ♻️

ermish (changes requested):
just slap on some comments and you're good to go!

ermish (approved):
Ship it! 🚢

PR #704, opened 2024-08-19
fix: add id missing in update submission

### LINKED ISSUE #701: FIX: No id provided when updating a submission
# Problem
When clicking on the "Save & Exit" button, an error occurs due to a missing ID.

# Success
Successfully make an update submission request without encountering the missing ID error.
### Requirements

- [x] Debug and resolve the issue causing the missing ID error when clicking "Save & Exit."

NabbeunNabi:
Resolves #701

## 🔧 What changed
Before the `ModEditSubmissionTopBar` was not working for Save & Exit. As well as this the submission itself wasn't working due to the variable name not matching what was expected. Now it is updated to Save and Exit successfully from the Top Bar. Otherwise if you press enter or just submit it doesn't automatically exit so you can keep working on the form.

## 🧪 Testing instructions
Go to the backend and start the localdb with
```
yarn dev:startlocaldb
```
and start the backend with
```
yarn dev
```

Then start the frontend with 
``` 
yarn dev:localserver
```

## 📸 Screenshots

-   ### Before
N/A
-   ### After

https://github.com/user-attachments/assets/f5124c5a-7554-4a7a-aa70-d6c7f2449978

### DIFF
--- a/components/ModEditSubmissionForm.vue
+++ b/components/ModEditSubmissionForm.vue
@@ -662,10 +662,14 @@ async function submitForm(e: Event) {
             updateFacilitySubmissionGqlMutation,
             submissionInputVariables
         )
-        router.push('/moderation')
+        if (moderationSubmissionStore.updatingMutationFromTopBar) {
+            moderationSubmissionStore.setUpdatingMutationFromTopBar(false)
+            router.push('/moderation')
+        }
     } catch (error) {
         console.error('Failed to update submission:', error)
         moderationSubmissionStore.setDidMutationFail(true)
+        moderationSubmissionStore.setUpdatingMutationFromTopBar(false)
     }
 }
 
--- a/components/ModEditSubmissionForm.vue
+++ b/components/ModEditSubmissionForm.vue
@@ -709,8 +709,8 @@ onMounted(() => {
 })
 
 const updateFacilitySubmissionGqlMutation = gql`
-mutation Mutation($updateSubmissionId: ID!, $input: UpdateSubmissionInput!) {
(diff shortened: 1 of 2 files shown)


components/ModEditSubmissionTopbar.vue
```diff
@@ -76,7 +76,7 @@ const copySubmissionId = async () => {
 }
 
 const saveAndExit = () => {
-    moderationSubmissionStore.setUpdatingMutationFromTopBar(true)
+    moderationSubmissionStore.setUpdatingMutationFromTopBar(!moderationSubmissionStore.updatingMutationFromTopBar)
```
ermish, line 79: Passing a state into its own set method doesn't really make sense 

Might be good to add a comment explain this and see if there's a better way to code this

ermish (changes requested):
Added comments

components/ModEditSubmissionTopbar.vue
NabbeunNabi: I think I was stuck in React mode since it is common to do `prev => !prev`

NabbeunNabi: Updated

NabbeunNabi: I didn't add a comment as I am not passing the state into its own set method anymore. just the boolean itself again. So I do not think a comment would be necessary anymore

theyokohamalife (approved):
LGTM ✨

ermish (approved):
ID ship it!

PR #718, opened 2024-08-30
feat: implement submission acceptance

### LINKED ISSUE #697: FEAT: Add healthcare professional to submitForm function
Depends On: #597 

# Problem
When submitting our form, it should create both a facility and a healthcare professional. The healthcare professional can only be created after the facility has been successfully created, as it depends on the facility ID.

# Success
Upon form submission, the system should successfully create both the facility and the healthcare professional.

### Requirements

- [x] Add a graphQL query for healthcare professional
- [x] Modify the form submission process to first create a facility, and upon success, use the facility ID to create a healthcare professional.

NabbeunNabi:
Resolves #697

## 🔧 What changed
Functionality to approve facilities and healthcare professionals was added.

## 🧪 Testing instructions
There were tests written to make sure the components were found and shown correctly in the `moderation.cy.ts` file.

Manual instructions would be to run the local db and server on the backend and then to run the frontend server with
```
yarn dev:localserver
```
Then go to the mod panel and click on a submission. Try to update it or submit it without the forms filled in correctly or updated. This will throw a toast notification. The video shows it successfully being submitted.

## 📸 Screenshots

-   ### Before
There was no functionality to show
-   ### After

https://www.loom.com/share/25ca3c7f5ecf471cac5ccc9584057344


utils/triggerFormValidationErrorMessages.ts
```diff
@@ -0,0 +1,9 @@
+export const triggerFormValidationErrorMessages = (className: string) => {
+    //We are going to find all the inputs from the form to blur them to show errors for submitted doctors for the admin users
+    const modInputFieldElements = Array.from(document.body.getElementsByClassName(className))
+    //This will foucs and then blur each input so that we can see the error messages on validation and in this case the semicolon is necessary for the function to run correctly and have no errors
```
theyokohamalife, line 4:
```suggestion
    //This will focus and then blur each input so that we can see the error messages on validation and in this case the semicolon is necessary for the function to run correctly and have no errors
```

nuxt.config.js
```diff
@@ -85,6 +85,11 @@ export default defineNuxtConfig({
         link: [{ rel: 'icon', type: 'image/x-icon', href: '/assets/favicon.ico' }]
     },
 
+    //On build this will tranpile the module to be available in deployment
```
theyokohamalife, line 88:
```suggestion
    //On build this will transpile the module to be available in deployment
```

assets/icons/person-icon.svg
ermish, line 1: we already have an icon like this. See the results for an example

components/ModEditSubmissionForm.vue
```diff
@@ -604,86 +609,216 @@ function autofillEditSubmissionForm(submissionData: Submission | undefined) {
...
+            updateSubmissionGqlMutation,
             submissionInputVariables
         )
-        if (moderationSubmissionStore.updatingMutationFromTopBar) {
-            moderationSubmissionStore.setUpdatingMutationFromTopBar(false)
+        if (moderationSubmissionStore.updatingSubmissionFromTopBar) {
+            moderationSubmissionStore.setUpdatingSubmissionFromTopBar(false)
             router.push('/moderation')
         }
+        autofillEditSubmissionForm(result.updateSubmission)
```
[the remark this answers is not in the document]
ermish: I gotta head to bed at this point 😬  I'll take a look at the second half tomorrow

components/ModEditSubmissionForm.vue
```diff
@@ -303,52 +288,47 @@
...
-                    class="grid grid-cols-subgrid col-span-4"
+                    v-for="(healthcareProfessionalName, index) in healthCareProfessionalNameArray"
+                    :key="`${healthcareProfessionalName.firstName}-${index}`"
+                    class="flex flex-row basis-1/3 justify-start items-center p-2"
                 >
                     <div
-                        v-for="(healthcareProfessionalName, index) in healthCareProfessionalNameArray"
-                        :key="`${healthcareProfessionalName.firstName}-${index}`"
-                        class="grid grid-cols-subgrid col-span-4 bg-tertiary-bg"
+                        class="flex flex-row justify-center items-center rounded-lg p-3  border-2 border-primary"
```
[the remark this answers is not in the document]
NabbeunNabi: I changed the styling completely. It was a table before. But now it is little boxes of people more like what was shown in the Figma. So it was just due to the updated styling. It now looks like this
image

components/ModEditSubmissionForm.vue
```diff
@@ -303,52 +288,47 @@
...
+                            <div class="flex flex-row font-bold">
+                                <SVGProfileIcon class="flex items-center justify-center w-6 h-6" />
+                                <span>{{ healthcareProfessionalName.lastName }}</span>
+                                <span class="mx-2">{{ healthcareProfessionalName.firstName }}</span>
+                                <span v-show="healthcareProfessionalName.middleName">
+                                    {{ healthcareProfessionalName.middleName }}
+                                </span>
+                            </div>
+                            <div
+                                class="w-fit px-2 py-[1px] my-2 border border-primary/40 rounded-full shadow text-sm text-center
```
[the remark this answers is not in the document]
NabbeunNabi: With grow it looks strange:

## grow
image


## fit
image

components/ModEditSubmissionForm.vue
```diff
@@ -561,22 +556,32 @@ const validateFields = () => {
...
+      && isAddressLine1EnValid && isAddressLine2EnValid && isAddressLine1JpValid && isAddressLine2JpValid && isCityEnValid
+      && isCityJpValid && isPostalCodeValid && isLatitudeValid && isLongitudeValid
+        ? true
+        : false
+
+    return areAllTheFacilityFieldsValid
+}
+
+const validateHealthcareProfessionalFields = () => {
+    const areHealthcareProfessionalNamesRegisteredToFacility: boolean
```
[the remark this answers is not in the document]
NabbeunNabi: 😅

components/ModEditSubmissionForm.vue
```diff
@@ -604,86 +609,216 @@ function autofillEditSubmissionForm(submissionData: Submission | undefined) {
                 case 'googleMapsUrl':
                     googlemapsURL.value = submissionData['facility']?.contact?.googleMapsUrl || submissionData['googleMapsUrl']
                     break
+                case 'healthcareProfessionals':
+                    healthCareProfessionalNameArray.value = submissionData?.healthcareProfessionals?.[0]?.names
+                    ?? [{
+                        firstName: submissionData?.healthcareProfessionalName?.split(' ')[0] || '',
+                        lastName: submissionData?.healthcareProfessionalName?.split(' ')[1] || '',
```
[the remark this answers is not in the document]
NabbeunNabi: Right now both the first and last name are necessary for the submission. This is just to autofill from a submission the doctor. There is no middle name field on the submission form from a user either. It defaults to an empty string if undefined. I would think that would be ok for now. Also this is just for a user submitted doctor so no matter what, the moderator needs to do the work of making sure this name is correct or even exists. That is how I understood the intent? If we do have ones with middle names or last names only. This could be updated, but it still falls on the moderator to confirm.

components/ModEditSubmissionForm.vue
```diff
@@ -604,86 +609,216 @@ function autofillEditSubmissionForm(submissionData: Submission | undefined) {
...
 
-async function submitForm(e: Event) {
-    // stop the form submitting before we validate
-    e.preventDefault()
+//Helps add the localized name to the submission
+const handleLocalizedNameToSubmission = () => {
+    const localizedNameToAdd: LocalizedNameInput = {
+        firstName: localizedFirstName.value,
+        lastName: localizedLastName.value,
+        locale: nameLocale.value || Locale.EnUs,
```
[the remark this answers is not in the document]
NabbeunNabi: I have added a type on the backend now so need the following ticket approved to have that type on the frontend.
https://github.com/ourjapanlife/findadoc-server/pull/588

ermish:
@NabbeunNabi is actively splitting this into separate PRs so putting back in draft

components/ModEditSubmissionForm.vue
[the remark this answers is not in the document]
NabbeunNabi: MERGED AND UPDATED

components/ModEditSubmissionForm.vue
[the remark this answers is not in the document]
ermish: it's technically fine as is right now and we'd probably be fine when the data changes since our team is small and we understand our code. The risk increases when you can't track what needs to change when your data/API does (example: middle name would silently break this name split logic since it won't throw an error, but it will start showing in last name field). To prevent this, **the recommended practice is to always assume your data coming from the API could be changed/missing/different format**. 
If/when that happens, you want the UI to at least not crash completely. 
Usually there's a UX decision on showing partial successful rendering or friendly crashing the page and redirecting the user. 

An alternative approach is you do all of the proper data integrity checks when you receive data from the API instead of when it's used. 

I leave it up to you if you want to change it now, in a future ticket, or just not worry about it until we hit that circumstance 😸

components/ModEditSubmissionForm.vue
```diff
@@ -692,61 +745,133 @@ const submissionHasUnsavedChanges = () => {
...
+            ]
         }
     }
+    try {
+        const result = await moderationSubmissionStore.updateSubmission(submissionInputVariables)
+        const submissionResult = result as { updateSubmission: Submission }
+        initializeSubmissionFormValues(submissionResult.updateSubmission as Submission)
+        toast.success(t('modSubmissionForm.successMessageUpdated'))
+    } catch (error) {
+        console.error('Failed to update submission:', error)
```
ermish, line 820: should be a translation string

ermish: n/m, I see the toast to the user is translated 👍

components/ModEditSubmissionForm.vue
```diff
@@ -692,61 +745,133 @@ const submissionHasUnsavedChanges = () => {
...
+                    spokenLanguages: submissionFormFields.healthcareProfessionalLocales.value,
+                    names: submissionFormFields.healthCareProfessionalNameArray.value,
+                    facilityIds: []
+                }
+            ]
         }
     }
+    try {
+        const result = await moderationSubmissionStore.updateSubmission(submissionInputVariables)
+        const submissionResult = result as { updateSubmission: Submission }
```
ermish, line 816: we should check for errors returned from the server

components/ModEditSubmissionForm.vue
```diff
@@ -692,61 +745,133 @@ const submissionHasUnsavedChanges = () => {
...
+    }
+
+    if (!isValidHealthcareProfessional) {
+        toast.error(t('modSubmissionForm.errorMessageHealthcareInputsInvalid'))
+        return
+    }
 
-    moderationSubmissionStore.updateSubmission(submissionInputVariables)
+    try {
+        await moderationSubmissionStore.approveSubmission()
```
ermish, line 857: we should check the server response for errors

components/ModEditSubmissionForm.vue
```diff
@@ -692,61 +745,133 @@ const submissionHasUnsavedChanges = () => {
...
+                    specialties: submissionFormFields.healthcareProfessionalSpecialties.value,
+                    spokenLanguages: submissionFormFields.healthcareProfessionalLocales.value,
+                    names: submissionFormFields.healthCareProfessionalNameArray.value,
+                    facilityIds: []
+                }
+            ]
         }
     }
+    try {
+        const result = await moderationSubmissionStore.updateSubmission(submissionInputVariables)
```
ermish, line 815: When we call `updateSubmission`, I believe it will return the object with the valid `facilityIds` array and `healthcareProfessionalIds` array populated. We should update the object we're using here as I think it might try to wipe the relationships if you hit update again with an empty array.
While I'm writing this, I can see how this wouldn't have been an intuitive logic to pick up on....I'll have to think how we can improve this 🤔

components/ModEditSubmissionForm.vue
```diff
@@ -515,15 +519,22 @@ import { validateAddressLineEn,
...
+import { onBeforeRouteLeave } from '#app'
+import { useI18n } from '#imports'
 import { multiSelectWithoutKeyboard } from '~/utils/multiSelectWithoutKeyboard'
 import { useModalStore } from '~/stores/modalStore'
-import { onBeforeRouteLeave } from '#app'
+import SVGTrashCan from '~/assets/icons/trash-can.svg'
+import SVGProfileIcon from '~/assets/icons/profile-icon.svg'
+import { triggerFormValidationErrorMessages } from '~/utils/triggerFormValidationErrorMessages'
+
+let toast: ToastInterface
```
ermish, line 532: Is there a reason this isn't a `const` assigned to `useToast()`?

ermish (changes requested):
woo! made it through! excited to have this logic in soon!

components/ModEditSubmissionForm.vue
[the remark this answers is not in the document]
NabbeunNabi: I just want some way to display the name that was entered by the submission. But I really do not understand what you mean. If it shows the middle name instead of the last name. It is still showing something and no matter what the moderator has to check that the submission is a real one. I can check and validate and check the length and render based on that. Does that make sense @ermish.

assets/icons/trash-can.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="14" height="14" viewBox="0 0 14 14" fill="none" xmlns="http://www.w3.org/2000/svg">
```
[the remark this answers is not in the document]
NabbeunNabi: Why do these always exist but not show in the preview extension? Is it cuz of the fill color?

components/ModEditSubmissionForm.vue
NabbeunNabi: Am I not updating with the response in the next line with the `initialize` function?

components/ModEditSubmissionForm.vue
NabbeunNabi: This is the thing I have been talking about with documentation. I will add comments. But if you run it before onMounted it crashes the build cuz it is only a user component. You have to do it `onMounted` or everything will fail due to the way Nuxt builds with the SSR and setup. It won't be able to find `useToast`

assets/icons/trash-can.svg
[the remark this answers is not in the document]
NabbeunNabi: Also what is the name of this file? That exists?

components/ModEditSubmissionForm.vue
NabbeunNabi: is this not what the catch block is doing?

ermish: no, that's a failure in the http request and shouldn't catch validation/business logic errors from the server. those are sent via the response object. 
Why? As a practice, you want to avoid exceptions as your error handling for planned errors (like bad user input, validation issues, etc) as they are slow and heavyweight, and should be reserved for errors we can't code against

components/ModEditSubmissionForm.vue
ermish: just wrap it in a client side check and it should work fine?

assets/icons/trash-can.svg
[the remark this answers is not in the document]
NabbeunNabi: We talked about this

assets/icons/trash-can.svg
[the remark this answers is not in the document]
NabbeunNabi: We talked about this

components/ModEditSubmissionForm.vue
ermish: oof, I must have missed that call. That should work 👍 
Let's add a comment to explain that you're updating the form with the values from the api that has updated ids

components/ModEditSubmissionForm.vue
NabbeunNabi: We talked about this

components/ModEditSubmissionForm.vue
NabbeunNabi: We talked about this

components/ModEditSubmissionForm.vue
NabbeunNabi: It hasn't with Nuxt but I will try

components/ModEditSubmissionForm.vue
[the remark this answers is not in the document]
NabbeunNabi: I did it!

components/ModEditSubmissionForm.vue
NabbeunNabi: @ermish  This is the error I get if I try without doing it onMounted

image

ermish (approved):
Thanks for the contribution, Mr. Contributor.

ermish (approved):

PR #734, opened 2024-09-11
chore: change modalStore into Vue setup store

### LINKED ISSUE #733: REFACTOR: Use composition api in modalStore
# Problem
Since we use Vue3 we are using the composition API (read more about it here: https://vuejs.org/about/faq)
This means we want to use that approach as well in our Pinia stores (Pinia calls this Setup Store).

Currently `modalStore` is not using the Setup store.

# Success
Refactor the `modalStore` using the Setup store. You can use the other stores as an example. 
Here is the documentation of Pinia (setup store): https://pinia.vuejs.org/core-concepts/
### Requirements

- [ ] Refactor `modalStore` using the setup store approach

ShizuokaTerrier:
Resolves #733 

## 🔧 What changed 

We refactored the modalStore from option store to setup store using loadingStore as an example to follow.

### DIFF
--- a/stores/modalStore.ts
+++ b/stores/modalStore.ts
@@ -1,16 +1,16 @@
 import { defineStore } from 'pinia'
+import { ref, type Ref } from 'vue'
 
-export const useModalStore = defineStore({
-    id: 'modal-store',
-    state: () => ({
-        isOpen: false
-    }),
-    actions: {
-        showModal(): void {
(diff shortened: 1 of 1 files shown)


ShizuokaTerrier:
We checked to see whether the modals were working as expected in dev and had no problems. Can anyone shed some light on why we are failing the pinia test?

PR #739, opened 2024-09-17
feat: implement dashboard topbar

### LINKED ISSUE #735: Create topbar for mod content in the list component
# Problem
Right now we still have just an empty topnav with a placeholder. This needs to have a UI/UX that looks like that of the [FIGMA](https://www.figma.com/design/QXQxlw4jNFIhMxkKU7bbwq/Find-a-doc%2C-Japan?node-id=1865-19285&t=aJF73Sq2EtV5kYJ9-4)

# Success
It has the UI that matches that of the Figma
### Requirements

- [ ] UI matches the Figma

# Implementation notes
Do not worry about the functionality at the moment of the search function. As of 9/13/2024 we still need to discuss how we should handle search functionality.

NabbeunNabi:
Resolves #735 

## 🔧 What changed
Before we didn't even have a topbar UI/UX. This does not have any search functionality as of now. Since we need to decide how we want to do this in the future.

## 🧪 Testing instructions
Manually test it by making sure it updates as expected.

- run the backend with
1. `npm run dev:startlocaldb`
2. `npm run dev`

- go to the frontend and run
1. yarn dev:localserver
2.  go to the moderation panel and switch the dashboard views

## 📸 Screenshots

-   ### Before
image

-   ### After
image
image
image
image
image

### DIFF
--- a/components/ModDashboardTopbar.vue
+++ b/components/ModDashboardTopbar.vue
@@ -1,0 +1,60 @@
+<template>
+    <div class="flex flex-row items-center justify-between">
+        <h1
+            v-if="moderationSubmissionsStore.selectedModerationListViewChosen === SelectedModerationListView.Submissions"
+            class="font-semibold text-xl mx-1"
+        >
+            {{ updateTextForSubmissionDashboard(moderationSubmissionsStore.selectedModerationListViewTabChosen) }}
+        </h1>
+        <h1
+            v-else-if="moderationSubmissionsStore.selectedModerationListViewChosen === SelectedModerationListView.Facilities"
+            class="font-semibold text-xl mx-1"
+        >
+            {{ $t("modDashboardTopbar.facilities") }}
+        </h1>
+        <h1
+            v-else-if="moderationSubmissionsStore.selectedModerationListViewChosen
+                === SelectedModerationListView.HealthcareProfessionals"
+            class="font-semibold text-xl mx-1"
+        >
+            {{ $t("modDashboardTopbar.healthcareProfessionals") }}
+        </h1>
+        <div class="flex flex-col mt-4">
+            <div class="justify-start items-start flex">
+                <input
+                    type="text"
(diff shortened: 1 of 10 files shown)


components/ModDashboardTopbar.vue
```diff
@@ -0,0 +1,60 @@
+<template>
+    <div class="flex flex-row items-center justify-between">
```
ermish, line 2: Flex-row is redundant

components/ModDashboardTopbar.vue
```diff
@@ -0,0 +1,60 @@
...
+        <h1
```
ermish, line 3: Would it be simpler to just swap out the text based on a switch statement? Or do you think the styling might change based on the page? (Because that would make this approach better )

components/ModDashboardTopbar.vue
```diff
@@ -0,0 +1,60 @@
...
+            {{ $t("modDashboardTopbar.facilities") }}
+        </h1>
+        <h1
+            v-else-if="moderationSubmissionsStore.selectedModerationListViewChosen
+                === SelectedModerationListView.HealthcareProfessionals"
+            class="font-semibold text-xl mx-1"
+        >
+            {{ $t("modDashboardTopbar.healthcareProfessionals") }}
+        </h1>
+        <div class="flex flex-col mt-4">
```
ermish, line 22: Add an id or class to help identify what these div sections are doing

ermish (changes requested):
Nice! Just a few changes

components/ModDashboardTopbar.vue
NabbeunNabi: I thought of both ways. I have part of the switch statement for swapping out text.  I decided to put the other two in for the reasons you said. But I feel like for now. If there is a reason to change the styling later maybe a switch statement would be better for now.

ermish (approved):
topbar activate!

PR #751, opened 2024-09-30
feat: create back button submission form modal 

### LINKED ISSUE #700: FEAT: Create a go back button for form -> overview submissions
# Problem
When a user clicks on a submission and arrives at the form, there is no easy way to return to the submissions overview if no changes are made.

# Success
Provide a button that allows users to navigate back from the form to the submissions overview page.

### Requirements

- [ ] Create a "Go Back" button that takes the user from the form back to the submissions overview page.

ShizuokaTerrier:
Resolves #700 

## 🔧 What changed

A modal was added that only triggers on use of the back button in the browser.

## 🧪 Testing instructions

No tests were added. 

## 📸 Screenshots

-   ### Before

It didn't exist before. 

-   ### After

https://github.com/user-attachments/assets/62b0dcf1-436b-45ec-a8b5-77c3f0afb8f5

### DIFF
--- a/components/ModEditSubmissionForm.vue
+++ b/components/ModEditSubmissionForm.vue
@@ -1,4 +1,23 @@
 <template>
+    <div
+        v-show="modalStore.isOpen"
+        class="fixed top-0 left-0 flex items-center justify-center h-full w-full z-10 bg-secondary bg-opacity-40"
+    >
+        <Modal
+
+            class=" min-h-20 min-w-20"
+        >
+            <div class="flex flex-col aspect-square h-96 items-center justify-around bg-primary-inverted p-10 rounded">
+                <span class="font-bold text-3xl">{{ $t('modSubmissionForm.confirmationModal') }}</span>
+                <button
+                    class="bg-secondary p-4 rounded-full my-8 font-semibold text-xl hover:bg-primary"
+                    @click="handleConfirmationOfBack"
+                >
+                    {{ $t('modSubmissionForm.confirmationButton') }}
+                </button>
+            </div>
+        </Modal>
+    </div>
     <form
         class="p-4 h-full overflow-y-auto"
         @submit="submitForm"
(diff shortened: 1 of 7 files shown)


ShizuokaTerrier:
Got a lot of support from @ProgressChris88 in the initial stage and an enormous amount of help in solving the mystery of the maddeningly inconsistent modal from @NabbeunNabi. Big thanks guys.

theyokohamalife:
Great work getting the modal working! First thought that comes to mind is that we probably only want it to appear if there are _actual unsaved changes_. I noticed in the demo video that the modal opened if a user navigated to the edit page and hit the back button without making any changes.

components/ModEditSubmissionForm.vue
```diff
@@ -1,4 +1,24 @@
 <template>
+    <Modal
+        v-if="modalStore.isOpen"
+        class=" min-h-20 min-w-20 border-8"
+    >
+        <p>Are you sure you want to leave? You have unsaved changes!</p>
```
theyokohamalife: It's a good thing to avoid adding raw strings, because we may forget to localize them later. Let's add this line to the localization files!

components/ModEditSubmissionForm.vue
```diff
@@ -1,4 +1,24 @@
...
+        <div class="flex">
+            <button
+                class="flex justify-center items-center rounded-full bg-primary w-28 text-primary-inverted text-sm mr-2 "
+                @click="screenStore.activeScreen = ModerationScreen.Dashboard"
+            >
+                Leave
```
theyokohamalife: ♻️

components/ModEditSubmissionForm.vue
```diff
@@ -1,4 +1,24 @@
...
+            </button>
+            <button
+                class="flex justify-center items-center rounded-full bg-primary w-28 text-primary-inverted text-sm mr-2 "
+                @click="modalStore.hideModal"
+            >
+                Stay
```
theyokohamalife: ♻️

components/ModEditSubmissionForm.vue
```diff
@@ -712,6 +737,16 @@ onMounted(() => {
     )
 })
 
+// This doesn't work consistently as sometimes the to.path is not '/moderation', it also blocks our save & exit button and triggers when making this component visable. Also doesn't work at all when imported from vue-router.
```
theyokohamalife: Question: Why do we care about the to/from path here? Is it necessary for the modal to work? Let's say the user accidentally clicks a button that will take them to a different route, shouldn't we still display the modal?

components/ModEditSubmissionForm.vue
```diff
@@ -1,4 +1,23 @@
...
+    <div
+        v-show="modalStore.isOpen"
+        class="fixed top-0 left-0 flex items-center justify-center h-full w-full z-10 bg-secondary bg-opacity-40"
+    >
+        <Modal
+
+            class=" min-h-20 min-w-20"
+        >
+            <div class="flex flex-col aspect-square h-96 items-center justify-around bg-primary-inverted p-10 rounded">
+                <span class="font-bold text-3xl">{{ $t('modSubmissionForm.confirmationModal') }}</span>
```
theyokohamalife, line 11: Nice! I would add these changes to the first commit and skip checking in the raw text changes.

theyokohamalife:
Nit: As a good practice, commit message titles should be limited to about 50 characters. Anything longer than that could be added to the message body. You may want to think about how you could write this message more concisely:
Screenshot 2024-09-30 at 5 44 06 PM

components/ModEditSubmissionForm.vue
ShizuokaTerrier: Sorry for the confusion. If you take a look at the most recent commit you will find the raw strings aren't there and we have updated the en locale file, too.

components/ModEditSubmissionForm.vue
ShizuokaTerrier: At the time of writing that specific code, I cared about the path because it was the only way to get the modal to function as intended. The ticket was to create a modal that pops up when the back button in the browser is clicked, so that's what I focused on. I do agree that focusing on a more broad approach might have been more logical.

theyokohamalife (approved):
LGTM ✨  Please remember to squash your fixup commit before merging and also write a ticket for what needs to be done next like we discussed on Slack.

theyokohamalife:
@all-contributors
please add @ShizuokaTerrier for coding.
please add @ProgressChris88 for coding.

theyokohamalife:
@all-contributors please add @ProgressChris88 for coding

theyokohamalife:
@all-contributors please add @ShizuokaTerrier for coding

PR #759, opened 2024-10-02
fix: improve variable names in submission store

NabbeunNabi:
## 🔧 What changed
Before the variable names in the store were reading `mutation` and `query`. It wasn't desirable to name a ref the same as the actual API calls, so it was updated to be a name that fit with what the variable is doing and then updated where the variables were being previously used.


ermish (approved):
So easy to review!

PR #761, opened 2024-10-02
refactor: remove required validation check from mod input

### LINKED ISSUE #760: Update the Mod Input component to not have required validation
# Problem
Right now the validation is required for this reusable component. But not all fields in a form may be required due to the schema in the backend. Most of the form has required text. But this can make this component reusable even more when it comes to possibly having notes in the future. Or for #718, middle name is not required

# Success
The validation input is not required and the validation check still works

### Requirements

- [x] refactor validation input to not be required
- [x] refactor validation check based on required prop

NabbeunNabi:
Resolves #760 

## 🔧 What changed
Before the validation check was required. But not every field on the form will always be required. For instance, in Japan middle names are rare but are optional when accepting a doctor or submission. This reworks the logic for the validation check in order to fit the reusability within the moderation panel.

### DIFF
--- a/components/ModInputField.vue
+++ b/components/ModInputField.vue
@@ -14,7 +14,7 @@
             :type="type"
             :placeholder="placeholder"
             :required="required"
-            class="mb-5 px-3 py-3.5 w-96 h-12 bg-secondary-bg rounded-lg border border-primary-text-muted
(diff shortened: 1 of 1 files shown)


ermish (approved):
Ship it

PR #765, opened 2024-10-02
refactor: change toast from build to module in config

NabbeunNabi:
## 🔧 What changed
This refactors the toast to be a module in the nuxt config instead of in the build for consistency.


ermish (approved):
this is better. More consistent with the other modules

PR #767, opened 2024-10-03
chore: update gqlTypes.ts based on updated backend schema

NabbeunNabi:
## 🔧 What changed
Types were updated in the backend schema. But at the moment we do not have a script that checks if the schema has changed. And if it does to update the types on the frontend. This will be discussed and added as a possible future ticket.
This updates our `gqlTypes` file based on the new schema changes.


NabbeunNabi:
@ermish @theyokohamalife Do you think we should make a script that checks weekly or monthly for `gqlSchema` by running generate and if it does have file changes commits those changes?

theyokohamalife (approved):
LGTM

theyokohamalife:
> @ermish @theyokohamalife Do you think we should make a script that checks weekly or monthly for `gqlSchema` by running generate and if it does have file changes commits those changes?

That would be really nice to have and would help maintain the types!

PR #769, opened 2024-10-03
feat: implement modal for submission acceptance

NabbeunNabi:
## 🔧 What changed
This is part of breaking out #718  into smaller PRs. This implements the Modal showing from the top bar when accept is clicked, so that you can confirm that you actually want to submit the facility into the database and you are sure about the changes you made. This keeps the user from accidentally submitting when on the page with a mis click.

### After
image

### DIFF
--- a/components/ModEditSubmissionForm.vue
+++ b/components/ModEditSubmissionForm.vue
@@ -7,7 +7,25 @@
 
             class=" min-h-20 min-w-20"
         >
-            <div class="flex flex-col aspect-square h-96 items-center justify-around bg-primary-inverted p-10 rounded">
+            <div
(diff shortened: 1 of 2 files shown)


theyokohamalife:
The code looks good! Since this is a UI change, it's a good idea to include a before and after screenshot.

NabbeunNabi:
> The code looks good! Since this is a UI change, it's a good idea to include a screenshot of the modal in action ✨

@theyokohamalife screenshot added!

theyokohamalife (approved):
LGTM

ermish:
I think I commented this before, but why is this a separate modal from our standard modal? We could skip this PR if we were using the shared one

PR #772, opened 2024-10-03
feat: add workflow for validating pr titles

theyokohamalife:
## 🔧 What changed

We'd like to ensure that commit messages that go into main are consistent and high quality. Since we use the "squash and merge" setting, the PR title is what becomes the commit message when it is merged. This workflow will check that the PR title follows [conventional commit specifications](https://www.conventionalcommits.org/en/v1.0.0/).

## Resources
[Workflow documentation](https://github.com/amannn/action-semantic-pull-request)
[Slack discussion](https://find-a-doc.slack.com/archives/C02663QK8TD/p1727925045163499)

### DIFF
--- a/.github/workflows/lint-pr-title.yml
+++ b/.github/workflows/lint-pr-title.yml
@@ -1,0 +1,21 @@
+name: "Lint PR title"
+
+on:
+  pull_request_target:
+    types:
+      - opened
+      - edited
+      - synchronize
+      - reopened
+
+permissions:
+  pull-requests: read
+
+jobs:
+  main:
+    name: Validate PR title
+    runs-on: ubuntu-latest
+    steps:
+      - uses: amannn/action-semantic-pull-request@v5
+        env:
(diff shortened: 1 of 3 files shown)


.github/workflows/lint-pr-title.yml
```diff
@@ -0,0 +1,21 @@
+name: "Lint PR title"
+
+on:
+  pull_request_target:
```
theyokohamalife, line 4: By using `pull_request_target` instead of `pull_request` it'll allow the workflow to run on forked PRs as well. As a downside, it will only run the action once it's on main (meaning not for this PR).

NabbeunNabi: @theyokohamalife You can trick github into registering an action. But in this case I do not think that is necessary. It loos good to me.

NabbeunNabi (approved):
💯

ermish (approved):
Ship it!

ermish:
Before you merge, I would update the documentation @theyokohamalife so people understand why their PR is failing.

PR #778, opened 2024-10-06
fix: use Ja for Japanese locale

theyokohamalife:
We are incorrectly using `Jp` instead of `Ja` in many variable names where the value is localized in Japanese. `Ja` is the locale and `Jp` is the country.


NabbeunNabi (approved):

NabbeunNabi:
This is 100 percent correct. Ja is the standard used for localization of Japanese language

ermish (approved):
Clear as Japanese!

PR #780, opened 2024-10-07
fix: register toast plugin on client

### LINKED ISSUE #781: Fix toast plugin import
# Problem
We tried many ways to import the toast from the library. But due to it being a client component the deployment would fail on branches using the toast notifications.

# Success
The plugin is imported, used, and tested by trying to deploy with a `toast` notification. After successful deployment, the test `toast` should be removed.

### Requirements

- [x] Correct plugin import (none due to Nuxt documentation)
- [x] Used and tested for a successful deployment
- [x] Toast removed

NabbeunNabi:
Resolves #781 

## 🔧 What changed
Before we were incorrectly importing and using the toast library. Based on the Nuxt documentation to import a client plugin we just need to name the file with `filename.client.ts`. It is then auto-imported. But due to the nature of our project with SSR and Netlifiy's building in deployment, if we just used the import from the library directly `cjs` in the library would cause the build to fail. If we set a variable to `let toast: ToastInterface` and then `onMounted` set `toast` to `useToast` if builds correctly as everything is done on the client side. There is a ticket made to update the documentation so other developers will understand how to use this library in #782 

## 🧪 Testing instructions
1. Create a branch with this as a base branch
2. On the `about.vue` page import toast and use it on mounted based on the necessary steps above.
3. Create a draft PR to see the build work and use the preview to see your toast message that you put when the page is mounted
4. Delete PR and branch

## 📸 Screenshots
N/A due to it being a fix of an improper plugin and module import


theyokohamalife (approved):
LGTM

ermish (approved):
Nice!

PR #794, opened 2024-10-08
feat: implement weekly gql types update

### LINKED ISSUE #791: Write script to auto update gqlTypes if backend schema changes
# Problem
The way this project works, we generate the types on the frontend based on our backend schema. We now have to manually run and make a PR whenever there are changes to generate those types on the frontend. This could lead to us forgetting to update the types on the frontend. 

# Success
A script is written that runs weekly to generate the types on the front end. And if there are new types in the file, commits these changes and makes a PR

### Requirements

- [x] Write a script that runs weekly to auto update types
- [x] If there are changes this script will create a PR

NabbeunNabi:
Resolves #791 

## 🔧 What changed
Before we had to manually update the types for `gql` on the frontend whenever there was a change on the backend. This could cause us to miss types or not have an updated file for months. This workflow will run weekly at Monday at midnight UTC. It will first generate the types and see if there is a difference from the main branch. If there is it will open a PR against main with a conventional commit message from its own branch.


.github/workflows/generate-gql-types.yml
```diff
@@ -0,0 +1,58 @@
...
+        with:
+          node-version: ${{ matrix.node-version }}
+          cache: 'yarn'
+      - name: Install Dependencies
+        run: yarn install --immutable
+      - name: Generate GQL Types
+        run: yarn generate
+      - name: Check for changes
+        run: git diff --exit-code -- typedefs/
+        continue-on-error: true
```
theyokohamalife, line 37: Why do we want to continue on error here? 🤔

ermish (approved):
This is great to have! My only concern would be that it looks to run when there's a PR and will commit into that PR which could cause confusion and changes that the Author didn't write. This would be better to run as an automated job (which I'm not sure if GitHub supports yet) similar to how our dependabot runs on a schedule independent of our PRs.

NabbeunNabi:
Does it not open its own branch @ermish? Maybe I misunderstood how it works?

ermish:
> Does it not open its own branch @ermish? Maybe I misunderstood how it works?

it does. I'm referring to when it's triggered

.github/workflows/generate-gql-types.yml
NabbeunNabi: @theyokohamalife Based on my understanding of how these work. The `git diff --exit-code` would return a non zero code and therefore return an error. Letting it not proceed even though we want the rest of the workflow to run in opening a branch and committing the changes.

.github/workflows/generate-gql-types.yml
```diff
@@ -0,0 +1,58 @@
+name: Generate and Commit GQL Types
+
+on:
+  schedule:
+    # Run the workflow every Monday at midnight (UTC)
+    - cron: '0 0 * * MON'
```
NabbeunNabi, line 6: @ermish It isn't triggered on a PR. That usually looks like this I though in `yml` based on our other workflows as follows

```
on:
  schedule:
    - cron: '0 0 * * MON'
  pull_request:
    branches:
      - main
```

This should only run as scheduled once a week and act as a bot similar to our renovate bot.

ermish: Ohhh hmm I was thinking the cron job scheduler was triggered by a PR. Well it looks good then if it auto runs independently!

Issue #771, opened 2024-10-03
Update mod panel modal logic


ShizuokaTerrier:
# Problem
The modal currently opens when a user tries to navigate away from the moderation submission edit page using the back button in the browser. Logically we should only do this if there is unsaved information written in the form that may be lost. 

# Success
The modal only appears when a user has entered information in the form and tries to leave the page without saving it. 

### Requirements

- [x] Refactor the modal to only appear when input is not saved. 

# Implementation notes
Only the logic for the second **if** block condition needs changing the rest of the statement should not need refactoring.

NabbeunNabi:
Closed as shown as the merge from #800

PR #800, opened 2024-10-09
refactor: add check for unsaved changes to modEditSubmissionForm

theyokohamalife:
Related issue: #771

## 🔧 What changed
This PR refactors the `modEditSubmissionForm` component so that the modal only shows if a user hits the back button and has unsaved changes.

## 🧪 Testing instructions
In your local environment: 
1. Make a change to a submission and hit the back button. You should see the modal.
2. Make a change to a submission and hit the "Save and exit" button. You shouldn't see the modal.
3. Click a submission and hit the back button without making a change. You shouldn't see the modal.

See the steps in the README for how to log into the moderation panel.

## 📸 Screenshots
-   ### Before
Screenshot 2024-10-10 at 1 30 53 AM

-   ### After
Screenshot 2024-10-10 at 1 29 16 AM

### DIFF
--- a/components/ModEditSubmissionForm.vue
+++ b/components/ModEditSubmissionForm.vue
@@ -1,41 +1,40 @@
 <template>
-    <div
+    <Modal
         v-show="modalStore.isOpen"
-        class="fixed top-0 left-0 flex items-center justify-center h-full w-full z-10 bg-secondary bg-opacity-40"
+        data-testid="submission-form-modal"
+        class=" min-h-20 min-w-20 fixed top-0 left-0 flex items-center justify-center h-full w-full z-10
+         bg-secondary bg-opacity-40"
     >
-        <Modal
-
-            class=" min-h-20 min-w-20"
+        <div
+            v-if="moderationSubmissionStore.approvingSubmissionFromTopBar"
+            class="flex flex-col aspect-square h-96 items-center justify-around bg-primary-inverted p-10 rounded"
         >
-            <div
-                v-if="moderationSubmissionStore.approvingSubmissionFromTopBar"
-                class="flex flex-col aspect-square h-96 items-center justify-around bg-primary-inverted p-10 rounded"
+            <span class="font-bold text-3xl">
+                {{ $t('modSubmissionForm.submissionConfirmationMessage') }}
+            </span>
+            <button
+                type="button"
+                class="bg-primary p-4 rounded-full my-8 font-semibold text-xl"
+                @click="submitForm"
             >
-                <span class="font-bold text-3xl">
-                    {{ $t('modSubmissionForm.submissionConfirmationMessage') }}
-                </span>
-                <button
-                    type="button"
-                    class="bg-primary p-4 rounded-full my-8 font-semibold text-xl"
-                    @click="submitForm"
-                >
-                    {{ $t('modSubmissionForm.submissionConfirmationAcceptanceButton') }}
-                </button>
-            </div>
-            <div
-                v-else
-                class="flex flex-col aspect-square h-96 items-center justify-around bg-primary-inverted p-10 rounded"
+                {{ $t('modSubmissionForm.submissionConfirmationAcceptanceButton') }}
+            </button>
+        </div>
+        <div
+            v-else
+            class="flex flex-col aspect-square h-96 items-center justify-around bg-primary-inverted p-10 rounded"
+        >
+            <span class="font-bold text-3xl">{{ $t('modSubmissionForm.hasUnsavedChanges') }}</span>
+            <button
+                data-testid="submission-form-modal-confirmation-btn"
+                class="bg-secondary p-4 rounded-full my-8 font-semibold text-xl hover:bg-primary"
+                @click="handleNavigateToModerationScreen"
             >
-                <span class="font-bold text-3xl">{{ $t('modSubmissionForm.confirmationModal') }}</span>
-                <button
-                    class="bg-secondary p-4 rounded-full my-8 font-semibold text-xl hover:bg-primary"
-                    @click="handleConfirmationOfBack"
-                >
-                    {{ $t('modSubmissionForm.confirmationButton') }}
-                </button>
-            </div>
-        </Modal>
-    </div>
+                {{ $t('modSubmissionForm.confirmationButton') }}
+            </button>
+        </div>
+    </Modal>
+
     <form
         class="p-4 h-full overflow-y-auto"
         @submit="submitForm"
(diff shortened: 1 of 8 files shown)


components/ModEditSubmissionForm.vue
```diff
@@ -513,36 +513,40 @@ const router = useRouter()
...
+    localizedLastName: ref('') as Ref<string>,
+    localizedMiddleName: ref('') as Ref<string>,
+    nameLocale: ref(Locale.EnUs) as Ref<Locale>,
+    healthcareProfessionalAcceptedInsurances: ref([]) as Ref<Array<Insurance>>,
+    healthcareProfessionalDegrees: ref([]) as Ref<Array<Degree>>,
+    healthcareProfessionalSpecialties: ref([]) as Ref<Array<Specialty>>,
+    healthcareProfessionalLocales: ref([]) as Ref<Array<Locale>>
+} as { [key: string]: Ref }
+
+const submissionBeforeChanges: { [key: string]: (string | []) } = {}
```
theyokohamalife, line 549: These fields were moved into an object to make it easier to determine if changes were made to the form.

components/ModEditSubmissionForm.vue
```diff
@@ -620,34 +624,46 @@ const validateFields = () => {
...
-                    googlemapsURL.value = submissionData['facility']?.contact?.googleMapsUrl || submissionData['googleMapsUrl']
+                    submissionFormFields.googlemapsURL.value = submissionData['facility']?.contact?.googleMapsUrl || submissionData['googleMapsUrl']
                     break
             }
         }
     }
+    for (const key in submissionFormFields) {
+        submissionBeforeChanges[key] = submissionFormFields[key].value
+    }
+}
```
theyokohamalife, line 658: This function was updated to create a snapshot of the form before changes are made in `submissionBeforeChanges`.

components/ModEditSubmissionForm.vue
```diff
@@ -723,54 +739,43 @@ watch(moderationSubmissionStore, newValue => {
...
         extractLocaleOptions
     )
 })
 
-/* This handler function confirms that the user wants to return to the mod page, updates the modal state, updates the active
-screen and the visible url and router path so that the onBeforeRouteLeave function works as intended.
-If you don't update the route using push, the logic for the leave handler breaks.
-*/
-const handleConfirmationOfBack = () => {
+const handleNavigateToModerationScreen = () => {
```
theyokohamalife, line 771: I tried to get this function to route the user to whatever their previous route was, regardless of it was `/moderation` or not. However, I found that other routes skipped the `onBeforeRouteLeave` logic entirely.

components/ModEditSubmissionForm.vue
```diff
@@ -1,41 +1,38 @@
 <template>
-    <div
```
theyokohamalife, line 2: Removed this div and moved the styling and v-show to the modal.

components/ModEditSubmissionForm.vue
```diff
@@ -1,41 +1,38 @@
...
+    <Modal
```
NabbeunNabi, line 2: @theyokohamalife In this case I think we should go further and instead of in this component only. The logic should be moved into the `Modal.vue` component since the logic is consistent always for showing it in any future component across the whole map. Or the `ModMainContent` in the same way it is used in the `MainContentContainer.vue` component as follows

```
  <Modal
                v-show="store.$state.isOpen"
                class="max-h-[calc(100vh-12rem)] ml-8 mt-12 overflow-y-auto"
            >
                <SearchResultDetails />
            </Modal>
```

components/ModEditSubmissionForm.vue
NabbeunNabi: 💯 awesome!

components/ModEditSubmissionForm.vue
NabbeunNabi: If we used the `popState` and checked the `history` from the event, would we not be able to use a conditional to check if it is `moderation` and if it is to `router.push('/moderation')` and then `return` out otherwise to go back to the location pulled from the `popState` event?

NabbeunNabi (commented):
@theyokohamalife I asked some follow up questions. But it looks amazing and something I will need to refactor into my usage of my Big PR that is getting smaller and more manageable

components/ModEditSubmissionForm.vue
```diff
@@ -620,34 +621,46 @@ const validateFields = () => {
     )
 }
 
-function autofillEditSubmissionForm(submissionData: Submission | undefined) {
+function initializeSubmissionFormValues(submissionData: Submission | undefined) {
```
theyokohamalife, line 624: This function was updated to create a snapshot of the form before changes are made in submissionBeforeChanges.

components/ModEditSubmissionForm.vue
theyokohamalife: Sounds good! I can make the changes in a follow-up PR since it's a bit out of scope for this one.

components/ModEditSubmissionForm.vue
theyokohamalife: The issue isn't figuring out what the previous route is. We can get that from the `from` in `onBeforeRouteLeave`. What I observed is that none of the code will execute in this component if a users back to something other than `/moderation` like back to another submission they were viewing. It relaunches the auth, and none of the component logic is called.

components/ModEditSubmissionForm.vue
NabbeunNabi: Sounds good. Would you like me to write another ticket for that? Or you got that one? @theyokohamalife

ShizuokaTerrier (approved):
Learned a lot from this. Thanks!

components/ModEditSubmissionForm.vue
theyokohamalife: I'll write up a ticket for it. 😎

cypress/e2e/moderationDashboard.cy.ts
```diff
@@ -188,7 +188,7 @@ describe(
             })
 
             it('shows mod dashboard left navbar buttons with correct counts and functionality', () => {
-                //The number for include text is for the status in the mock data
+                // The number for include.text is for the status in the mock data.
```
ermish, line 191: Were these periods manually added? I'm not sure include.text is correct

i18n/locales/cn.json
```diff
@@ -228,7 +228,7 @@
     "selectDegrees": "Degrees",
     "selectLocales": "Select Locales Spoken",
     "confirmationButton": "Leave Page",
-    "confirmationModal": "Are you sure you want to leave?"
+    "hasUnsavedChanges": "You have unsaved changes. Are you sure you want to leave?"
```
ermish, line 231: Nice name change 👍

ermish (approved):
Approved, but you might want to fix that first comment change

ermish:
Looks like the tests are failing, can you take a look? @theyokohamalife

cypress/e2e/moderationDashboard.cy.ts
theyokohamalife: That period was intentional, because it's reading from `include.text`. I can remove it if it's confusing though.

Issue #801, opened 2024-10-10
Implement `updateSubmission` mutation into the `modSubmissionStore`


NabbeunNabi:
# Problem
We want all our API calls for data management to be stored in their respective stores. Currently this is being done within the file itself. This should be moved into a store.
# Success
The functionality of the `updateSubmission` is moved into the `modSubmissionStore`
### Requirements

- [x] Move the API call for the `graphQL` mutation into the `modSubmissionStore`
- [x] Use this refactored mutation call within the form itself.

PR #813, opened 2024-10-13
refactor: move update submission API call to store

theyokohamalife:
Resolves #[Issue number]
- [ ] PR title follows [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/#summary) specs

## 📸 Screenshots

-   ### Before

-   ### After

### DIFF
--- a/components/ModEditSubmissionForm.vue
+++ b/components/ModEditSubmissionForm.vue
@@ -498,8 +498,6 @@
 <script lang="ts" setup>
 import { onMounted, type Ref, ref, watch } from 'vue'
 import { useRouter } from 'vue-router'
-import { gql } from 'graphql-request'
-import { gqlClient, graphQLClientRequestWithRetry } from '~/utils/graphql.js'
 import { useModerationSubmissionsStore } from '~/stores/moderationSubmissionsStore'
 import { Locale,
     type Submission,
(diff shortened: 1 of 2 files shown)


NabbeunNabi (approved):
💯 Looks like a perfect refactoring of moving it into the store. Could use a test @theyokohamalife in the `vitest` tests of the stores either in this PR or a follow up one

ermish (approved):
nice improvement! 
- I would +1 adding a pinia test now that it's properly in the store, this is a good method to have a test on. 
- your description is missing the issue #. I would add one or just remove that line

NabbeunNabi:
> nice improvement!
> 
> * I would +1 adding a pinia test now that it's properly in the store, this is a good method to have a test on.
> * your description is missing the issue #. I would add one or just remove that line

I think I made an issue for this. Might have assigned myself @ermish @theyokohamalife

NabbeunNabi:
> > nice improvement!
> > 
> > * I would +1 adding a pinia test now that it's properly in the store, this is a good method to have a test on.
> > * your description is missing the issue #. I would add one or just remove that line
> 
> I think I made an issue for this. Might have assigned myself @ermish @theyokohamalife

I had. It has a requirement for a test. But that can be done in a future PR if you would like. But I assigned it to you @theyokohamalife . It is #801

NabbeunNabi:
@ermish @theyokohamalife @Anissa3005. The testing of this due to it calling the API requires a lot of setup with a `mock server request` and change of the `vitest.config.ts` to be changed to use `nuxt` instead of the way it is set up now so it can correctly resolve imports. @gminetoma And I looked and read up on this a lot today. So maybe the tests for this and #819 should be completed after tickets to reconfigure the `vitest.config` are made and PR is merged as well as a mock server is setup properly

PR #819, opened 2024-10-15
feat: implement approve submission mutation in store

### LINKED ISSUE #802: Implement `graphQL` mutation in the `modSubmissionStore` to approve a facility
# Problem
Right now there is no mutation and function written for approving a facility. We need this implemented within the store for use in the `ModEditSubmissionForm` component and testing within the standards of this project, where we test stores and functions within `vitest`

# Success
A graphQL mutation is made for approving a facility within the `modSubmissionStore` This mutation is tested using `vitest`

### Requirements

- [ ] Implement `graphQL` mutation for approving a submission
- [ ] Appropriate tests are written for this functionality within `vitest`

NabbeunNabi:
Resolves #802

## 🔧 What changed
Before there was not a specific function that was to just approve a facility. This has now been implemented in the store so it can be tested as well as used in the `ModEditSubmissionForm` in #718.

## 🧪 Testing instructions
TBA > Before tests can be done we need to update our `vitest.config.ts` to be integrated with Nuxt to allow imports from the `nuxtApp` itself

This is being done in #821

### DIFF
--- a/stores/moderationSubmissionsStore.ts
+++ b/stores/moderationSubmissionsStore.ts
@@ -101,6 +101,21 @@ export const useModerationSubmissionsStore = defineStore(
             }
         }
 
+        async function approveSubmission() {
+            if (!selectedSubmissionId.value) {
+                console.error('Submission Id is required')
+                return
+            }
+
+            const approveSubmissionInput: MutationUpdateSubmissionArgs = {
+                id: selectedSubmissionId.value,
+                input: {
+                    isApproved: true
+                }
+            }
+            updateSubmission(approveSubmissionInput)
+        }
+
         return { getSubmissions,
             submissionsData,
             filterSubmissionByStatus,
@@ -120,7 +135,8 @@ export const useModerationSubmissionsStore = defineStore(
             setApprovingSubmissionFromTopBar,
             selectedModerationListViewTabChosen,
             setSelectedModerationListViewTabChosen,
-            updateSubmission }
+            updateSubmission,
+            approveSubmission }
     }
 )
 


stores/moderationSubmissionsStore.ts
```diff
@@ -101,6 +101,21 @@ export const useModerationSubmissionsStore = defineStore(
             }
         }
 
+        async function approveSubmission() {
+            if (!selectedSubmissionId.value) {
```
theyokohamalife, line 105: Unblocking: I think we should probably log an error if someone tries to approve a submission without a valid ID.

NabbeunNabi: I'll add that as well as toasts.

I can do that in this PR @theyokohamalife or a future one.

I wanted to add the toast notifications to yours as well.

Maybe after this one I can have one for these that adds the error messaging?

🤔

theyokohamalife (approved):
Nit: I think a more accurate PR title would be something like "implement approve submission mutation in store"

Other than that it LGTM, but I think it would be a good idea to wait for @ermish to also have a look.

stores/moderationSubmissionsStore.ts
NabbeunNabi: @theyokohamalife Once this is merged I will make a ticket and add all the toast errors that show when incorrect actions are taken or there is failure.

theyokohamalife: At the very least we can `console.error` here with the ID.

stores/moderationSubmissionsStore.ts
```diff
@@ -101,6 +101,21 @@ export const useModerationSubmissionsStore = defineStore(
...
+                return
+            }
+
+            const approveSubmissionInput: MutationUpdateSubmissionArgs = {
+                id: selectedSubmissionId.value,
+                input: {
+                    isApproved: true,
+                    isUnderReview: false
```
ermish, line 113: I don't think you actually need to set this. It will be updated on backend

ermish (changes requested):
+1 to @theyokohamalife 's comment and one quite change and it's in!

stores/moderationSubmissionsStore.ts
```diff
@@ -101,6 +101,21 @@ export const useModerationSubmissionsStore = defineStore(
             }
         }
 
+        async function approveSubmission() {
+            if (!selectedSubmissionId.value) {
+                console.error('Submission Id is required')
```
NabbeunNabi, line 106: This is hardcoded because using translatable strings within stores in Pinia from our app causes the build to fail as it is expected in the setup of a Vue component

ermish: Hmmm sounds like it's running during compilation time which it shouldn't be 🤔. It's fine for now but we should look into this for pinia stores running at the wrong time

ermish (approved):
Ship it!

PR #822, opened 2024-10-16
test: implement nuxt as vitest environment

### LINKED ISSUE #821: Update vitest config to use Nuxt
# Problem
When vitest runs it does not have access to our imports and other parts of our app compiled by `Nuxt`

# Success
We follow the [testing documentation](https://nuxt.com/docs/getting-started/testing) Nuxt has to implement the config to run on the Nuxt app.
### Requirements

- [ ] Vitest config uses nuxt
- [ ] Current tests still pass
- [ ] Any packages regarding vitest that need to be upgraded are done so

NabbeunNabi:
Resolves #821 

## 🔧 What changed
The `vitest.config.ts` was updated to use our nuxtApp as the base since our app runs through `Nuxt`. This will give our testing files access to our `#imports` and `#app` directories.
## 🧪 Testing instructions
The tests are still passing

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,13 +46,14 @@
         "@graphql-codegen/typescript-resolvers": "^4.2.1",
         "@microsoft/eslint-formatter-sarif": "^3.1.0",
         "@nuxt/eslint": "^0.5.0",
+        "@nuxt/test-utils": "^3.14.3",
         "@nuxt/types": "^2.18.1",
         "@nuxtjs/i18n": "^8.3.3",
         "@stylistic/eslint-plugin": "^2.6.1",
         "@swc/core": "^1.7.6",
         "@testing-library/vue": "^8.1.0",
         "@vitejs/plugin-vue": "^5.1.2",
-        "@vitest/coverage-v8": "^2.0.5",
+        "@vitest/coverage-v8": "^2.1.3",
         "@vue/test-utils": "^2.4.6",
         "all-contributors-cli": "^6.26.1",
         "autoprefixer": "^10.4.20",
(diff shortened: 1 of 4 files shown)


vitest.config.ts
```diff
@@ -1,19 +1,8 @@
-import path from 'path'
-import { defineConfig } from 'vitest/config'
-import Vue from '@vitejs/plugin-vue'
+import { defineVitestConfig } from '@nuxt/test-utils/config'
 
-export default defineConfig({
-    plugins: [Vue()],
+export default defineVitestConfig({
```
ermish, line 3: I don't like how this completely replaces the whole native vitest config with their own nuxt version :( 
Is there any way to use nuxt within the original config style?

package.json
```diff
@@ -62,6 +63,7 @@
         "eslint-plugin-cypress": "^3.4.0",
         "eslint-plugin-json": "^4.0.0",
         "eslint-plugin-vue": "^9.27.0",
+        "happy-dom": "^15.7.4",
```
ermish, line 66: what's this for?

theyokohamalife: I did a quick look at [the Nuxt docs](https://nuxt.com/docs/getting-started/testing) and found:

> you can choose between happy-dom and jsdom for a runtime Nuxt environment

NabbeunNabi: I found the option to use jsdom. I followed the error messages which was referring to `happy-dom` even though I saw in the docs that it can use `jsdom`. So it has been changed back,

vitest.config.ts
NabbeunNabi: @ermish The answer seems to be no.
Based on this link and article https://masteringnuxt.com/blog/unit-testing-in-nuxt

And the official docs which include this:

image

ermish (approved):
nuxtified

PR #825, opened 2024-10-16
refactor: move rejectSubmission from component to store

theyokohamalife:
## 🔧 What changed

This PR moves the `rejectSubmission` API call from the top bar component to the `moderationSubmissionsStore`.
Tests will be added following #821

### DIFF
--- a/stores/moderationSubmissionsStore.ts
+++ b/stores/moderationSubmissionsStore.ts
@@ -102,6 +102,11 @@ export const useModerationSubmissionsStore = defineStore(
         }
 
         async function approveSubmission() {
+            if (!selectedSubmissionId.value) {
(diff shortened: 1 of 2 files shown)


stores/moderationSubmissionsStore.ts
```diff
@@ -101,6 +101,27 @@ export const useModerationSubmissionsStore = defineStore(
             }
         }
 
+        async function rejectSubmission() {
+            const facilityInputVariables = {
```
ermish, line 105: similar to your comment on the other PR, let's add a check or log here for selectedSubmissionId

ermish (changes requested):
just one quick change!

ermish (approved):
ship it!

Issue #770, opened 2024-10-03
Show the modal on submission rejection


NabbeunNabi:
# Problem
Rejecting a submission is a big process. So we want to be sure before rejecting the facility. Right now the modal only displays on the `ModEditSubmissionForm` for accepting a facility and the back button. We also want it to show for rejection.

# Success
The modal shows with correct messaging for clicking on the rejection button. Make sure there are tests for this in `moderation.cy.ts`
### Requirements

- [x] Implement modal display on rejection button press.
- [x] Have tests for reject back button display

NabbeunNabi:
This will be ready once this PR is merged https://github.com/ourjapanlife/findadoc-web/pull/769

theyokohamalife:
@NabbeunNabi For the requirements, what do we mean by tests for each back button display? We just need tests for the rejection behavior, right?

NabbeunNabi:
@theyokohamalife 

Yes...but maybe I was trying to be a bit sneaky by saying in this they could add the tests for the others as well since those were not implemented in the previous ones that implemented this behavior.

Should I pull the tests to a separate ticket?

theyokohamalife:
@NabbeunNabi Yeah, I think we should create a separate ticket for unrelated tests.

theyokohamalife:
@NabbeunNabi This ticket is bit confusing, because the title says to show the modal on submission rejection. However, the success and requirements details add additional features. At first glance I thought this ticket was for "Show modal when user clicks reject button."

NabbeunNabi:
@theyokohamalife I updated it so the words only focused on the reject button

PR #826, opened 2024-10-16
refactor: move state to modal component and add emit

theyokohamalife:
Part of #770

## 🔧 What changed
This change adds a couple small improvements to the `Modal` component.
1. Moves the `v-show` to the component itself so we don't need to keep track of it everywhere it is used.
2. Emits `modal-closed` when the user hits the "X" button or clicks off screen. This will be useful for knowing when to reset the state of whatever called the modal (such as `approvingSubmissionFromTopBar`).

### DIFF
--- a/components/MainContentContainer.vue
+++ b/components/MainContentContainer.vue
@@ -7,7 +7,6 @@
         >
             <Loader />
             <Modal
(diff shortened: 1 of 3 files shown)


ermish (approved):
nice!

PR #827, opened 2024-10-18
feat: implement section of form to edit existing facility

### LINKED ISSUE #785: Move Facility related UI from `ModEditSubmissionForm` into its own component
# Problem
The `ModEditSubmissionForm` is way too long. We would like to move the parts of the `<template>` that are related to Facilities into their own component.

# Success
The Facility form is created as its own component with reference to how the UI/UX is displayed and written in #718.
### Requirements

- [x] Move the UI/UX for Facility from the `ModEditSubmissionForm` into its own component
- [x] Matches the styling of #718 
- [x] Has cypress test
- [x] PR includes pictures of the Component

NabbeunNabi:
Resolves #785 

## 🔧 What changed
Before there was no form to edit existing facilities in our database. This now creates the UI for that form that can be used in the `edit-facility` page, in which the functionality can be added later. It is also reusable and can be imported into the `ModEditSubmissionForm` to make the component smaller and easier to read.

## 🧪 Testing instructions
Run the cypress tests. As tests were added for the UI of this component.

## 📸 Screenshots

-   ### Before
N/A
-   ### After
image
image
image
image


components/ModEditFacilityForm.vue
```diff
@@ -1,5 +1,208 @@
...
+        <h1
+            class="mb-3.5 text-start text-primary-text text-3xl font-bold font-sans leading-normal"
+        >
+            {{ $t('modSubmissionForm.facilityHeading') }}
+        </h1>
+        <span class="mb-3.5 text-center text-primary-text text-2xl font-bold font-sans leading-normal">
+            {{ $t('modSubmissionForm.contactInformation') }}
+        </span>
+        <ModInputField
+            data-testid="submission-form-nameEn"
```
theyokohamalife, line 15: I wonder if we should name the test IDs and localization variables something more component specific. Something along the lines of `mod-edit-facility-form-nameEn`. What do you think? Are the localization strings being reused elsewhere?

components/ModEditFacilityForm.vue
```diff
@@ -0,0 +1,290 @@
...
+    addressLine2Ja: ref('') as Ref<string>,
+    // googleMapsFields
+    googlemapsURL: ref('') as Ref<string>,
+    mapLatitude: ref('') as Ref<string>,
+    mapLongitude: ref('') as Ref<string>
+} as { [key: string]: Ref }
+
+const listPrefectureJapanEn: Ref<string[]> = ref([
+    'Hokkaido', 'Aomori', 'Iwate', 'Miyagi', 'Akita',
+    'Yamagata', 'Fukushima', 'Ibaraki', 'Tochigi', 'Gumma', 'Saitama', 'Chiba', 'Tokyo', 'Kanagawa',
```
theyokohamalife, line 279: Perhaps we should move these to the store since they're being used in multiple components. It would be more inline with "DRY" principles.

components/ModEditFacilityForm.vue
NabbeunNabi: We were thinking that. But this component should be reused and similar to the whole edit submission form. So weren't sure how to go about it

components/ModEditFacilityForm.vue
```diff
@@ -0,0 +1,290 @@
...
+    validateCityEn,
+    validateEmail,
+    validateFloat,
+    validatePostalCode,
+    validateWebsite,
+    validateCityJa } from '~/utils/formValidations'
+
+const screenStore = useModerationScreenStore()
+
+const submissionFormFields = {
```
theyokohamalife, line 254: Let's update these variable names as well.

components/ModEditFacilityForm.vue
theyokohamalife: It's fine that elements are being reused. However, IDs should always be unique. Otherwise, it's hard to guarantee that you're actually testing what you think you are.

theyokohamalife:
It's off to a really great start. Left a couple suggestions. Nice work with the PR description and commits! Noticing and improvement. 🙌🏾

components/ModEditFacilityForm.vue
Ntlja: Since this is a component that’ll be reused, should we shorten the id to mod-edit-form-etc? As far as I understood, we’ll use this component for submission forms, facility forms, healthcare prof. forms and so on eventually? Or is that not the case?

NabbeunNabi: So @theyokohamalife we can conditionally render the `id` based on the screen? Is that the best way?

components/ModEditFacilityForm.vue
NabbeunNabi: Makes sense to me

components/ModEditFacilityForm.vue
NabbeunNabi: How will it work with `v-model` if it is being used in the Submission or the Facility. How will we be able to choose between the two?

components/ModEditFacilityForm.vue
theyokohamalife: @Ntlja @NabbeunNabi The name of the component implies that it is for editing facilities. Facilities, healthcare professionals, and submissions are separate entities. If this is meant to be raised for something other than facilities, the component name should be changed.

NabbeunNabi: @theyokohamalife this is meant to be used to edit the facilities portion of an existing facility or a submission facility. This is part of the refactoring of the submission form for being way too long. This was talked about in the long standup 2 weeks ago. That we want to reuse it since the data style is the same for both. And it will make the file much smaller and can be used for the submission or the existing facility

NabbeunNabi: So we can rename the fields `facilityFormFields` but it is so that we can be DRY and not have two duplicate forms for editing and existing facility and one that was submitted since the data is the exact same in terms of what is necessary for our DB

theyokohamalife: If that's the case, the IDs should probably be more component specific as mentioned before. Something like the shorter name @Ntlja  suggested. `mod-edit-facility-nameEn` for example.

components/ModEditFacilityForm.vue
theyokohamalife: @NabbeunNabi I think we should probably move this to a huddle with @ermish. I think I'm also leaning away from trying to have everything as just one form.

theyokohamalife:
@all-contributors 
please add @Ntlja for coding.
please add @Ntlja for design.
please add @NabbeunNabi  for coding.

theyokohamalife:
@all-contributors please add @NabbeunNabi for coding

theyokohamalife:
@all-contributors please add @Ntlja for coding

theyokohamalife:
@all-contributors add @Ntlja for coding

PR #843, opened 2024-10-22
feat: show modal to confirm submission rejection

### LINKED ISSUE #770: Show the modal on submission rejection

theyokohamalife:
Resolves #770 

## 🔧 What changed

1. A modal will now show when the user clicks the "Reject" button. Clicking the confirmation button on the modal will reject the submission and navigate the user back to the moderation screen.
2. The div that was wrapped around the modal was removed, and its classes were moved to the `<Modal>` component.

## TODO
We should probably consider fetching the submissions from the database again once the user clicks the rejection button. Otherwise, it won't appear under Rejected until the page is refreshed.

## 🧪 Testing instructions
1. On your local dev server, select a submission and click the Reject button.
2. Click the X button to close the modal without rejecting the submission.
4. Click the confirmation button on the modal to reject the submission and return to the moderation screen.
5. Refresh the page to confirm that the submission is now rejected.

## 📸 Demo
https://github.com/user-attachments/assets/8b60c892-04f7-4116-a250-72cc9ab9308f


components/ModEditSubmissionForm.vue
```diff
@@ -855,6 +870,17 @@ async function submitCompletedForm(e: Event) {
 
 const syntheticEvent = new Event('submit', { bubbles: true, cancelable: true })
 
+const resetModalRefs = async () => {
+    moderationSubmissionStore.setShowRejectSubmissionConfirmation(false)
+    moderationSubmissionStore.setApprovingSubmissionFromTopBar(false)
+}
+
+const rejectSubmission = async () => {
+    moderationSubmissionStore.rejectSubmission()
```
ermish, line 879: this should be `await`ed

components/ModEditSubmissionForm.vue
```diff
@@ -855,6 +870,17 @@ async function submitCompletedForm(e: Event) {
 
 const syntheticEvent = new Event('submit', { bubbles: true, cancelable: true })
 
+const resetModalRefs = async () => {
+    moderationSubmissionStore.setShowRejectSubmissionConfirmation(false)
+    moderationSubmissionStore.setApprovingSubmissionFromTopBar(false)
+}
+
+const rejectSubmission = async () => {
```
ermish, line 878: should we show a loading indicator here while it's being rejected? We have our loader component we could use

ermish (changes requested):
nice! just a few changes

components/ModEditSubmissionForm.vue
theyokohamalife: That's a good idea! What do you think about showing something like a  "Sending your request" loader? I think it would be better for the user since we aren't actually waiting to load the next page, but we're waiting for the request to be completed.

ermish: sounds like a good next issue! can you make one for it?

NabbeunNabi:
@theyokohamalife Is the ToDo still a thing. I went the your branch and when I reject a submission it changes to rejected when it redirects me

components/ModEditSubmissionForm.vue
theyokohamalife: @ermish Here's the ticket: https://github.com/ourjapanlife/findadoc-web/issues/884

theyokohamalife:
> @theyokohamalife Is the ToDo still a thing. I went the your branch and when I reject a submission it changes to rejected when it redirects me

@NabbeunNabi Here's the follow-up ticket for the TODO: https://github.com/ourjapanlife/findadoc-web/issues/885

test/cypress/e2e/moderationDashboard.cy.ts
```diff
@@ -415,3 +415,59 @@ describe('Moderation Edit Submission Modal', () => {
...
+
+            cy.url({ timeout: 10000 }).should('include', '/moderation')
+
+            cy.wait('@getSubmissions', { timeout: 10000 })
+
+            cy.get('[data-testid="mod-submission-list-item-1"]', { timeout: 10000 }).click()
+        })
+
+        after(() => {
+            Cypress.session.clearCurrentSessionData()
```
ermish, line 459: is there a reason we need this in `after()` and in the `before()`?

ermish: if so, it would be good to add comments explaining why

ermish (approved):
approved! Added a comment to check first

PR #860, opened 2024-10-25
chore: set skip ci to false for all contributors config

theyokohamalife:
We're unable to merge PRs created by all-contributors with `skipCI` set to `true` because of branch protections.


ermish (approved):
Contributors UP!

PR #867, opened 2024-10-28
test: set defaultCommandTimeout to 10000ms for stability

### LINKED ISSUE #866: Move the timeouts in the cypress tests to the `defaultCommandTimeout` provided by Cypress
# Problem
Before we had written in every line to have the timeout be 10000 ms, due to some of our elements taking more than the default of cypress being 4000 ms. This is not dry as cypress allows for a change in the default.
# Success
The `defaultCommandTimeout` in the cypress configuration file is added to the `e2e` testing and set to 10000 ms. The tests still pass and the repeated code of `{ timeout: 10000 }` is removed due to it being unnecessary.
### Requirements

- [x] The `defaultCommandTimeout` in cypress is set to 10000 ms.
- [x] The repeated code is removed
- [x] The tests on production still pass

NabbeunNabi:
Resolves #866 

## 🔧 What changed
Before we were not maintaining good code quality by repeating the same timeout all over the cypress tests to make sure that our elements would be found. Cypress provides this setting in the configuration files as seen in the [documentation](https://docs.cypress.io/app/references/configuration#Timeouts). So we set this in the configuration files and removed the unnecessary code.

## 🧪 Testing instructions
You can run the tests locally with `yarn cypress open`, but the tests in CI/CD reflect the changes that were done as it changed the global timeouts only

### DIFF
--- a/cypress.config.ts
+++ b/cypress.config.ts
@@ -15,7 +15,8 @@ const config = defineConfig({
         experimentalRunAllSpecs: true,
         requestTimeout: 10000,
         responseTimeout: 30000,
-        pageLoadTimeout: 60000
+        pageLoadTimeout: 60000,
+        defaultCommandTimeout: 10000
     }
 })
 
--- a/test/cypress/e2e/about.cy.ts
+++ b/test/cypress/e2e/about.cy.ts
@@ -3,7 +3,7 @@ import enUS from '../../../i18n/locales/en.json'
 describe('About page', () => {
     context('Desktop resolution', () => {
         before(() => {
-            cy.visit('/about', { timeout: 10000 })
+            cy.visit('/about')
         })
 
         beforeEach(() => {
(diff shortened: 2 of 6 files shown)


cypress.config.ts
```diff
@@ -15,7 +15,8 @@ const config = defineConfig({
         experimentalRunAllSpecs: true,
         requestTimeout: 10000,
         responseTimeout: 30000,
-        pageLoadTimeout: 60000
+        pageLoadTimeout: 60000,
+        defaultCommandTimeout: 10000
```
ermish, line 19: this is the original approach we had, but we had issues with the timeouts not being applied correctly 🤔 
I'm not sure if it was due to an unrelated error. I'm nervous to commit, but I see the tests are passing and I think it's much cleaner

ermish (approved):
you need a `timeout` cause your so fast!

PR #875, opened 2024-10-30
feat: add server response type to frontend

### LINKED ISSUE #877: Add a `ServerResponse` type to the frontend
# Problem
Right now we do not have a type for the frontend to return a response from the server to display good error messages to the frontend. Due to the graphQL schema coming built in with a default error message we should add these types as if the frontend and backend are separate units where we do not own the backend.

# Success
The types that match the backend are added to the frontend for us to later display good error messages for the user  upon an error sent from the backend.
### Requirements

- [x] File added with a `ServerResponse` type
- [x] This file matches the types being sent from the backend

NabbeunNabi:
Resolves #877 

## 🔧 What changed
Before we didn't have types that matched our responses from the backend. These have now been added so in the future we can return this type and display error messages to the user.


ermish (approved):
no errors

PR #883, opened 2024-11-06
feat: implement autofill to edit existing facility section

### LINKED ISSUE #863: Autofill the facility section of the form with the chosen facility
# Problem
We now have a facility section of the form but on the `ModerationScreen.EditFacility` we are still not auto filling it with the chosen facility.

# Success
The form should be auto filled with the data and the tests should be written to test that they are auto filled with the chosen facility data.
### Requirements

- [x] Autofill the facility section on the `ModerationScreen.EditFacility` screen with the selected facility
- [x] Cypress tests are written to show that the fields are auto filled with the fake data we provide in our cypress tests

# Implementation notes
Look to see how auto filling is accomplished in the `ModEditSubmissionForm.vue`

NabbeunNabi:
Resolves #863 

## 🔧 What changed
Before we were not autofilling the form for the facility based on the selected Facility from the list. Now we can autofill the form when we go to this route.

## 🧪 Testing instructions
Cypress tests were used to make sure the form was filled with our fake data. You can run by going to this branch and running:
`yarn cypress open` and navigate to the tests for `test/cypress/e2e/moderationEditFacility.cy.ts`

## 📸 Screenshots

-   ### Before
image

-   ### After

https://github.com/user-attachments/assets/ed0300ec-c18d-4dae-94d4-e682f0ec2d86

### DIFF
--- a/components/ModEditFacilitySection.vue
+++ b/components/ModEditFacilitySection.vue
@@ -1,243 +1,247 @@
 <template>
-    <div
-        class="mod-facility-section"
-    >
-        <h1
-            class="mb-3.5 text-start text-primary-text text-3xl font-bold font-sans leading-normal"
+    <div v-if="isFacilitySectionInitialized">
+        <div
+            class="mod-facility-section"
         >
-            {{ $t('modFacilitySection.facilityHeading') }}
-        </h1>
-        <span class="mb-3.5 text-center text-primary-text text-2xl font-bold font-sans leading-normal">
-            {{ $t('modFacilitySection.contactInformation') }}
-        </span>
-        <ModInputField
-            v-model="facilityStore.facilitySectionFields.nameEn"
-            data-testid="mod-facility-section-nameEn"
-            :label="$t('modFacilitySection.labelFacilityNameEn')"
-            type="text"
-            :placeholder="$t('modFacilitySection.placeholderTextFacilityNameEn')"
-            :required="true"
-            :input-validation-check="validateNameEn"
-            :invalid-input-error-message="$t('modFacilitySection.inputErrorMessageFacilityNameEn')"
-        />
-        <ModInputField
-            v-model="facilityStore.facilitySectionFields.nameJa"
-            data-testid="mod-facility-section-nameJa"
-            :label="$t('modFacilitySection.labelFacilityNameJa')"
-            type="text"
-            :placeholder="$t('modFacilitySection.placeholderTextFacilityNameJa')"
-            :required="true"
-            :input-validation-check="validateNameJa"
-            :invalid-input-error-message="$t('modFacilitySection.inputErrorMessageFacilityNameJa')"
-        />
-        <ModInputField
-            v-model="facilityStore.facilitySectionFields.phone"
-            data-testid="mod-facility-section-phone"
-            :label="$t('modFacilitySection.labelFacilityPhoneNumber')"
-            type="text"
-            :placeholder="$t('modFacilitySection.placeholderTextFacilityPhoneNumber')"
-            :required="true"
-            :input-validation-check="validatePhoneNumber"
-            :invalid-input-error-message="$t('modFacilitySection.inputErrorMessageFacilityPhoneNumber')"
-        />
-        <ModInputField
-            v-model="facilityStore.facilitySectionFields.email"
-            data-testid="mod-facility-section-email"
-            :label="$t('modFacilitySection.labelFacilityEmail')"
-            type="email"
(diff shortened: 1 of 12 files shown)


components/ModSubmissionListContainer.vue
```diff
@@ -141,17 +143,27 @@ const submissionListItemTableColumns = computed(() => {
...
 
 const handleClickToFacilityForm = (id: string) => {
-    useModerationScreenStore().setActiveScreen(ModerationScreen.EditFacility)
-    modSubmissionsListStore.selectedFacilityId = id
+    facilitiesStore.selectedFacilityId = id
+    facilitiesStore.setSelectedFacilityData(id)
+    moderationScreenStore.setActiveScreen(ModerationScreen.EditFacility)
 }
+
+/** This ensures that the data for the selected facility is loaded before mounting the component,
```
theyokohamalife, line 161: Non-blocking: I don't think it has been formally decided if comments in our `.vue` files should be written with `//` or `/** **/`, but we should probably pick one and stick with it throughout the codebase. Something we should probably discuss on Slack!

stores/facilitiesStore.ts
```diff
@@ -29,9 +30,67 @@ export const useFacilitiesStore = defineStore(
...
+            selectedFacilityData.value = facilityData.value
+                .find((facility: Facility) => facility.id === facilityId)
+            initializeFacilitySectionValues(selectedFacilityData.value)
+        }
+
+        /**This function is a helper that checks for if the data passed is a Facility or Submission
+            in order to properly handle the conditional to autofill this section of the form **/
+        function isFacility(data: Facility | Submission): data is Facility {
+            return (data as Submission).facility === undefined
+        }
```
theyokohamalife, line 47: We may have to rethink this function a bit. It's possible for it to always return True, because `data.facility` can be `undefined` for both a `Facility` and a `Submission`:
Screenshot 2024-11-07 at 1 06 14 AM

NabbeunNabi: You are correct. Actually I believe I was overthinking it.

When we eventually use the facility section in the submission that autofill can be in the submission store since I am autofilling based on the path and screen we are on which already is choosing whether it is a submission or existing facility.

I'll cut this and the autofill for submission out of this PR and wait until later when there is more refactoring.

components/ModSubmissionListContainer.vue
```diff
@@ -141,17 +143,27 @@ const submissionListItemTableColumns = computed(() => {
...
+    thus avoiding a flash in the ui loading the refs**/
+onBeforeRouteLeave((to, _, next) => {
```
ermish, line 163: while this works, it would be better to have an implementation on the page itself for two reasons: 
1) Some edge cases will circumvent this. Ex: what if a moderator wants  to send a link to another moderator directly to that facility? This wouldn't run in this case
2) When you read the facility page, you don't realize that there's logic outside of the page here that affects it. Ideally you  don't want related logic outside of a component.

ermish: on the page, I believe we could use beforeMount or something similar. Basically set it before the render is displayed

ermish (changes requested):
one change, then it's ready!

components/ModEditFacilitySection.vue
```diff
@@ -266,4 +270,38 @@ const listPrefectureJapanJa: Ref<string[]> = ref([
...
+    // Wait for the route to be fully resolved
+    await nextTick()
+
+    if (!facilityStore.facilityData.length) {
+        await facilityStore.getFacilities()
+    }
+
+    // Ensure the route param `id` is available before proceeding
+    const id = route.params.id
+    if (id) {
```
ermish, line 288: should throw  an error here if it's not available. I'd do an inverted if.
```
if(!id) {
    console.error(...)
   toast()
return
}
```

components/ModEditFacilitySection.vue
```diff
@@ -266,4 +270,38 @@ const listPrefectureJapanJa: Ref<string[]> = ref([
...
+        facilityStore.selectedFacilityId = id as string
+    }
+
+    // Set the active screen and ensure the UI state is consistent
+    screenStore.setActiveScreen(ModerationScreen.EditFacility)
+
+    // Make sure the selected facility data is up-to-date
+    if (facilityStore.selectedFacilityId) {
```
ermish, line 296: I don't think we need this anymore if it's always set earlier

ermish (changes requested):
re-reviewed

ermish (approved):
ship it!

PR #887, opened 2024-11-07
fix: improve submission form ui and ux

### LINKED ISSUE #873: BUG: fix `SubmissionForm.vue` - No loading for stores
# Problem
Currently, the data from the [submit form](https://www.findadoc.jp/submit) is loading without a loading indicator, causing the inputs to reset, resulting in a flickering effect.

# Success
Load the form only after the stores have finished loading.

### Requirements
- [x] Display loading
- [x] The form loads properly

NabbeunNabi:
Resolves #873

## 🔧 What changed

- the `ServerResponse` type was changed to match what is delivered from the `resolvers` from the backend based on the built in graphQLError we must covert our backend to before messaging to the frontend.
- Error toasts and success toasts are displayed to the user on the frontend depending on the submission status
- The api call now uses the built in retry and returns the response where we have access to the submission and to the errors
- Updated `SubmissionForm.vue` to use lifecycle hooks in order to render the form once the refs are loaded within the component
- Component specific refs for the submission were moved to the component and our of the store
- The thank you component has a button to submit another doctor and the forms are conditionally rendered in order to remount and reset the form
- Locale keys added to English and updated throughout the locale files.

## 📸 Screenshots

-   ### Before

image

https://github.com/user-attachments/assets/293739a1-9142-461b-9082-0b3279476e95

-   ### After
Testing locally to show it goes into the DB
image

https://github.com/user-attachments/assets/eaed5c68-31a6-4ced-ac03-0014fc1bb173

### DIFF
--- a/components/ModEditSubmissionForm.vue
+++ b/components/ModEditSubmissionForm.vue
@@ -480,7 +480,7 @@
             </select>
         </div>
         <button
-            type="submit"
+            type="button"
             class="bg-currentColor text-white font-bold py-2 px-4 my-2 rounded w-56"
             @click="submitUpdatedSubmission"
         >
@@ -521,6 +521,7 @@ import SVGTrashCan from '~/assets/icons/trash-can.svg'
 import SVGProfileIcon from '~/assets/icons/profile-icon.svg'
 import { triggerFormValidationErrorMessages } from '~/utils/triggerFormValidationErrorMessages'
 import { useLocaleStore, localeDisplayOptions, type LocaleDisplay } from '~/stores/localeStore'
+import { handleServerErrorMessaging } from '~/utils/handleServerErrorMessaging'
 /**
 This initalizes the variable that needs to be set on mount.
 If this is set as a const the build will fail since the plugin
@@ -805,18 +806,21 @@ async function submitUpdatedSubmission(e: Event) {
             ]
         }
     }
-    try {
-        const result = await moderationSubmissionStore.updateSubmission(submissionInputVariables)
-        const submissionResult = result as { updateSubmission: Submission }
-        // This updates the submission in the form with the values stored in the db on success
-        initializeSubmissionFormValues(submissionResult.updateSubmission as Submission)
-        toast.success(t('modSubmissionForm.successMessageUpdated'))
-        if (moderationSubmissionStore.updatingSubmissionFromTopBar) {
-            router.push('/moderation')
-            moderationSubmissionStore.setUpdatingSubmissionFromTopBar(false)
-        }
-    } catch {
-        toast.error(t('modSubmissionForm.errorMessageUpdated'))
+
+    const result = await moderationSubmissionStore.updateSubmission(submissionInputVariables)
+    // This is used in the component and not graphQL call as it is user messaging and needs the mounted toast library
+    if (result?.errors?.length) {
+        handleServerErrorMessaging(result.errors, toast, t)
+        return
+    }
+
+    const submissionResult = result.data
+    // This updates the submission in the form with the values stored in the db on success
+    if (submissionResult) initializeSubmissionFormValues(submissionResult)
+    toast.success(t('modSubmissionForm.successMessageUpdated'))
+    if (moderationSubmissionStore.updatingSubmissionFromTopBar) {
+        router.push('/moderation')
+        moderationSubmissionStore.setUpdatingSubmissionFromTopBar(false)
     }
 }
 
--- a/components/SubmissionCompleted.vue
+++ b/components/SubmissionCompleted.vue
@@ -14,18 +14,40 @@
             {{ $t('thankYouPage.heading') }}
         </p>
         <p
-            class="mb-40 text-neutral-600 text-sm font-bold font-sans"
+            class="mb-20 text-center text-neutral-600 text-sm font-bold font-sans"
         >
             {{ $t('thankYouPage.submission') }}
         </p>
-        <NuxtLink
-            class="flex"
-            to="/"
+        <div
+            class="action-button-container flex flex-col"
         >
             <button
                 class="px-20 py-3 rounded-full bg-currentColor w-96 text-center
+        text-white text-base font-medium font-sans mb-2"
+                @click="resetFormCompleted"
+            >
+                {{ $t('thankYouPage.submitAnotherDoctor') }}
+            </button>
+            <NuxtLink
+                class="flex"
+                to="/"
+            >
+                <button
+                    class="px-20 py-3 rounded-full bg-currentColor w-96 text-center
         text-white text-base font-medium font-sans"
-            >{{ $t('thankYouPage.home') }}</button>
-        </NuxtLink>
+                >{{ $t('thankYouPage.home') }}</button>
+            </NuxtLink>
+        </div>
     </div>
 </template>
+
+<script setup lang="ts">
+import { useSubmissionStore } from '~/stores/submissionStore'
+
+const submissionStore = useSubmissionStore()
+
+const resetFormCompleted = () => {
+    submissionStore.submissionCompleted = false
+}
+</script>
+
--- a/components/SubmissionForm.vue
+++ b/components/SubmissionForm.vue
@@ -1,162 +1,190 @@
 <template>
-    <form class="flex flex-col place-self-center">
-        <h1
-            data-testid="submit-heading"
-            class="mb-3.5 text-center text-primary-text text-2xl font-bold font-sans leading-normal"
+    <div>
+        <Loader v-show="!isFormInitialized" />
+        <form
+            v-show="isFormInitialized"
+            class="flex flex-col place-self-center"
         >
-            {{ $t('submitPage.heading') }}
-        </h1>
-        <p
-            data-testid="submit-subheading"
-            class="mb-10 w-96 text-center text-primary-text-muted text-sm font-normal font-sans"
-        >
-            {{ $t('submitPage.subheading') }}
-        </p>
-        <span
-            class="mb-2 text-primary-text text-sm font-normal font-sans"
-        >{{ $t('submitPage.googleMaps') }}</span>
-        <p
-            v-show="!isValidInput.googleMapsUrl.value"
-            class="text-error text-xs font-sans"
-        >
-            {{ $t('submitPage.googleMapsValidation') }}
-        </p>
-        <input
-            v-model="submissionStore.location"
-            data-testid="submit-input-googlemaps"
-            type="text"
-            required
-            class="mb-5 px-3 py-3.5 w-96 h-12 bg-secondary-bg rounded-lg border border-primary-text-muted
-            text-primary-text text-sm font-normal font-sans placeholder-primary-text-muted"
-            :placeholder="$t('submitPage.location')"
-            @blur="initialValidationCheck(submissionStore.location, 'googleMaps')"
-        >
-        <span
-            class="mb-2 text-primary-text text-sm font-normal font-sans"
-        >{{ $t('submitPage.healthcareProfessionalName') }}</span>
-        <p class="flex text-error text-xs font-sans">
-            <span
-                :class="[
-                    !isValidInput.lastName.value ? 'visible' : 'invisible',
-                    'w-44',
-                    'mr-1',
-                ]"
+            <h1
+                data-testid="submit-heading"
+                class="mb-3.5 text-center text-primary-text text-2xl font-bold font-sans leading-normal"
             >
-                {{ $t('submitPage.lastNameValidation') }}
-            </span>
-            <span
-                :class="[
-                    !isValidInput.firstName.value ? 'visible' : 'invisible',
-                    'w-44',
-                ]"
+                {{ $t('submitPage.heading') }}
+            </h1>
+            <p
+                data-testid="submit-subheading"
+                class="mb-10 w-96 text-center text-primary-text-muted text-sm font-normal font-sans"
             >
-                {{ $t('submitPage.firstNameValidation') }}
-            </span>
-        </p>
-        <div>
+                {{ $t('submitPage.subheading') }}
+            </p>
+            <span
+                class="mb-2 text-primary-text text-sm font-normal font-sans"
+            >{{ $t('submitPage.googleMaps') }}</span>
+
             <input
-                v-model="submissionStore.lastName"
-                data-testid="submit-input-lastname"
-                class="mb-5 mr-1 px-3 py-3.5 w-48 h-12 bg-secondary-bg rounded-lg border border-primary-text-muted
-            text-primary-text text-sm font-normal font-sans placeholder-primary-text-muted"
+                v-model="location"
+                data-testid="submit-input-googlemaps"
                 type="text"
                 required
-                maxlength="30"
-                :placeholder="$t('submitPage.lastName')"
-                @blur="initialValidationCheck(submissionStore.lastName, 'lastName')"
+                class="px-3 py-3.5 w-96 h-12 bg-secondary-bg rounded-lg border border-primary-text-muted
+                text-primary-text text-sm font-normal font-sans placeholder-primary-text-muted"
+                :placeholder="$t('submitPage.location')"
+                @blur="initialValidationCheck(location, 'googleMaps')"
             >
-            <input
-                v-model="submissionStore.firstName"
-                data-testid="submit-input-firstname"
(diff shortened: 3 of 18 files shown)


components/SubmissionCompleted.vue
```diff
@@ -14,18 +14,40 @@
...
         <p
-            class="mb-40 text-neutral-600 text-sm font-bold font-sans"
+            class="mb-20 text-center text-neutral-600 text-sm font-bold font-sans"
         >
             {{ $t('thankYouPage.submission') }}
         </p>
-        <NuxtLink
-            class="flex"
-            to="/"
+        <div
```
ermish, line 21: add an id or class to explain what this div is

components/SubmissionCompleted.vue
```diff
@@ -14,18 +14,40 @@
...
+            class="flex flex-col"
         >
             <button
-                class="px-20 py-3 rounded-full bg-currentColor w-96 text-center
+                class="px-20 py-3 rounded-full bg-currentColor max-w-96 text-center
```
ermish, line 25:
```suggestion
                class="px-20 py-3 rounded-full bg-currentColor w-96 text-center
```

components/SubmissionForm.vue
```diff
@@ -1,163 +1,185 @@
...
-                maxlength="30"
-                :placeholder="$t('submitPage.firstName')"
-                @blur="initialValidationCheck(submissionStore.firstName, 'firstName')"
+            <span
+                class="mb-2 text-primary-text text-sm font-normal font-sans"
+            >{{ $t('submitPage.healthcareProfessionalName') }}</span>
+            <p class="flex text-error text-xs font-sans">
+                <span
+                    :class="[
+                        !isValidInput.lastName.value ? 'visible' : 'invisible',
```
ermish, line 44: I think you can just use v-show or v-if here.
If the page layout messes up, you can have this in a container with a fixed size that doesn't disappear and then just conditionally render the element within.

components/SubmissionForm.vue
```diff
@@ -1,163 +1,185 @@
...
+                        'w-44',
+                        'mr-1',
+                    ]"
+                >
+                    {{ $t('submitPage.lastNameValidation') }}
+                </span>
+                <span
+                    :class="[
```
ermish, line 52: ♻️

components/SubmissionForm.vue
```diff
@@ -1,163 +1,185 @@
...
-        >
-            <option
-                v-for="(locale, index) in localeStore.localeDisplayOptions"
-                :key="index"
-                selected
-                :value="locale.code"
+                <option
+                    v-for="(locale, index) in localeStore.localeDisplayOptions"
+                    :key="index"
+                    selected
```
ermish, line 103: double check this. I'm pretty sure this is making EVERY option selected. you might want to just have the first one in the array as selected or have none of them.

components/SubmissionForm.vue
```diff
@@ -1,163 +1,185 @@
...
-             text-primary-inverted font-medium font-sans"
-            @click="validateFields"
-        >
-            {{ $t('submitPage.submitButton') }}
-        </button>
-    </form>
+                {{ $t('submitPage.invalidOption') }}
+            </p>
+            <select
+                v-model="selectLanguage2"
```
ermish, line 119: future feature: we should be able to support `n` number of languages.

components/SubmissionForm.vue
```diff
@@ -1,163 +1,185 @@
...
+                data-testid="submit-select-language2"
+                class="mb-5 px-3 py-3.5 w-96 h-12 bg-primary-text-inverted rounded-lg border border-primary-text-muted
+                        text-primary-text text-sm font-normal font-sans placeholder-primary-text-muted"
+                :placeholder="$t('submitPage.selectLanguage2')"
+                @change="initialValidationCheck(selectLanguage1, 'secondaryLanguage')"
+            >
+                <option
+                    v-for="(locale, index) in localeStore.localeDisplayOptions"
+                    :key="index"
+                    selected
```
ermish, line 129: ♻️

components/SubmissionForm.vue
```diff
@@ -198,7 +220,48 @@ const validateFields = (e: Event) => {
         e.preventDefault()
         return
     }
-    submissionStore.submit()
+    submitNewSubmission()
```
ermish, line 223: should await this function

ermish: also, this shouln't be called in validateFields. It should be the other way around where submit should call validateFields.

components/SubmissionForm.vue
```diff
@@ -198,7 +220,48 @@ const validateFields = (e: Event) => {
...
+        googleMapsUrl: location.value,
+        healthcareProfessionalName: `${firstName.value} ${lastName.value}`,
+        spokenLanguages: spokenLanguages,
+        notes: otherNotes.value
+    } }
+
+    const response = await submissionStore.createNewSubmission(newSubmission)
+    if (response?.errors?.length) {
+        response.errors.map(error => {
+            toast.error((t(`serverErrorMessages.${error.extensions.code}`)))
```
ermish, line 248: should check if the code exists first. The server could have a new one that we haven't supported yet on the front end. In the case where it doesn't exist, we can just send a generic error message to the user and log the error code that was unrecognized.

components/SubmissionForm.vue
```diff
@@ -232,29 +295,38 @@ const initialValidationCheck = async (inputValue: string, field: string) => {
...
     if (validationCheckedPreviously.secondarySpokenLanguage.value) {
         isValidInput.secondarySpokenLanguage.value = validations.validateSecondSpokenLanguage(newValue)
     }
 })
+
+onMounted(async () => {
+    toast = useToast()
+
+    resetForm()
+    await nextTick()
```
ermish, line 328: should leave comment for people that don't know what this does

components/SubmissionForm.vue
```diff
@@ -232,29 +295,38 @@ const initialValidationCheck = async (inputValue: string, field: string) => {
...
+
+    isSubmissionFormReady.value = true
```
ermish, line 330: nit: could make this slightly more clear to indicate it's around being loaded or not like `isFormLoaded` or `isFormInitialized`

components/SubmissionForm.vue
```diff
@@ -1,163 +1,185 @@
 <template>
-    <form class="flex flex-col place-self-center">
-        <h1
-            data-testid="submit-heading"
-            class="mb-3.5 text-center text-primary-text text-2xl font-bold font-sans leading-normal"
+    <div>
+        <form
+            v-show="isSubmissionFormReady"
```
ermish, line 4: should we add a loader for this? (you could test using the chrome debugger tools fake 3g connection and see if the user is waiting on a blank page.

stores/submissionStore.ts
```diff
@@ -1,57 +1,45 @@
...
 import { defineStore } from 'pinia'
-import { ref, type Ref } from 'vue'
-import { gqlClient } from '../utils/graphql.js'
-import type { Locale, CreateSubmissionInput, Submission } from '~/typedefs/gqlTypes'
+import { type Ref, ref } from 'vue'
+import { gqlClient, graphQLClientRequestWithRetry } from '../utils/graphql.js'
+import type { ServerResponse, ErrorCode, ServerErrorResponse } from '~/typedefs/serverResponse.js'
+import type { MutationCreateSubmissionArgs, Submission } from '~/typedefs/gqlTypes'
+
+export interface SubmissionFormInputFields {
```
ermish, line 8: this isn't used. DELETE 🔥 🤖 🔥

stores/submissionStore.ts
```diff
@@ -1,57 +1,45 @@
...
+            response = await graphQLClientRequestWithRetry(
+                gqlClient.request.bind(gqlClient),
+                createSubmissionMutation,
+                newSubmission
+            ) as ServerResponse<Submission>
         } catch (e) {
             console.error(`There was an error creating the submission ${e}`)
+            const ServerErrorResponse = e as ServerErrorResponse
+
+            ServerErrorResponse.response.errors.forEach(errorResponse => {
```
ermish, line 29: It would be more readable if you used a `map()` function and set the response.errors to the new map array. 
Tip:
any time you're changing the shape of an object or iterating through something that you are going to assign to another variable, use `map()`
`forEach()` is better for iterating and doing an action on each item. (like rendering or saving each one)

ermish, line 29: I think this error handling is nice and since we want to do something similar on each request to the server, we should move this into the request method so every request doesn't have to write the same logic

typedefs/serverResponse.ts
```diff
@@ -1,40 +1,29 @@
...
+    locations: ErrorLocation[]
+    extensions: {
+        code: ErrorCode
+    }
 }
 
 export enum ErrorCode {
-    SERVER_ERROR = 'SERVER_ERROR',
-    NOT_FOUND = 'NOT_FOUND',
-    REQUIRED = 'REQUIRED',
```
ermish, line 16: why get rid of these? We might as well leave them even if we're not using them

ermish (changes requested):
nice improvements! sugoi!

components/SubmissionForm.vue
NabbeunNabi: @ermish That is a really cool tool I didn't know about. It is definitely necessary in that case

components/SubmissionForm.vue
NabbeunNabi: yea it isn't necessary

components/ModEditSubmissionForm.vue
```diff
@@ -803,7 +803,7 @@ async function submitUpdatedSubmission(e: Event) {
     }
     try {
         const result = await moderationSubmissionStore.updateSubmission(submissionInputVariables)
-        const submissionResult = result as { updateSubmission: Submission }
+        const submissionResult = result as unknown as { updateSubmission: Submission }
```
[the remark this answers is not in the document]
NabbeunNabi: @ermish It says because of the generic data type within the function call here I have to convert it to unknown since it uses `T` in the utility function before converting it to its type

utils/graphql.ts
```diff
@@ -24,29 +25,42 @@ export const initializeGqlClient = () => {
...
+
+            // This is a consistent error messaging no matter the type of query or mutation
+            console.error(`There was an error executing the request: ${error}`)
+            const serverError = error as ServerErrorResponse
+
+            // This map transforms errors if they exist
+            const errors = serverError.response?.errors?.map(errorResponse => ({
+                message: errorResponse.message,
+                locations: errorResponse.locations,
+                extensions: { code: errorResponse.extensions.code as ErrorCode }
```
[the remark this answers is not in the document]
NabbeunNabi: What do you mean? the map should be called code? Or you saying to change the extensions part to `code`. But this is just replicating the `graphql` error being sent to the frontend

components/ModEditSubmissionForm.vue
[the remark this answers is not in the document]
ermish: re-reviewed, for this, look at where we define `T`. You can set constraints for T

ermish (changes requested):
re-reviewed! just the two comments left

typedefs/serverResponse.ts
```diff
@@ -1,13 +1,26 @@
...
+
+export type ErrorLocation = {
+    line: number
+    column: number
+}
+
+export type ServerError = {
+    message: string
+    locations: ErrorLocation[]
+    extensions: {
```
ermish, line 21: let's remove extensions and just have `code`

utils/graphql.ts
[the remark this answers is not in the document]
NabbeunNabi: I think I did it

NabbeunNabi:
@ermish Figured out the types and also made a utility function since we want error messaging to be shown that can be reused since our messages is going to come from the server

components/ModEditSubmissionForm.vue
```diff
@@ -805,18 +808,20 @@ async function submitUpdatedSubmission(e: Event) {
...
-        toast.success(t('modSubmissionForm.successMessageUpdated'))
-        if (moderationSubmissionStore.updatingSubmissionFromTopBar) {
-            router.push('/moderation')
-            moderationSubmissionStore.setUpdatingSubmissionFromTopBar(false)
-        }
-    } catch {
-        toast.error(t('modSubmissionForm.errorMessageUpdated'))
+
+    const result = await moderationSubmissionStore.updateSubmission(submissionInputVariables)
+    const submissionResult = result as ServerResponse<Mutation>
```
ermish, line 813: let's set this as  a  return type of the `updateSubmission()`  function so the caller doesn't have to  set and figure out the types each time

NabbeunNabi: So `updateSubmission` is `Maybe<Submission>` so I will set it to `Submission` as the errors will be caught in the following part

stores/moderationSubmissionsStore.ts
```diff
@@ -85,18 +87,25 @@ export const useModerationSubmissionsStore = defineStore(
...
+            if (response.errors?.length) {
                 setDidMutationFail(true)
                 setUpdatingSubmissionFromTopBar(false)
             }
+
+            serverResponse.data = response.data?.updateSubmission
+            serverResponse.errors = response.errors ? response.errors : []
+            serverResponse.hasErrors = response.hasErrors
+
+            return serverResponse
```
NabbeunNabi, line 108: @ermish I had to change the way the function was written in order to show what type of mutation it was. And make a new object to send since we do a Mutation, and get back the type of updateSubmission. But the api call doesn't work this way. They can sese that it is updateSubmission on line 104. But their isn't really a way to access the key since the key as a type that I can see as that causes a mismatch of what we get

NabbeunNabi: If I try to specify the type I get this
image
From trying the following code

```

        async function updateSubmission(submission: MutationUpdateSubmissionArgs):
        Promise<ServerResponse<Mutation['updateSubmission']>> {
            const response = await graphQLClientRequestWithRetry<Mutation['updateSubmission']>(
                gqlClient.request.bind(gqlClient),
                updateFacilitySubmissionGqlMutation,
                submission
            )

            if (response.errors?.length) {
                setDidMutationFail(true)
                setUpdatingSubmissionFromTopBar(false)
            }

            return response
        }
        ```

Comment (NabbeunNabi) [reply stores/moderationSubmissionsStore.ts:108]: I then cannot access this key due to typing

File: utils/handleServerErrorMessaging.ts [hunk 21]
```diff
@@ -0,0 +1,18 @@
+import type { ToastInterface } from 'vue-toastification'
+import type { ServerError } from '~/typedefs/serverResponse'
+
+export function handleServerErrorMessaging(
```
ermish, line 4: nit: you could put this in the graphql file since I don't think it'll be used outside of it

components/ModEditSubmissionForm.vue
```diff
@@ -805,18 +806,21 @@ async function submitUpdatedSubmission(e: Event) {
...
+    const result = await moderationSubmissionStore.updateSubmission(submissionInputVariables)
+    // This is used in the component and not graphQL call as it is user messaging and needs the mounted toast library
+    if (result?.errors?.length) {
+        handleServerErrorMessaging(result.errors, toast, t)
+        return
+    }
+
+    const submissionResult = result.data
+    // This updates the submission in the form with the values stored in the db on success
+    if (submissionResult) initializeSubmissionFormValues(submissionResult)
```
ermish, line 819: this looks cleaner now! nice work

stores/moderationSubmissionsStore.ts
ermish: you've done some good research on typing  😸  this is honestly an issue in pretty much every typed language. 

My advice at this point is it's more hassle to make it work for the nested Mutation type. (hence why it's not on the server side) 
I think it's fine to cast when accessing the inner object in the <mutation>, but the <serverResponse> and <Mutation> should at least be in the function's return type. So basically, `Promise<ServerResponse<Mutation>>` is good 👍 

Looking at what you got above, I think that is ideal ✨

ermish (approved):
woo! nice work updating through the changes and getting this through

utils/handleServerErrorMessaging.ts
NabbeunNabi: Due to toast being accessed on build and also translation not be available. We talked about this. This can be addressed in a future issue if we can figure out a better way to load toasts not on mount and have translations available outside of the components.

PR #901, opened 2024-11-18
fix(lang): update language dropdown options for search and i18n

### LINKED ISSUE #900: Remove untranslated Swahili from dropdown options for language selection
# Problem
Right now we have a dropdown menu item to choose Swahili for translation. Even though we have a doctor that speaks Swahili we do not currently have translations at all on the website for the language. Our selecting of a language of a doctor should be all languages whether it is translated and it should be independent of what is translated as a choice.

# Success
The language dropdown items do not include Swahili. The select a language a doctor speaks include all the language options we provide as `localeDisplayOptions`
### Requirements

- [x] The language translation dropdown items do not include Swahili
- [x] The language selection of the doctors themselves include all the languages we plan to provide going forward (we have a doctor that speaks Tagalog but it isn't an option to choose) 
- [x] Screenshots of before and after are included in the PR

NabbeunNabi:
Resolves #900 

## 🔧 What changed
1. Swahili was removed from the options of what we have translated on our site. Even though we have a doctor that speaks Swahili. There is no translation of Swahili.
2. The search bar languages were changed to all the ones we include as `localeDisplayOptions` since even though we might not have the site translated yet within the language. We might have a doctor who speaks that language. As we do for Tagalog.

## 📸 Screenshots

-   ### Before
image
image

-   ### After
image
image

### DIFF
--- a/components/SearchBar.vue
+++ b/components/SearchBar.vue
@@ -115,21 +115,28 @@ const specialtiesStore = useSpecialtiesStore()
 
 await locationsStore.fetchLocations()
 
-const languageOptions = [{
-    code: '',
-    simpleText: '----Any----',
-    displayText: '----Any----'
-}, ...localeStore.mvpLocaleDisplayOptions]
+const languageOptions = localeStore.localeDisplayOptions
+const languageOptionsWithPlaceHolder = setSearchBarLanguageDropdownOptions()
 
 const locationDropdownOptions: Ref<string[]> = ref(locationsStore.citiesDisplayOptions)
 const specialtyDropdownOptions: Ref<SpecialtyDisplayOption[]> = ref(specialtiesStore.specialtyDisplayOptions)
-const languageDropdownOptions: Ref<LocaleDisplay[]> = ref(languageOptions)
+const languageDropdownOptions: Ref<LocaleDisplay[]> = ref(languageOptionsWithPlaceHolder)
 
 const selectedSpecialty: Ref<Specialty | string> = ref('')
 const selectedLocation: Ref<string> = ref('')
 const selectedLanguage: Ref<Locale | string> = ref('')
 
 const placeHolderTextDisplay = '----Any----'
+
+function setSearchBarLanguageDropdownOptions() {
+    // This will remove any codes code that is falsy
+    const localeOptionsWithLocaleCodes = languageOptions.filter(locale => locale.code)
+    // This will add the placeholder text
+    localeOptionsWithLocaleCodes
+        .unshift({ code: '', simpleText: placeHolderTextDisplay, displayText: placeHolderTextDisplay })
+
+    return localeOptionsWithLocaleCodes
+}
 
 interface CityDisplayItems {
     [cityName: string]: {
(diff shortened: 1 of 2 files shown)


ermish:
This feels like a backwards direction. 
Can we add Swahili translations instead of removing the language?

NabbeunNabi:
> This feels like a backwards direction. Can we add Swahili translations instead of removing the language?

I would love to do that. But currently the standards are that a native speaker needs to translate the languages. If we could use a translation API or Chatgpt to translate our text and the have a native speaker verify it that would be great. In this case @ermish it is something that has no functionality and shows that we have a language translated we do not. I would be happy to search for alternatives.

So currently the solutions I can see are:
1. Remove the facade that we have it translated in Swahili but show all the options to select doctors (As I said we have two doctors who speak Tagalog and didn't have a search possibility and that could be the same going forward.
2. Change our standards on who translates our sites to an API of some sort and let those contact us if they find flaws as a native speaker
3. Contact someone who is a native Swahili speaker. But I do not know any as of this moment.

@theyokohamalife any thoughts?

ermish:
> > This feels like a backwards direction. Can we add Swahili translations instead of removing the language?
> 
> I would love to do that. But currently the standards are that a native speaker needs to translate the languages. If we could use a translation API or Chatgpt to translate our text and the have a native speaker verify it that would be great. In this case @ermish it is something that has no functionality and shows that we have a language translated we do not. I would be happy to search for alternatives.
> 
> So currently the solutions I can see are:
> 
> 1. Remove the facade that we have it translated in Swahili but show all the options to select doctors (As I said we have two doctors who speak Tagalog and didn't have a search possibility and that could be the same going forward for other languages)
> 2. Change our standards on who translates our sites to an API of some sort and let those contact us if they find flaws as a native speaker
> 3. Contact someone who is a native Swahili speaker. But I do not know any as of this moment.
> 
> @theyokohamalife any thoughts?

I think this is a good question for @theyokohamalife to confirm. After thinking it through again, option #1 maybe is the best and we can provide a way to request a language or contribute to one. Perhaps we can make issues for two future features and go with your PR here

components/SearchBar.vue
```diff
@@ -131,6 +127,9 @@ const selectedLanguage: Ref<Locale | string> = ref('')
 
 const placeHolderTextDisplay = '----Any----'
 
+languageOptions[0].simpleText = placeHolderTextDisplay
```
ermish, line 130: is this going to overwrite the language at `0`?

ermish: you can use`[ `unshift()` ](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/unshift)`for this

theyokohamalife:
Approach #1 sounds good to me.

components/SearchBar.vue
NabbeunNabi: There is no language at 0 in these display options. I thought time complexity wise this would be better than unshifting and shifting. It is written this way as for some components you want the first in a select field to be empty. This is a special case for the search bar. @ermish What do you think?

components/SearchBar.vue
```diff
@@ -131,6 +127,9 @@ const selectedLanguage: Ref<Locale | string> = ref('')
...
+languageOptions[0].displayText = placeHolderTextDisplay
```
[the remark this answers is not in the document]
NabbeunNabi: This is related to the other one. I am not sure as the best way. That was just my thought process. Would it be acceptable if comments are left explaining it?

ermish (approved):
shift it!

Issue #538, opened 2024-06-07
Enable Cypress tests to run using fake data


ermish:
# Problem
Our cypress tests are running against the api and every test is pulling fresh data which can accrue costs and eat our bandwith limits for our API + Databases. 
To solve this, we can have cypress run against fake data and not call the API.  Cypress has a [guide to this](https://docs.cypress.io/guides/end-to-end-testing/testing-your-app#Seeding-data)

To ensure we can still connect to the API successfully, we should be able to toggle between the real API or fake data.

# Success

We can run against fake data or connect to the API with a simple toggle

### Requirements

- [ ] Update `graphql.ts` to include a boolean that toggles between using fake data or the real API calls.
- [ ] Create fake data to mimic the API calls that we use.
- [ ] Update the tests to toggle this boolean on or off before any tests are run.
- [ ] Connect this toggle to the yarn script so we can run with fake or real data locally or on CI, using commands like `yarn test` and `yarn test:realdata`.
- [ ] Create a custom command that acts like an intercept that runs before each suite to intercept every POST request.
- [ ] Create an ESLint rule to prevent the use of intercept inside a spec file for easier management of fake/real data.
- [ ] Create a custom command for the login routine.
- [ ] Refactor tests to adhere to the new intercept rule.
- [ ] Improve test efficiency by removing unnecessary assertions.

NabbeunNabi:
@ermish is this separate to the quick fix to keep it from posting submissions. Or do you want this all to be the same PR?

ermish:
Likely separate, but the quick fix could reference this as this is a bit broader. I think it depends on if you're doing a quick hack for the submissions or if you're trying to do a proper cypress mock

NabbeunNabi:
Blocked while @ermish and team discuss the best way to create a testing environment

NabbeunNabi:
Fake data is being used and we will use snapshot testing. Need to update this idea

Issue #531, opened 2024-06-04
Display the healthcare professionals 


Anissa3005:
# Problem
This issue relies: 
#529 
#530 

At the moment we only can search using the Healthcare Professional ID, so there will not be a list of Healthcare Professional since the ID is unique.
We need to display the healthcare professionals we fetch. When there are no healthcare professionals found we need to display a message.
# Success
![Image](https://github.com/ourjapanlife/findadoc-web/assets/114712265/02849810-6c13-4b79-a345-e2816832c8b9)

### Requirements

- [ ] Display the healthcare professionals data under the search bar
- [ ] Display a message when no healthcare professionals are found

PR #902, opened 2024-11-18
feat: implement update existing facility functionality

### LINKED ISSUE #796: Add functionality to Mod Facility Form for updating an existing facility
# Problem
We need the facilites in our database to be able to be updated based on the UI of the form in #785. Right now there is no `graphQL `mutation to our backend to update an existing facility.

# Success
The functionality with the `graphQL` mutation is implemented from the Mod Facility Form in the moderation panel.
### Requirements

- [x] `graphQL` mutation is implemented and works as desired
- [x] Error toasts are shown on failure with proper localization in the corresponding component

NabbeunNabi:
Resolves #796 #531 

## 🔧 What changed
1. Update button moved to top bar
2. Update function uses new API call
3. Implement fuzzy search to add HPs that we have to our db (This does not hinder performance since it uses filter with less than 1000 entries and we are unlikely to have that many for a while) Without this there is no way to add an existing HP as we do not have any way to easily search them unless you know the exact id.
4. Added update functionality with a function to check for changes to make sure there were changes before sending the request to the API
5. Add HP card that is used in this component and can be used for the submission with styling discussed with a designer for best practices
6. Added gray to our theme colors due to it passing the standards of [Web Content Accessibility Guidelines](https://www.w3.org/WAI/standards-guidelines/wcag/)

## 🧪 Testing instructions
Tests are being added in #538 

## 📸 Screenshots

-   ### Before
N/A
-   ### After

https://github.com/user-attachments/assets/9ad528eb-0320-461f-9b1c-b138335231c0


components/ModEditFacilityOrProfessionalTopbar.vue
```diff
@@ -93,7 +111,79 @@ function setSelectedId() {
...
+    return areThereUnsavedChanges
+}
+
+const updateFacilityOrHealthcareProfessional = async () => {
+    // This makes the on click update the facility if the screen is EditFacility
+    if (moderationScreenStore.activeScreen === ModerationScreen.EditFacility) {
+        const checkForUnsavedChanges = facilityHasUnsavedChanges()
+
+        // This prevents us from sending a requested unnecessarily if the user has not made changes
+        if (!checkForUnsavedChanges) return
```
ermish, line 156: we should disable this button during the action.

components/ModEditFacilityOrProfessionalTopbar.vue
```diff
@@ -93,7 +111,79 @@ function setSelectedId() {
...
+
+        const response = await facilitiesStore.updateFacility()
+
+        if (response.errors?.length) {
+            handleServerErrorMessaging(response.errors, toast, t)
+            return response
+        }
+
+        facilitiesStore.initializeFacilitySectionValues(response.data as Facility)
```
ermish, line 165: let's add a comment to explain what you're doing with this response data.

components/ModEditFacilityOrProfessionalTopbar.vue
```diff
@@ -93,7 +111,79 @@ function setSelectedId() {
...
+
+const updateFacilityOrHealthcareProfessional = async () => {
+    // This makes the on click update the facility if the screen is EditFacility
+    if (moderationScreenStore.activeScreen === ModerationScreen.EditFacility) {
+        const checkForUnsavedChanges = facilityHasUnsavedChanges()
+
+        // This prevents us from sending a requested unnecessarily if the user has not made changes
+        if (!checkForUnsavedChanges) return
+
+        const response = await facilitiesStore.updateFacility()
```
ermish, line 158: should we add a loader here? 
Could be in a future issue.

components/ModEditFacilityOrProfessionalTopbar.vue
```diff
@@ -93,7 +111,79 @@ function setSelectedId() {
...
+        toast.success(t('modEditFacilityOrHPTopbar.facilityUpdatedSuccessfully'))
+        return response
+    }
+}
+
+const updateFacilityOrHealthcareProfessionalAndExit = async () => {
+    const response = await updateFacilityOrHealthcareProfessional()
+
+    if (response && response.errors?.length) {
+        return
```
ermish, line 175: silent exception! let's at least log to console. eventually this should be a log

components/ModEditFacilitySection.vue
```diff
@@ -225,22 +225,55 @@
...
+            <span
+                v-if="moderationScreenStore.activeScreen === ModerationScreen.EditFacility"
+                class="mb-1 text-primary-text text-2xl font-bold font-sans leading-normal"
+            >
+                {{ $t('modFacilitySection.healthcareProfessionalToAdd') }}
+            </span>
+            <span
+                v-show="!facilityStore.healthProfessionalsRelationsForDisplay"
+                class="font-semibold"
+            >- {{ $t('modFacilitySection.noHPSelected') }}
```
ermish, line 238: this `-` formatting look right?

components/ModEditFacilitySection.vue
```diff
@@ -253,16 +286,30 @@ import { validateAddressLineEn,
...
 const facilityStore = useFacilitiesStore()
+const healthcareProfessionalsStore = useHealthcareProfessionalsStore()
 
 const isFacilitySectionInitialized: Ref<boolean> = ref(false)
+const healthcareProfessionalsRelatedToFacility: Ref<string[]>
+= ref([])
+const healthcareProfessionalRelatedToFacilityFiltered: Ref<HealthcareProfessional[]> = ref([])
+
+const setHealthcareProfessionalsRelatedToFacility = () => {
+    healthcareProfessionalRelatedToFacilityFiltered.value = healthcareProfessionalsRelatedToFacility.value.flatMap(
```
ermish, line 306: first time I've seen flatMap actually used! ....I don't think we need it though. This can just be `map()` I think?

components/ModEditFacilitySection.vue
```diff
@@ -304,18 +351,54 @@ onBeforeMount(async () => {
         await facilityStore.getFacilities()
     }
 
+    if (!healthcareProfessionalsStore.healthcareProfessionalsData) {
```
ermish, line 354: good comment material

components/ModEditFacilitySection.vue
```diff
@@ -304,18 +351,54 @@ onBeforeMount(async () => {
...
 
     // Set the active screen and ensure the UI state is consistent
-    screenStore.setActiveScreen(ModerationScreen.EditFacility)
+    moderationScreenStore.setActiveScreen(ModerationScreen.EditFacility)
 
     facilityStore.setSelectedFacilityData(facilityStore.selectedFacilityId)
 
     facilityStore.initializeFacilitySectionValues(facilityStore.selectedFacilityData)
 
+    await nextTick()
```
ermish, line 367: good comment material

components/ModEditFacilitySection.vue
```diff
@@ -304,18 +351,54 @@ onBeforeMount(async () => {
...
+    if (healthcareProfessionalsStore.healthcareProfessionalsData
```
ermish, line 368: I think this can be merged with the map function to just be a one line assignment 
like `healthcareProfessionalsStore.healthcareProfessionalsData?.map(...)`

ermish: ah, I see this method is reused. I would move this check ^ into the method.

Perhaps `sync` instead of `set` would be a better prefix for the method?

components/ModHealthcareProfessionalSearchbar.vue
```diff
@@ -9,34 +9,138 @@
...
+const moderationScreenStore = useModerationScreenStore()
+const facilitiesStore = useFacilitiesStore()
 
 const requestedHealthcareProfessionalId: Ref<string> = ref('')
 // `healthcareProfessionalById` will be used in the UI component in an upcoming PR
-const healthcareProfessionalById: Ref<HealthcareProfessional[] | []> = ref([])
+const healthcareProfessionalByIdOrName: Ref<HealthcareProfessional[] | []> = ref([])
+
+const addHealthcareProfessional = (id: string) => {
+    if (moderationScreenStore.activeScreen === ModerationScreen.EditFacility) {
```
ermish, line 81: nice clear method! I know this logic is complex.

components/ModHealthcareProfessionalSearchbar.vue
```diff
@@ -9,34 +9,138 @@
...
+    const healthcareProfessionalByIdOrNamesFound = healthcareProfessionalsStore.healthcareProfessionalsData
+        .filter(healthcareProfessional => {
+            const caseInsensitiveIdMatch = healthcareProfessional.id
+                .toLowerCase()
+                .startsWith(requestedHealthcareProfessionalId.value.toLowerCase().trim())
+
+            const nameMatches = healthcareProfessional.names.some(name => {
+                const firstNameMatch = name.firstName
+                    .toLowerCase()
+                    .includes(requestedHealthcareProfessionalId.value.toLowerCase())
```
ermish, line 125: it looks like you're comparing the wrong value for the names here. It's checking name matching the id.
.....and as I'm writing this I realized your variable is just id but it's also a name input. Let's change the variable name to be something like `searchValue`

stores/facilitiesStore.ts
```diff
@@ -1,38 +1,45 @@
...
+import type { ServerError, ServerResponse } from '~/typedefs/serverResponse'
 
 export const useFacilitiesStore = defineStore(
     'facilitiesStore',
     () => {
         const facilityData: Ref<Facility[]> = ref([])
         const selectedFacilityId: Ref<string> = ref('')
         const selectedFacilityData: Ref<Facility | undefined> = ref()
-        const facilitySectionFields = {
+        const facilitySectionFields = reactive({
```
ermish, line 17: nice! much better than an object of refs.

stores/facilitiesStore.ts
```diff
@@ -42,41 +49,81 @@ export const useFacilitiesStore = defineStore(
...
+            }
+
+            const response = await graphQLClientRequestWithRetry<Mutation>(
+                gqlClient.request.bind(gqlClient),
+                updateExistingFacilityGqlMutation,
+                updateFacilityInput
+            )
+
+            serverResponse.data = response.data?.updateFacility
+            serverResponse.errors = response.errors ? response.errors : []
```
ermish, line 118: should add a log at least if there's errors

tailwind.config.js
```diff
@@ -30,6 +30,7 @@ module.exports = {
                 'rgb(var(--color-primary-inverted) / <alpha-value>)',
             'primary-hover': 'rgb(var(--color-primary-hover) / <alpha-value>)',
             'primary-bg': 'rgb(var(--color-bg-primary) / <alpha-value>)',
+            'primary-bg-gray': 'rgb(var(--color-primary-gray) / <alpha-value>)',
```
ermish, line 33: ♻️

ermish (changes requested):
reviewed!

components/ModDashboardHealthProfessionalCard.vue
```diff
@@ -0,0 +1,137 @@
...
+                    role="img"
+                    alt="profile icon"
+                    title="profile icon"
+                    class="profile-icon stroke-primary w-16 h-16 stroke-1 inline mx-1 self-start"
+                />
+            </div>
+            <div v-if="moderationScreenStore.activeScreen === ModerationScreen.EditFacility">
+                <div class="flex flex-col h-full w-64 pl-1 mb-1">
+                    <div class="flex font-bold pt-2">
+                        <span>{{ healthcareProfessional.names[0].lastName }}</span>
```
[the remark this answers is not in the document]
NabbeunNabi: I did it where it defaults to 0 index since I believe at this current moment we don't require an English name. It can be localized to any language when adding to the database

stores/facilitiesStore.ts
NabbeunNabi: A log is recorded here

```
export const graphQLClientRequestWithRetry = async <T>(
    gqlClientRequestFunction: (
        queryOrMutation: RequestDocument,
        variables?: unknown
    ) => Promise<T>,
    queryOrMutation: RequestDocument,
    variables: unknown,
    retryOptions?: graphQLClientRequestWithRetryOptions
): Promise<ServerResponse<T>> => {
    let attempts = 0
    const retryAmount = retryOptions?.retryAmount || 3
    const requestTimeoutInMilliseconds = retryOptions?.requestTimeoutInMilliseconds || 5000

    const executeGQLClientRequest = async (): Promise<ServerResponse<T>> => {
        try {
            const data = await gqlClientRequestFunction(queryOrMutation, variables)
            return { data, errors: [], hasErrors: false }
        } catch (error) {
            if (attempts < retryAmount) {
                attempts++
                if (attempts > 1) {
                    await new Promise(resolve => setTimeout(resolve, requestTimeoutInMilliseconds))
                }
                return executeGQLClientRequest() // retry request
            }

            // This is a consistent error messaging no matter the type of query or mutation
            console.error(`There was an error executing the request: ${error}`)
            const serverError = error as ServerErrorResponse

            // This map transforms errors if they exist
            const errors = serverError.response?.errors?.map(errorResponse => ({
                message: errorResponse.message,
                fieldWithError: errorResponse.locations,
                code: errorResponse.extensions.code as ErrorCode
            })) || []

            return { data: {} as T, errors, hasErrors: true }
        }
    }

    return executeGQLClientRequest()
}
```

It happens if there is every an error and there are user toast messages displayed for the specific error

components/ModEditFacilityOrProfessionalTopbar.vue
NabbeunNabi: Yea. Could add a loader. I am not sure of the speed of it at the moment. Of course locally it runs quickly. Maybe a future issue as this one is big enough already

components/ModEditFacilityOrProfessionalTopbar.vue
NabbeunNabi: there is always an error logged to the console with the api call. I added the toast messaging that should have been added

components/ModEditFacilitySection.vue
NabbeunNabi: This is just a hardcoded - to show the text. It doesn't need to be translated

components/ModEditFacilitySection.vue
NabbeunNabi: Correct. I used to make sure it was the correct type but if I use `as HealthcareProfessional` it works as a map

NabbeunNabi: If I do not use flatMap it doesn't remove any undefined values. This is to safely remove anything that is undefined. Otherwise the website could break on reload

NabbeunNabi:
@ermish The reactive object is in the store due to it being accessed in multiple components. The api form that validates and sends the updated store values for the facility value are in the `topbar`. The values for display are in the `ModEditFacilitySection`
The updated refs to be sent to the backend come from the searchbar. I do not know if this should be consolidated. But for now maybe it is better?

ermish (approved):
ship it!

PR #911, opened 2024-11-20
fix: remove bubbling from synthetic event

### LINKED ISSUE #912: Remove event bubbling from synthetic event in `ModEditSubmissionForm.vue`
# Problem
The event bubbling is causing the frontend validation to fail for healthcare professionals as it deselects the multiselect fields on click. This needs to be removed so the validation doesn't fail as the submission is being successfully updated in the backend and should be able to be approved right after all fields are updated.

# Success
Event bubbling is removed and the multiselect stays selected and the user flow isn't interrupted when trying to approve after updating the whole submission.
### Requirements

- [x] Event bubbling on the synthetic event is removed from the `ModEditSubmissionForm.vue`
- [x] Flow isn't interrupted due to the removal of event bubbling

NabbeunNabi:
Resolves #912

## 🔧 What changed
- The synthetic event was bubbling up causing the multiselect fields to be deselected. This made the frontend validation fail when the submission had been updated correctly.

## 🧪 Testing instructions
Tests are currently being written #538

### DIFF
--- a/components/ModEditSubmissionForm.vue
+++ b/components/ModEditSubmissionForm.vue
@@ -863,7 +863,7 @@ async function submitCompletedForm(e: Event) {
     }
 }
 
-const syntheticEvent = new Event('submit', { bubbles: true, cancelable: true })
+const syntheticEvent = new Event('submit', { bubbles: false, cancelable: true })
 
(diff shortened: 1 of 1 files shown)


ermish (approved):
very eventful

PR #922, opened 2024-11-26
fix: change watch to selected submission Id to avoid unwanted UI updates

### LINKED ISSUE #921: Fix the edit submission form to initialize form data by watching `selectedId` instead of whole store
# Problem
The user interface was a bit unreliable since it would sometimes refresh the healthcare professional form data that was initialized after watching the whole store. We do not want this to happen. We only want this to happen if the ref in the store `selectedSubmissionId` changes. 

# Success
The watch is changed to watch the singular value of `selectedSubmissionId` instead of the whole store. This then updates the form with the new values if this changes due to someone being sent to this form with a link.

### Requirements

- [x] The watch only watches the `selectedSubmissionId` and updates the form in the case that this variable in the store is updated

NabbeunNabi:
Resolves #921 

## 🔧 What changed
The watch is changed to watch the singular value of selectedSubmissionId instead of the whole store. This then updates the form with the new values if this changes due to someone being sent to this form with a link instead of anything in the store changing. If it watches the whole store it can have unreliable ui updates that prevent the acceptance of a submission after an update to that submission

## 🧪 Testing instructions
This bug was found in the creation and implementation of #538.

### DIFF
--- a/components/ModEditSubmissionForm.vue
+++ b/components/ModEditSubmissionForm.vue
@@ -872,9 +872,9 @@ watch(moderationSubmissionStore, newValue => {
     }
 })
 
-watch(moderationSubmissionStore, newValue => {
-    moderationSubmissionStore.filterSelectedSubmission(newValue.selectedSubmissionId)
-    initializeSubmissionFormValues(newValue.selectedSubmissionData)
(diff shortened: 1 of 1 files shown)


ermish (approved):
Easy improvement!

PR #923, opened 2024-11-27
fix: implemented changes to display Search bar only on the index page

### LINKED ISSUE #913: Show the search bar only on the home page
# Problem
Right now if we are on pages that are not the home page where you can search. We can still see the search bar. We only want it to be displayed on the index page.

Not here or here
Image
Image

# Success
The search bar is only shown on the index pages
### Requirements

- [ ] Search bar is shown and functional on the index page
- [ ] Search bar is not visible on the about us page or add a doctor page

# Implementation notes
You could use the `v-show` directive in Vue and `useRoute` to show based on a check of the path

karlk08:
Resolves #913 

## 🔧 What changed
1. Display the Search bar  on the index page only.
2. Make sure the Search bar is hidden on the about us page or add a doctor page.

## 🧪 Testing instructions

N/A

## 📸 Screenshots

-   ### Before
Screenshot 2024-11-27 at 18 25 16
Screenshot 2024-11-27 at 18 25 28

-   ### After

Screenshot 2024-11-27 at 18 38 35
Screenshot 2024-11-27 at 18 38 22

### DIFF
--- a/components/SearchBar.vue
+++ b/components/SearchBar.vue
@@ -1,7 +1,7 @@
 <template>
     <div class="flex flex-0 max-w-4xl">
         <div
-            v-show="route.path === '/'"
+            v-show="isHomepage"
             id="search-fields"
             class="grid-cols-3 mx-4"
         >
@@ -78,7 +78,7 @@
             </div>
         </div>
         <div
-            v-show="route.path === '/'"
+            v-show="isHomepage"
             id="search-button"
             class="flex items-center"
         >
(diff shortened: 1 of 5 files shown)


NabbeunNabi (approved):
💯 Great job changing from hidden to `v-show`

NabbeunNabi (commented):
Do not forget to update the tests based on this behavior @karlk08

NabbeunNabi (changes requested):
Do not forget to update the tests based on this behavior @karlk08

ermish (changes requested):
much appreciate the bug fix! I think we should fix this in the parent component/layout that is using the searchbar. 
A component shouldn't be aware of where it lives in the page

karlk08:
Sorry if it took a while. I reverted the changes on the child component SearchBar.vue and instead implemented it to the parent component TopNav.vue as requested.

ermish (approved):
thanks for the fix!

ermish:
hey @karlk08 it looks like the test is failing like @NabbeunNabi was referring to. Can you take a look?

Issue #924, opened 2024-11-29
Create a facility search bar for existing facilities


NabbeunNabi:
# Problem
We can search for healthcare professionals to add them to existing facilities. But in this project we also can have a relationship where we add a facility to an existing healthcare professional in case they move or start at another hospital. We currently do not have this component.

# Success
A facility search bar component is made with similar styling to the `healthcareProfessionalSearchBar`.
### Requirements

- [x] The styling for the search bar for facilities is made along with the template. 
- [x] The placeholders are written using translatable strings and added to the `locales` we have. You can use `yarn lint:locales` if you have already added the English
- [x] The functionality is written in which the searched facility by ID or Name is shown with styling. Similar styling has been done in PR #902 

# Implementation notes
Testing is on hold while @ermish and the team discuss whether it is better to change to a monorepo and spin up a docker instance for the testing envrionment going forward.

gminetoma:
Waiting for #855

PR #925, opened 2024-11-30
feat: add facility search bar component

gminetoma:
Resolves [#924](https://github.com/ourjapanlife/findadoc-web/issues/924)

## 🔧 What changed

Introduces a new search component to search for existing facilities.
Handles input changes, searches through `facilitiesStore.facilityData`,
and displays the best matches. Users can select a facility using
the mouse or arrow keys + Enter. Pressing the Esc key or blurring
the search input clears the input field.

Uses a store to keep the results, accessible via 
`useFacilitySearchbarStore().selectedFacilities`.

Adds a `.gitignore` rule for VS Code settings: `.vscode`.

## 📸 Screenshots

-   ### Before
- N/A

-   ### After

https://github.com/user-attachments/assets/e8c39169-94dd-418c-9f1b-3b81364c0989


components/ModFacilitySearchbar.vue
```diff
@@ -0,0 +1,181 @@
+<template>
+    <div
+        class="flex flex-col gap-1 w-fit"
+    >
+        <div
+            class="flex items-center px-3 py-3.5 w-96 h-12 bg-secondary-bg
+            rounded-lg border border-primary-text-muted text-primary-text
+            text-sm font-normal font-sans placeholder-primary-text-muted
+            outline outline-0 -outline-offset-2 gap-2 divide-x-2
```
ermish, line 9: there's an extra `-` in this line

components/ModFacilitySearchbar.vue
```diff
@@ -0,0 +1,181 @@
...
+                @click="searchInputElement?.focus()"
+            >
+                <SVGLookingGlass
+                    role="img"
+                    title="searching icon"
+                    class="h-6 pl-2"
+                />
+            </button>
+        </div>
+        <div class="relative">
```
ermish, line 38: this looks like a good reusable component. Let's split this out.

components/ModFacilitySearchbar.vue
```diff
@@ -0,0 +1,181 @@
...
+                    :class="[
+                        selectedFacilities.has(facility) && 'bg-primary-text-muted',
+                        selectedFacilityIndex === index ? 'bg-primary-hover text-primary-inverted' : 'opacity-95',
+                    ]"
+                    @click="(event) => { event.preventDefault(); handleListItemClick(facility) }"
+                    @mousedown="(event) => {
+                        // Prevent search input from being blurred
+                        event.preventDefault()
+                    }"
+                    @mousemove="selectedFacilityIndex = index"
```
ermish, line 60: mousemove is a heavy event handler as it's triggered for every mouse pixel movement. Can we just use hover as a trigger here to set the id?

components/ModFacilitySearchbar.vue
```diff
@@ -0,0 +1,181 @@
...
+                "
+                :class="computedDropdownClass"
+            >
+                <!-- Search limited to 10 items -->
+                <li
+                    v-for="(facility, index) in filteredFacilities.slice(0, 10)"
+                    :key="facility.id"
+                    class="flex justify-between divide-x cursor-pointer"
+                    :class="[
+                        selectedFacilities.has(facility) && 'bg-primary-text-muted',
```
ermish, line 52: UX: an active selected facility seems a bit strange as gray. How about we do a faded primary instead? something like `bg-primary/70`

components/ModFacilitySearchbar.vue
```diff
@@ -0,0 +1,181 @@
...
+                                selectedFacilities.has(facility) ? 'opacity-100' : 'opacity-10',
+                            ]"
+                        />
+                    </div>
+                </li>
+                <li
+                    v-if="!filteredFacilities.length"
+                    class="m-3 cursor-default"
+                >
+                    <!-- TODO: Change i18n key if necessary -->
```
ermish, line 79: can remove this comment now I believe

components/ModFacilitySearchbar.vue
```diff
@@ -0,0 +1,181 @@
...
+                    <div class="flex items-center">
+                        <SVGCheckMark
+                            class="h-4 m-3"
+                            :class="[
+                                selectedFacilities.has(facility) ? 'opacity-100' : 'opacity-10',
+                            ]"
+                        />
+                    </div>
+                </li>
+                <li
```
ermish, line 75: let's add a some names for these divs, li, and ul to explain what each is

components/ModFacilitySearchbar.vue
```diff
@@ -0,0 +1,181 @@
...
+const facilitySearchbarStore = useFacilitySearchbarStore()
+
+const searchInput = ref('')
+const searchInputElement = ref<HTMLInputElement | null>(null)
+const filteredFacilities = ref<Facility[]>([])
+const selectedFacilities = ref(facilitySearchbarStore.selectedFacilities)
+const selectedFacilityIndex = ref(0)
+
+const handleListItemClick = (facility: Facility) => {
+    searchInputElement.value?.focus()
```
ermish, line 106: let's add a comment here to explain why were focusing here. To that point, I'm not sure why it's focused 🤔

components/ModFacilitySearchbar.vue
```diff
@@ -0,0 +1,181 @@
...
+    searchInput.value = ''
+}
+
+const handleSearchInputEsc = () => {
+    handleSearchInputBlur()
+}
+
+const handleSearchInputArrowUp = (event: KeyboardEvent) => {
+    // Prevent the cursor to be moving inside the search input
+    event.preventDefault()
```
ermish, line 126: I noticed you have prevent default happening here but then also in some of the html event hooks. Let's make it more consistent and move them all to the event handlers like you have here.

components/ModFacilitySearchbar.vue
```diff
@@ -0,0 +1,181 @@
...
+    if (selectedFacilityIndex.value > -1) selectedFacilityIndex.value -= 1
```
ermish, line 127: shouldn't this be `> 0`?

components/ModFacilitySearchbar.vue
```diff
@@ -0,0 +1,181 @@
...
+        const isMatch
+        = nameEn.toLowerCase().includes(facilityIdOrName)
+        || nameJa.toLowerCase().includes(facilityIdOrName)
+        // TODO: Is the id case sensitive?
+        || id.toLowerCase() === facilityIdOrName
+        return isMatch
+    })
+}
+
+// TODO: Clear the store to prevent bugs. This might change in the future.
```
ermish, line 160: can you make a github issue for this todo?

components/ModFacilitySearchbar.vue
```diff
@@ -0,0 +1,181 @@
...
+    selectedFacilities.value = new Set()
+})
+
+watch(searchInput, handleSearchInputChange)
+
+// Displays the dropdown above the searchInputElement if its position is greater than window.innerHeight / 2
+const computedDropdownClass = computed(
+    () => {
+        if (searchInputElement.value && searchInputElement.value.getBoundingClientRect().bottom >= window.innerHeight / 2) {
+            return '-translate-y-full -top-14'
```
ermish, line 176: Is this going to work for mobile? instead of `top-14` can we use something that's responsive?

stores/facilitySearchbarStore.ts
```diff
@@ -0,0 +1,11 @@
+import { defineStore } from 'pinia'
+import { ref } from 'vue'
+import type { Facility } from '~/typedefs/gqlTypes'
+
+const initializeStore = () => {
```
ermish, line 5: does this actually need to be a store instead of just local state within the searchbar component? I think local state should be fine here 🤔

ermish (changes requested):
lots of great work in here! Complex dropdowns have a lot of UX and logic. Great work on making one for findadoc 🚀

components/ModFacilitySearchbar.vue
gminetoma: -outline-offset-2
image

outline-offset-2
image

Should I remove it?

components/ModFacilitySearchbar.vue
gminetoma: Ops... This is not an issue! I'll remove this comment. 😊

components/ModFacilitySearchbar.vue
gminetoma: This is a leftover. Removing it! 😊

gminetoma:
Requested changes:

- [ ] Remove the extra "-" from line 9.
- [ ] Extract the search input component.
- [ ] Replace specific class colors like "bg-white" with theme colors.
- [ ] Improve the display of search results.
- [ ] Replace "mousemove" with "mouseover".
- [ ] Change the active item color to a better one.
- [ ] Remove unnecessary comments.
- [ ] Improve div identification.
- [ ] Place all "event.preventDefault()" calls inside their respective handlers.
- [ ] Set the index limit for the arrow-up key to 0.
- [ ] Use a responsive alternative for '-translate-y-full -top-14'.
- [ ] Remove the store and use local state for search results.

After changes:

https://github.com/user-attachments/assets/931b510b-e633-4005-9280-c69291687873

components/ModFacilitySearchbar.vue
```diff
@@ -0,0 +1,178 @@
...
+const selectedItems = defineModel<Set<Facility>>({ required: true })
+
+//The limit for fetching facilities is 400 at the current date (30/11/2024)
+const facilitiesStore = useFacilitiesStore()
+const searchInputElement = ref<HTMLInputElement>()
+const searchInputValue = ref('')
+const filteredItems = ref<Facility[]>([])
+const selectedItemIndex = ref(0)
+
+const handleListScroll = () => {
```
ermish, line 86: nice! looks like you got the component separated! 
I think you can also move the list and the handlers into the component as well. It's good that the events are emitted and we should keep that in case the caller wants to add additional logic on the events. For most of these handlers, it looks like we're updating the selectedItems array which is likely logic we'll want to use for any v-model we pass in. (so we should just move the logic into the component) 

The only one that looks unique to facility is the handleSearchInputChange() which looks appropriate to keep here

components/ModFacilitySearchbar.vue
ermish: ohh sorry, this is correct for a negative value! I didn't realize this was a feature for negative values! (https://tailwindcss.com/docs/upgrade-guide#negative-values-removed)

components/ModFacilitySearchbar.vue
```diff
@@ -0,0 +1,178 @@
+<template>
+    <div
+        class="container flex gap-1 w-fit"
+        :class="[
+            isNearPageBottom ? 'flex-col-reverse' : 'flex-col',
+        ]"
+    >
+        <ModSearchBar
+            v-model="searchInputElement"
+            :items-count="filteredItems.length"
```
ermish, line 10: nit: could rename to `search-results-count`

components/ModEditHealthcareProfessionalSection.vue
```diff
@@ -202,18 +202,37 @@
...
                 </option>
             </select>
+            <h2
+                class="mod-healthcare-professional-section
+                 my-3.5 text-start text-primary-text text-2xl font-bold font-sans leading-normal"
+            >
+                {{ $t("modHealthcareProfessionalSection.facilities") }}
+            </h2>
+            <!-- This is just an example how to use it -->
+            <ModFacilitySearchbar v-model="selectedFacitilies" />
```
ermish, line 212: just realized this is misspelled. 
should be `selectedFacilities` not `facitilies`

components/ModFacilitySearchbar.vue
ermish: it's better now, but still don't love it's gray, but I think this is a theme issue and you're code looks fine. 👍

ermish:
nice updates @gminetoma ! Added 3 new comments and the rest look fixed!

components/ModFacilitySearchbar.vue
gminetoma: I don't think we can move the list to the input component. We can't make a generic component from the list, since (I think) not all types of search will have an ID, Japanese name, and English name. For example, if we want to search a Submission by ID. Additionally, 'HealthcareProfessional' does not have the same properties as 'Facility', such as 'nameEn' and 'nameJa'.

gminetoma: I was wrong 🤭. I was able to make it generic after all!

gminetoma:
@ermish -> [commit](https://github.com/ourjapanlife/findadoc-web/pull/925/commits/282b3653fadcff01600d55078c05982bba1785f1)

After changes:

Health Care Professionals:

https://github.com/user-attachments/assets/407e8c7a-0984-4891-8231-61e022d72171

Facilities:

https://github.com/user-attachments/assets/c488e811-68de-4abe-9cd6-3e52257b75bf

ermish (approved):
looks  solid! nice work with the changes @gminetoma ! 🤜 🤛

ermish:
@gminetoma FYI I added missing Vietnamese locale keys that were added from another PR merge so this one can be merged too 🚀

PR #952, opened 2024-12-24
feat: add lazy load to i18n

### LINKED ISSUE #889: FEAT: Lazy load localization
# Problem
Today, all of our localization is downloaded when someone visits the site. We should only load the current language the user is  using. 

image

- [x] The user doesn't have to "noticeably" wait for languages to load
- [x] The user can change a language without a page refresh
- [x] The user's default language is based on their browser's default language. (we can default to english first, then load theirs)

### LINKED ISSUE #1022: bug: Fix message in logs
# Problem
When running the app locally, we're getting this repeated error. 

Image

For debugging, I was running the cypress tests locally which could also be a culprit

# Success
It's gone!

### Requirements

- [ ] Fix it!

gminetoma:
Resolves #889
Resolves #1022 

## 🔧 What changed
We no longer need the file './i18n.options.ts' to configure i18n.

The configuration is now handled directly in the Nuxt config file.

We are loading the last language used via cookies; therefore, we need to disable the language selector and display a loading indicator for the user until the loading process is complete.

The browser's language code uses "-", so we need to apply a small adjustment to align with our backend typing, which uses "_".

## Notes
This PR initiates a discussion on refactoring/updating our backend. See it here: [GitBook](https://app.gitbook.com/o/Hmir5Cugknp7uJaXBpz1/s/znzEKbLPVwI181bff5yl/update-locale-enum-on-backend)

## 📑  References
[Nuxt i18n Lazy-load translations](https://i18n.nuxtjs.org/docs/guide/lazy-load-translations)
[Nuxt Vue I18n API](https://i18n.nuxtjs.org/docs/v8/api/vue-i18n)

## 📸 Screenshots (#889)

-   ### Before
https://github.com/user-attachments/assets/261c2340-12b4-4437-9624-e2833bcd4768

-   ### After
https://github.com/user-attachments/assets/f2db64fd-cf85-4a89-b4c8-092f357bff5e

## 📸 Screenshots (#1022)

-   ### Before
image

-   ### After
image

### DIFF
--- a/components/LocaleSelector.vue
+++ b/components/LocaleSelector.vue
@@ -18,17 +18,26 @@
 </template>
 
 <script setup lang="ts">
-import { ref, watch } from 'vue'
+// Nuxt vue-i18n documentation:
+// https://i18n.nuxtjs.org/docs/v8/api/vue-i18n
+
+import { onMounted, ref, watch } from 'vue'
 import { useLocaleStore } from '../stores/localeStore'
-import { Locale } from '~/typedefs/gqlTypes.js'
+import type { Locale } from '#i18n'
+import { Locale as GqlLocale } from '~/typedefs/gqlTypes.js'
 import { useI18n } from '#imports'
 
-const { locale } = useI18n()
+const { setLocale, getLocaleCookie } = useI18n()
 const localeStore = useLocaleStore()
-const selectedLocale = ref(Locale.EnUs)
 
-watch(selectedLocale, (newLocale: Locale) => {
-    localeStore.setLocale(newLocale)
-    locale.value = newLocale
+// getLocaleCookie to set the initial locale, if there is no cookie, it will default based on the nuxt.config.js
+// Browser's code language is using "-" instead of "_".
+const localeCookie = getLocaleCookie()?.replace('-', '_') || GqlLocale.EnUs
+const selectedLocale = ref(localeCookie)
+
+watch(selectedLocale, newLocale => {
+    localeStore.setLocale(newLocale as GqlLocale)
+    // setLocale to update the i18n plugin and set the 'i18n_redirected' cookie
+    setLocale(newLocale.replace('_', '-') as Locale)
 })
 </script>
--- a/components/LocaleSelector.vue
+++ b/components/LocaleSelector.vue
@@ -21,11 +21,8 @@
 // Nuxt vue-i18n documentation:
 // https://i18n.nuxtjs.org/docs/v8/api/vue-i18n
 
-import { onMounted, ref, watch } from 'vue'
-import { useLocaleStore } from '../stores/localeStore'
 import type { Locale } from '#i18n'
 import { Locale as GqlLocale } from '~/typedefs/gqlTypes.js'
-import { useI18n } from '#imports'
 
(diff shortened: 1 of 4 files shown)


components/LocaleSelector.vue
```diff
@@ -5,30 +5,48 @@
...
+const isReady = ref(false)
+
+//@ts-expect-error - This is a workaround for the fact that the i18n plugin is not typed
+const { setLocale, getLocaleCookie } = useI18n()
 const localeStore = useLocaleStore()
-const selectedLocale = ref(Locale.EnUs)
+
+// getLocaleCookie to set the initial locale, if there is no cookie, it will default based on the nuxt.config.js
+// HACK: Browser's code language is using "-" instead of "_".
+const localeCookie = getLocaleCookie() ? getLocaleCookie().replace('-', '_') : Locale.EnUs
```
ermish, line 40: if having the dash in there is hacky, could we just remove it? I don't think we technically need it, it was just there to try and match a standard. This would simplify the code and remove the hack 😸

i18n/index.ts
```diff
@@ -1,30 +1,53 @@
...
-    //eslint-disable-next-line
-    de_DE: de,
-    //eslint-disable-next-line
-    zh_CN: cn,
-    //eslint-disable-next-line
-    fr_FR: fr,
-    //eslint-disable-next-line
-    tl_PH: tl
+// HACK: Browser's code language is using "-" instead of "_".
+const Locale: Partial<CustomLocale> = {
```
ermish, line 6: can you add a comment here and on customLocale to explain what the partial and customLocale is for? I think I understand, but other devs might not.

ermish (changes requested):
nice work @gminetoma ! good find for lazy loading them and thinking of the loading state ✨

components/LocaleSelector.vue
gminetoma: Sorry @ermish, we cannot remove it because our location is based on the backend's data type 'Locale'. The backend uses '_' while the browser and Nuxt use '-' for locale. We are saving the cookie in the browser's format and then parsing it to our format. We could fix this by changing the server's data type.

i18n/index.ts
gminetoma: done! 🫡

components/LocaleSelector.vue
ermish: Hmmm I think we should change the servers type but let's get democracy in here! @Anissa3005  @theyokohamalife @NabbeunNabi cast ye's opinion 🏴‍☠️

NabbeunNabi: I think changing the server's type would also be better as an overall choice. Makes sense to me to match the browser, since most modern browsers use IETF BCP 47 language tags.

theyokohamalife: I'm in favor of changing it in the backend. @gminetoma it would be a good idea to include a link to [the Nuxt docs](https://i18n.nuxtjs.org/docs/guide/lazy-load-translations) in your PR description so reviewers can reference it easily.

ermish:
@gminetoma  moving this to Draft for now until the backend is updated 👍

ermish (approved):
Looks great!

PR #953, opened 2024-12-27
feat: autofill the healthcare professional info in mod

### LINKED ISSUE #864: Auto fill the Healthcare Professional Section of the form in the moderation panel
# Problem
The UI is created in #784 and needs to be auto fill with the selected Healthcare professional when the `ModerationScreen.EditHealthcareProfessional` screen is selected in the dashboard.
# Success
The form is auto filled
### Requirements

- [ ] Autofill the healthcare professional section with the data of the selected healthcare professional.

ShizuokaTerrier:
Resolves #864 

## 🔧 What changed

The selected healthcare professional's information is now autofilled. 

If the professional has name info in another locale that also autofills when the locale is selected. 

## 📸 Screenshots

-   ### Before
image

-   ### After

https://github.com/user-attachments/assets/27b7b571-da7e-4e78-9eeb-22fb18eeb82a

### DIFF
--- a/components/ModDashboardHealthProfessionalCard.vue
+++ b/components/ModDashboardHealthProfessionalCard.vue
@@ -11,7 +11,7 @@
                     class="profile-icon stroke-primary w-16 h-16 stroke-1 inline mx-1 self-start"
                 />
             </div>
-            <div v-show="healthcareProfessionalsRelatedToFacility">
+            <div v-if="healthcareProfessionalsRelatedToFacility">
                 <div class="flex flex-col h-full w-64 pl-1 mb-1">
                     <div class="flex font-bold pt-2">
                         <span>{{ healthcareProfessionalsStore
@@ -52,7 +52,7 @@
                 </div>
             </div>
             <div
-                v-show="!healthcareProfessionalsRelatedToFacility"
+                v-if="!healthcareProfessionalsRelatedToFacility"
             >
                 <div
                     v-for="(nameLocale, index) in healthcareProfessional.names"
@@ -72,6 +72,24 @@
                         {{ localeStore.formatLanguageCodeToSimpleText(
                             nameLocale.locale) }}
                     </span>
+                    <div>
+                        <button
+                            v-if="!isEditable && setOnClick"
+                            type="button"
+                            class="border-2 border-primary px-3 my-1 rounded-full font-semibold hover:bg-currentColor"
+                            @click="() => setChangeToEditable(nameLocale.locale)"
+                        >
+                            {{ $t("modHealthcareProfessionalCard.editName") }}
+                        </button>
+                        <button
+                            v-if="isEditable && updateOnClick"
+                            type="button"
+                            class="border-2 border-primary px-3 my-1 rounded-full font-semibold hover:bg-currentColor"
+                            @click="() => setToUneditableAndSave()"
+                        >
+                            {{ $t("modHealthcareProfessionalCard.saveName") }}
+                        </button>
+                    </div>
                 </div>
             </div>
             <div
@@ -99,6 +117,7 @@
 </template>
 
 <script setup lang="ts">
+import { type Ref, ref } from 'vue'
 import SVGTrashCan from '~/assets/icons/trash-can.svg'
 import SVGProfileIcon from '~/assets/icons/profile-icon.svg'
 import SVGUndoIcon from '~/assets/icons/undo-icon.svg'
@@ -112,6 +131,8 @@ const localeStore = useLocaleStore()
 const facilitiesStore = useFacilitiesStore()
 const moderationScreenStore = useModerationScreenStore()
 const healthcareProfessionalsStore = useHealthcareProfessionalsStore()
+
+const isEditable: Ref<boolean> = ref(false)
 
 // This checks whether an existing healthcare professional has been added for removal
 const isHealthcareProfessionalReadyForRemoval = (id: string) =>
@@ -142,8 +163,26 @@ const undoRemovalOfHealthcareProfessional = (id: string) => {
             .filter((healthcareProfessionalRelation: Relationship) => healthcareProfessionalRelation.otherEntityId !== id)
 }
 
+// This changes the component details to editable with an optional parameter to use in the callback function
+const setChangeToEditable = (param?: unknown) => {
+    if (props.setOnClick && param) props.setOnClick(param)
+    if (props.setOnClick && !param) props.setOnClick()
+
+    isEditable.value = true
+}
+
+// This changes the component to not editable and saves the changes based on updateOnClickFunction
+const setToUneditableAndSave = (param?: unknown) => {
+    if (props.updateOnClick && param) props.updateOnClick(param)
+    if (props.updateOnClick && !param) props.updateOnClick()
+
+    isEditable.value = false
+}
+
 const props = defineProps<{
     healthcareProfessional: HealthcareProfessional
     healthcareProfessionalsRelatedToFacility?: string[]
+    setOnClick?: (param?: unknown) => void
+    updateOnClick?: (param?: unknown) => void
 }>()
 </script>
--- a/components/ModDashboardHealthProfessionalCard.vue
+++ b/components/ModDashboardHealthProfessionalCard.vue
@@ -51,7 +51,10 @@
                     </div>
                 </div>
             </div>
-            <div v-if="moderationScreenStore.activeScreen === ModerationScreen.EditSubmission">
+            <div
+                v-if="moderationScreenStore.activeScreen === ModerationScreen.EditSubmission
+                    || moderationScreenStore.activeScreen === ModerationScreen.EditHealthcareProfessional"
+            >
                 <div
                     v-for="(nameLocale, index) in healthcareProfessional.names"
                     :key="`${nameLocale.firstName}-${nameLocale.lastName}-${index}`"
--- a/components/ModDashboardHealthProfessionalCard.vue
+++ b/components/ModDashboardHealthProfessionalCard.vue
@@ -11,7 +11,7 @@
                     class="profile-icon stroke-primary w-16 h-16 stroke-1 inline mx-1 self-start"
                 />
             </div>
-            <div v-if="moderationScreenStore.activeScreen === ModerationScreen.EditFacility">
+            <div v-show="healthcareProfessionalsRelatedToFacility">
                 <div class="flex flex-col h-full w-64 pl-1 mb-1">
                     <div class="flex font-bold pt-2">
                         <span>{{ healthcareProfessionalsStore
@@ -52,8 +52,7 @@
                 </div>
             </div>
             <div
-                v-if="moderationScreenStore.activeScreen === ModerationScreen.EditSubmission
-                    || moderationScreenStore.activeScreen === ModerationScreen.EditHealthcareProfessional"
+                v-show="!healthcareProfessionalsRelatedToFacility"
             >
                 <div
                     v-for="(nameLocale, index) in healthcareProfessional.names"
--- a/components/ModEditHealthcareProfessionalSection.vue
+++ b/components/ModEditHealthcareProfessionalSection.vue
@@ -58,7 +58,7 @@
                 <button
                     v-show="isEditSubmissionScreen"
                     type="button"
-                    class="bg-tertiary text-white font-bold py-2 px-4 my-2 rounded w-56"
+                    class="bg-tertiary text-primary-text font-bold py-2 px-4 my-2 rounded w-56"
                     @click="handleLocalizedName"
                 >
                     {{ $t('modHealthcareProfessionalSection.addHealthCareProfessionalLocaleName') }}
--- a/components/ModEditHealthcareProfessionalSection.vue
+++ b/components/ModEditHealthcareProfessionalSection.vue
@@ -46,7 +46,7 @@
                                 text-primary-text text-sm font-normal font-sans placeholder-primary-text-muted"
                     @change=" () => healthcareProfessionalsStore
                         .updateHealthcareProfessionalNameValues(healthcareProfessionalsStore.selectedHealthcareProfessionalData,
-                                                                sectionFields.nameLocale)"
+                                                                healthcareProfessionalModel.nameLocale)"
                 >
                     <option
                         v-for="(locale, index) in Locale"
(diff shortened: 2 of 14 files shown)


components/ModDashboardHealthProfessionalCard.vue
```diff
@@ -51,7 +51,10 @@
                     </div>
                 </div>
             </div>
-            <div v-if="moderationScreenStore.activeScreen === ModerationScreen.EditSubmission">
+            <div
+                v-if="moderationScreenStore.activeScreen === ModerationScreen.EditSubmission
```
ermish, line 55: a good design practice is for components to never know where they live on a page. (this encourages reusability)
let's move this v-if so the parent can pass in this condition as `isEditable`

stores/healthcareProfessionalsStore.ts
```diff
@@ -15,6 +20,61 @@ export const useHealthcareProfessionalsStore = defineStore(
         const healthcareProfessionalsData: Ref<HealthcareProfessional[]>
         = ref([])
         const selectedHealthcareProfessionalId: Ref<string> = ref('')
+        const selectedHealthcareProfessionalData: Ref<HealthcareProfessional | undefined> = ref()
+        const healthcareProfessionalSectionFields = reactive({
```
ermish, line 24: sectionFields to me is a bit confusing. Section of what? 
It's important to remember that stores are meant to be abstracted from the specific components using them. 

Is this `selectedHealthcareProfessional`?

stores/healthcareProfessionalsStore.ts
```diff
@@ -15,6 +20,61 @@ export const useHealthcareProfessionalsStore = defineStore(
...
+
+        function setSelectedHealthcareProfessional(healthcareProfessionalId: string) {
+            selectedHealthcareProfessionalData.value = healthcareProfessionalsData.value
+                .find((healthcareProfessional: HealthcareProfessional) => healthcareProfessional.id === healthcareProfessionalId)
+        }
+
+        function initializeHealthcareProfessionalValues(selectedHealthcareProfessionalData: HealthcareProfessional | undefined) {
+            if (!selectedHealthcareProfessionalData) return
+
+            healthcareProfessionalSectionFields.healthcareProfessionalNameArray = selectedHealthcareProfessionalData.names
```
ermish, line 44: This function seems more like a function for mapping from api data to component format rather than an initializer function. Perhaps this just needs a rename?

Also, it looks like where this is being used, we could just call this from the `setSelectedHealthcareProfessional` method internally instead of from the component. Is there a use case outside of the set method?

stores/healthcareProfessionalsStore.ts
```diff
@@ -15,6 +20,61 @@ export const useHealthcareProfessionalsStore = defineStore(
...
+                = selectedHealthcareProfessionalData.specialties
+            healthcareProfessionalSectionFields.healthcareProfessionalLocales
+                = selectedHealthcareProfessionalData.spokenLanguages
+        }
+
+        function updateHealthcareProfessionalNameValues(selectedHealthcareProfessionalData: HealthcareProfessional |
+          undefined, chosenLocale: Locale) {
+            if (!selectedHealthcareProfessionalData) return
+
+            const localeIndex = selectedHealthcareProfessionalData.names
```
ermish, line 65: good check! ✅

utils/multiSelectWithoutKeyboard.ts
```diff
@@ -28,13 +28,14 @@ export const multiSelectWithoutKeyboard = <T extends string>(
...
+            const index = arrayToUpdate.value.indexOf(value)
+            if (index === -1) {
+                arrayToUpdate.value.push(value)
+                return
+            }
 
-        if (index === -1) {
-            arrayToUpdate.value.push(value)
-            return
+            arrayToUpdate.value.splice(index, 1)
```
ermish, line 38: nice improvement! While I understand it after I read through it, it might be good to add some comments here to help other devs be understand what this function is doing (for example, what is the splice and push doing?

components/ModEditHealthcareProfessionalSection.vue
```diff
@@ -1,233 +1,199 @@
...
-                                $t('modHealthcareProfessionalSection.delete') }}
-                        </span>
-                    </div>
+                        {{ locale }}
+                    </option>
+                </select>
+                <button
+                    v-show="isEditSubmissionScreen"
+                    type="button"
+                    class="bg-tertiary text-white font-bold py-2 px-4 my-2 rounded w-56"
```
ermish, line 61: It looks like this class was carried over, but it's a good time to change hard coded colors like`text-white` to a theme color like `text-primary-text` or `text-primary-text-inverted`

components/ModEditHealthcareProfessionalSection.vue
```diff
@@ -1,233 +1,199 @@
...
+let toast: ToastInterface
+
+const route = useRoute()
+
+const { t } = useI18n()
 
 const moderationScreenStore = useModerationScreenStore()
-const professionalNameArray = healthcareProfessionalSectionFields.healthcareProfessionalNameArray
+const healthcareProfessionalsStore = useHealthcareProfessionalsStore()
+const sectionFields = healthcareProfessionalsStore.healthcareProfessionalSectionFields
```
ermish, line 196: can this just be `healthcareProfessional`, `healthcareProfessionalFields`, or `healthcareProfessionalModel`? 
Any of these work for me :)

components/ModEditHealthcareProfessionalSection.vue
```diff
@@ -238,52 +204,114 @@ const specialtyList = Object.values(Specialty) as Specialty[]
...
+        console.error(t('modHealthcareProfessionalSection.errorMessageHealthcareProfessionalId'))
+        toast.error(t('modHealthcareProfessionalSection.errorMessageHealthcareProfessionalId'))
+        return
+    }
+
+    if (!healthcareProfessionalsStore.healthcareProfessionalsData.length) {
+        await healthcareProfessionalsStore.getHealthcareProfessionals()
+    }
+
+    healthcareProfessionalsStore.selectedHealthcareProfessionalId = id as string
```
ermish, line 260: looking at this, I just thought of a way to simplify this. 
This id should be set within the `setSelectedHealthcareProfessional` function so the caller doesn't have to do it.

components/ModEditHealthcareProfessionalSection.vue
```diff
@@ -1,233 +1,199 @@
...
-                {{ $t('modHealthcareProfessionalSection.labelHealthcareProfessionalNameLocale') }}
-            </label>
-            <select
-                id="name_locales"
-                v-model="healthcareProfessionalSectionFields.nameLocale"
-                data-testid="mod-healthcare-professional-section-name-locale"
-                name="Name Locales"
-                class="mb-5 px-3 py-3.5 w-96 h-12 bg-secondary-bg rounded-lg border border-primary-text-muted
-                            text-primary-text text-sm font-normal font-sans placeholder-primary-text-muted"
+    <div v-if="isHealthcareProfessionalInitialized">
```
ermish, line 2: I think a great future issue would be to show a loader placeholder while this is being initialized! This can reduce the screen jumping all over while things are loading. 
Do you want to take a shot at writing one up for us @ShizuokaTerrier ?

components/ModEditHealthcareProfessionalSection.vue
```diff
@@ -238,52 +204,114 @@ const specialtyList = Object.values(Specialty) as Specialty[]
...
+        healthcareProfessionalDegreesArray.value,
+        healthcareProfessionalSpecialtiesArray.value,
+        healthcareProfessionalSpokenLanguages.value
+    ] as [
+        Insurance[],
+        Degree[],
+        Specialty[],
+        Locale[]
+    ],
+    ([newInsurances, newDegrees, newSpecialties, newLocales]) => {
```
ermish, line 310: might be good to explain why this watch is needed. (intuitively something would think they auto update if they're `ref`s)

ermish (changes requested):
nice work @ShizuokaTerrier !

components/ModDashboardHealthProfessionalCard.vue
NabbeunNabi: I know @ermish that components shouldn't know where they are as a design. But this has nothing to do with if it is editable. It just is whether it displays the names as a locale because that is what you change for a healthcare professional or whether it shows other information when in the `edit-facility` page based on the pertinent information to the user. 

Is there a more standard way there? Like giving it a type as a prop? Like `healthcarepProfessionalView` and `facilityView`

NabbeunNabi: Or maybe to be clearer. Not specifically if it is editable. Since they can be used to delete HPs in the facility as well

stores/healthcareProfessionalsStore.ts
ShizuokaTerrier: In combination with the initialize function it is. In that case, would a name like `selectedHealthcareProfessionalEditableFields` be better? It's clear, but a bit of a mouthful.

stores/healthcareProfessionalsStore.ts
ShizuokaTerrier: I'm presuming the `setSelectedHealthcareProfessional` is potentially going to be used elsewhere without needing to initialize values. Is that right @NabbeunNabi ?

components/ModEditHealthcareProfessionalSection.vue
ShizuokaTerrier: Sounds fun @ermish!

components/ModDashboardHealthProfessionalCard.vue
ermish: gotcha, to that point, it wasn't super clear to me when looking at it, what the v-if is for. If it's just to show names as a locale, then that would be your boolean condition. 
buuuut I think there's a better standard (like you asked @NabbeunNabi ) for when you have different conditional fields based on which page it's in. 
Its called `conditional rendering` or `dynamicUI`. It follows another similar pattern which I'm a huge fan of is `composition`. Instead of making arbitrary groupings, you make each independent piece conditional. Then the caller can pick and choose what they want. 

So instead of making an `if edit page` and `if x page` and `if this view`, you just make each field conditionally rendered IF the data is passed in. So in this example, just make each of the fields under this div only render if the parent passes it in. like `localNames` and whatever else is needed for `edit-facility` 

This is hard to explain over typing so we can slack chat if you want to walk through it :)

stores/healthcareProfessionalsStore.ts
ermish: that's fine by me if that's accurate. I'm not sure why this one is editable and the `selectedHealthcareProfessionalData` is not. (I think it  should be) Can we combine these two and just make the selected one reactive? I don't see why we need to offer two different ones.

stores/healthcareProfessionalsStore.ts
NabbeunNabi: At the moment it is only being used here to initialize the fields from the selected healthcare professional. This could be possibly used in the future. We definitely could call it when we set the selected healthcare professional. It does make sense now as @ermish is saying to call it internally

components/ModDashboardHealthProfessionalCard.vue
NabbeunNabi: It displays the information differently @ermish based on which route it is on because the pertinent information is different for the user depending on the view. Should it just allow for props that can be just the locale or the whole healthcare professional? I will try that out anyway and then you can let me know if there is a better way. I understand what `conditional rendering` is but am not too familiar with `dynamicUI`.

components/ModEditHealthcareProfessionalSection.vue
NabbeunNabi: @ShizuokaTerrier Did you write the ticket for this? Just curious as to whether I can resolve the conversation or not or plan on doing it this week?

ShizuokaTerrier: @NabbeunNabi it's done now. I'll resolve this conversation.

components/ModEditHealthcareProfessionalSection.vue
```diff
@@ -257,44 +255,51 @@ onBeforeMount(async () => {
...
 
     multiSelectWithoutKeyboard(
-        '#healthcare-professional-locales',
+        '#locales',
         healthcareProfessionalSpokenLanguages,
         selectedLocale
     )
 })
 
+/* This watch is necessary in order to update the reactive object
```
ermish, line 296: nice!

stores/healthcareProfessionalsStore.ts
```diff
@@ -20,60 +16,87 @@ export const useHealthcareProfessionalsStore = defineStore(
...
-            healthcareProfessionalNameArray: [] as Array<LocalizedNameInput>,
-            localizedFirstName: '' as string,
-            localizedLastName: '' as string,
-            localizedMiddleName: '' as string,
-            nameLocale: Locale.EnUs as Locale,
-            healthcareProfessionalAcceptedInsurances: [] as Array<Insurance>,
-            healthcareProfessionalDegrees: [] as Array<Degree>,
-            healthcareProfessionalSpecialties: [] as Array<Specialty>,
-            healthcareProfessionalLocales: [] as Array<Locale>
+        const healthcareProfessionalSectionFields = reactive<HealthcareProfessional>({
```
ermish, line 19: naming nit: I still don't know what "section" this is a part of, but if it makes sense to ya'll, I'll allow it @ShizuokaTerrier @NabbeunNabi

stores/healthcareProfessionalsStore.ts
```diff
@@ -20,60 +16,87 @@ export const useHealthcareProfessionalsStore = defineStore(
...
+        }
 
-            healthcareProfessionalSectionFields.healthcareProfessionalNameArray = selectedHealthcareProfessionalData.names
-            healthcareProfessionalSectionFields.localizedFirstName = selectedHealthcareProfessionalData.names[0].firstName
-            healthcareProfessionalSectionFields.localizedMiddleName = selectedHealthcareProfessionalData.names[0].middleName
-                ? selectedHealthcareProfessionalData.names[0].middleName
+        function setSelectedNameLocaleToUpdate(nameLocale: Locale) {
+            if (!healthcareProfessionalSectionFields.names.length) return
+
+            const nameLocaleToEdit = healthcareProfessionalSectionFields.names.find(name => name.locale === nameLocale)
```
ermish, line 65: nice find

ermish (approved):
nice updates! a lot of good improvements and valuable comments. The store functions make more sense now

PR #956, opened 2025-01-08
feat: implement updating an existing healthcare professional 

### LINKED ISSUE #795: Add functionality to Mod Healthcare Form for updating a healthcare professional
# Problem
We need the healthcare professionals in our database to be able to be updated based on the UI of the form in #784. Right now there is no graphQL mutation to our backend to update an existing healthcare professional.

# Success
The functionality with the graphQL mutation is implemented from the Mod Healthcare Professional form in the moderation panel.
### Requirements

- [x] graphQL mutation is implemented and works as desired
- [x] Error toasts are shown on failure with proper localization in the corresponding component

NabbeunNabi:
Resolves #795 

## 📸 Screenshots

-   ### Before

-   ### After

### DIFF
--- a/components/ModDashboardHealthProfessionalCard.vue
+++ b/components/ModDashboardHealthProfessionalCard.vue
@@ -55,14 +55,15 @@
                 </div>
             </div>
             <div
-                v-if="healthcareProfessionalNameByLocale"
+                v-if="healthcareProfessionalNameByLocale || removedHealthcareProfessionalName"
                 class="min-w-44"
             >
                 <div class="flex font-bold pt-2">
-                    <span>{{ healthcareProfessionalNameByLocale.lastName }}</span>
-                    <span class="mx-2">{{ healthcareProfessionalNameByLocale.firstName }}</span>
+                    <span>{{ healthcareProfessionalNameByLocale.lastName || removedHealthcareProfessionalName?.lastName }}</span>
+                    <span class="mx-2">{{ healthcareProfessionalNameByLocale.firstName
+                        || removedHealthcareProfessionalName?.firstName }}</span>
                     <span v-show="healthcareProfessionalNameByLocale.middleName">
-                        {{ healthcareProfessionalNameByLocale.middleName }}
+                        {{ healthcareProfessionalNameByLocale.middleName || removedHealthcareProfessionalName?.middleName }}
                     </span>
                 </div>
                 <span
@@ -92,7 +93,9 @@
                 </div>
             </div>
             <div
-                v-if="!isHealthcareProfessionalReadyForRemoval(healthcareProfessional.id)"
+                v-if="!isHealthcareProfessionalReadyForRemoval(healthcareProfessional.id)
+                    && !healthcareProfessionalsStore.removedHealthcareProfessionalNames
+                        .find(name => name.locale === healthcareProfessionalNameByLocale.locale)"
                 class="flex w-8 items-center justify-center
                     cursor-pointer font-bold text-secondary text-sm self-start p-1"
                 @click="() => removeHealthcareProfessional(healthcareProfessional.id)"
@@ -102,7 +105,9 @@
                 />
             </div>
             <div
-                v-if="isHealthcareProfessionalReadyForRemoval(healthcareProfessional.id)"
+                v-if="isHealthcareProfessionalReadyForRemoval(healthcareProfessional.id)
+                    || healthcareProfessionalsStore.removedHealthcareProfessionalNames
+                        .find(name => name.locale === healthcareProfessionalNameByLocale.locale)"
                 class="flex w-8 items-center justify-center
                     cursor-pointer font-bold text-secondary text-sm self-start p-1"
                 @click="() => undoRemovalOfHealthcareProfessional(healthcareProfessional.id)"
@@ -116,7 +121,7 @@
 </template>
 
 <script setup lang="ts">
-import { type Ref, ref } from 'vue'
+import { type Ref, ref, type ComputedRef, computed } from 'vue'
 import SVGTrashCan from '~/assets/icons/trash-can.svg'
 import SVGProfileIcon from '~/assets/icons/profile-icon.svg'
 import SVGUndoIcon from '~/assets/icons/undo-icon.svg'
@@ -124,12 +129,35 @@ import { useLocaleStore } from '~/stores/localeStore'
 import { useFacilitiesStore } from '~/stores/facilitiesStore'
 import { useHealthcareProfessionalsStore } from '~/stores/healthcareProfessionalsStore'
 import { ModerationScreen, useModerationScreenStore } from '~/stores/moderationScreenStore'
-import { RelationshipAction, type HealthcareProfessional, type LocalizedNameInput, type Relationship } from '~/typedefs/gqlTypes'
+import { RelationshipAction,
+    type HealthcareProfessional,
+    type LocalizedNameInput,
+    type Relationship } from '~/typedefs/gqlTypes'
 
 const localeStore = useLocaleStore()
 const facilitiesStore = useFacilitiesStore()
 const moderationScreenStore = useModerationScreenStore()
 const healthcareProfessionalsStore = useHealthcareProfessionalsStore()
+
+// This finds and keeps track of the index of the names if there are multiple for locales
+const indexOfNameLocale: ComputedRef<number | undefined> = computed(() => {
+    // Check the first array (names)
+    let index = healthcareProfessionalsStore.healthcareProfessionalSectionFields.names
+        .findIndex(name => props.healthcareProfessionalNameByLocale
+            ? name.locale === props.healthcareProfessionalNameByLocale.locale
+            : false)
+
+    // If not found, check the second array (removedHealthcareProfessionalNames)
+    if (index === -1) {
+        index = healthcareProfessionalsStore.removedHealthcareProfessionalNames
+            .findIndex(name => props.healthcareProfessionalNameByLocale
+                ? name.locale === props.healthcareProfessionalNameByLocale.locale
+                : false)
+    }
+
+    // Return the found index (or undefined if not found)
+    return index === -1 ? undefined : index
+})
 
 const isEditable: Ref<boolean> = ref(false)
 
@@ -139,7 +167,7 @@ const isHealthcareProfessionalReadyForRemoval = (id: string) =>
         .find(healthcareProfessionalRelation => healthcareProfessionalRelation.otherEntityId === id
         && healthcareProfessionalRelation.action === RelationshipAction.Delete)
 
-const removeHealthcareProfessional = (id: string) => {
+const removeHealthcareProfessional = (id: string = '0') => {
     switch (moderationScreenStore.activeScreen) {
         case ModerationScreen.EditFacility:
             if (props.healthcareProfessionalsRelatedToFacility && props.healthcareProfessionalsRelatedToFacility.includes(id)) {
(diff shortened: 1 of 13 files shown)


stores/healthcareProfessionalsStore.ts
```diff
@@ -107,16 +115,87 @@ export const useHealthcareProfessionalsStore = defineStore(
...
+            const serverResponse = { data: {} as Maybe<HealthcareProfessional>, errors: [] as ServerError[], hasErrors: false }
+
+            const facilitiesForRelationshipCreationArray = Array.from(selectedFacilities.value)
+
+            if (facilitiesForRelationshipCreationArray.length) {
+                //For each is used here to only add the necessary actions to the already created ref array of relationships
+                facilitiesForRelationshipCreationArray.forEach(facility => createFacilitiesRelationArray(facility))
+            }
+            const updateHealthcareProfessionalInput: MutationUpdateHealthcareProfessionalArgs = {
+                id: selectedHealthcareProfessionalId.value,
```
Anissa3005, line 129: I have been gone from the frontend sometime but, in this mutation, is it possible to only update a part of a healthcare professional. For example what if you don't need to update the acceptedInsurance, what value is healthcareProfessionalSectionFields.acceptedInsurance going to be when we make this request?

NabbeunNabi: It is possible. But as this PR is long and adds the basic functionality. I thought maybe in the future we can add validation to see that it has been updated.

Currently it will send whatever has been changed or the original array stored within the database.

I could add it here. I have a function that does check already if there are changes to some of those arrays. For pressing the button.

So that could be moved into utilities. But at the moment it only checks the differences for arrays with primitive types. So it wouldn't check with objects or arrays of arrays. So either I can pull it out and use it in this one or make a separate issue. 

I would vote have it be something I do after the one that sends all of them and make another ticket and have it be a refactor ticket where it sets those keys to undefined if there are no changes.

It is a good point though. What do you think @Anissa3005 @theyokohamalife @ermish. Do it in this PR or another one after. I am a bit worried about making this one too long.

Anissa3005: I feel my questions was maybe not clear but can we partially update a healthcare professional with this logic in place? Since to me (I have been gone from the front some time, maybe I'm just rusty) it looks like that mutation cannot be a partial update, it needs to contain all those fields?

components/ModDashboardHealthProfessionalCard.vue
```diff
@@ -55,15 +55,15 @@
                 </div>
             </div>
             <div
-                v-if="healthcareProfessionalNameByLocale || removedHealthcareProfessionalName"
```
theyokohamalife: Were these unused variables introduced by this PR? 🤔 If so, we should avoid creating and removing something in the same PR.

stores/healthcareProfessionalsStore.ts
NabbeunNabi: @Anissa3005 with this logic we update all of it except for the facility relationships at the moment

components/ModEditFacilityOrProfessionalTopbar.vue
```diff
@@ -178,17 +179,47 @@ const facilityHasUnsavedChanges = () => {
         !== JSON.stringify(facilitySections.healthcareProfessionalIds)
         || facilitySections.healthProfessionalsRelations.length
 
-    return areThereUnsavedChanges
+        return areThereUnsavedChanges
```
theyokohamalife, line 200: Since the value of this variable is a bool, a simpler name would be `hasUnsavedChanges`.

components/ModEditFacilityOrProfessionalTopbar.vue
```diff
@@ -178,17 +179,47 @@ const facilityHasUnsavedChanges = () => {
...
+    }
+    if (moderationScreenStore.activeScreen === ModerationScreen.EditHealthcareProfessional) {
+        const originalHealthcareProfessional = healthcareProfessionalsStore.healthcareProfessionalsData.find(
+            professional => professional.id === healthcareProfessionalsStore.selectedHealthcareProfessionalId
+        )
+
+        if (!originalHealthcareProfessional) return false // No match found, no changes to compare.
+
+        // Compare each field in the `healthcareProfessionalSectionFields` object with the original data.
+        const areThereUnsavedChanges
```
theyokohamalife, line 208: ♻️

components/ModEditFacilityOrProfessionalTopbar.vue
```diff
@@ -178,17 +179,47 @@ const facilityHasUnsavedChanges = () => {
...
 const updateFacilityOrHealthcareProfessional = async () => {
     // This makes the on click update the facility if the screen is EditFacility
-    if (moderationScreenStore.activeScreen === ModerationScreen.EditFacility) {
-        const checkForUnsavedChanges = facilityHasUnsavedChanges()
-
-        // This prevents us from sending a requested unnecessarily if the user has not made changes
-        if (!checkForUnsavedChanges) return
+    const checkForUnsavedChanges = facilityOrHPHasUnsavedChanges()
+    // This prevents us from sending a requested unnecessarily if the user has not made changes
+    if (!checkForUnsavedChanges) return
```
theyokohamalife, line 246: Could we simplify this to `if (!facilityOrHPHasUnsavedChanges()) return`

stores/healthcareProfessionalsStore.ts
Anissa3005: I think we should be able to only update the fields that are necessary, so partial updating for example the insurance changed we send a request to update that field only. (can be in another PR of course)

components/ModEditFacilityOrProfessionalTopbar.vue
```diff
@@ -178,17 +197,55 @@ const facilityHasUnsavedChanges = () => {
         !== JSON.stringify(facilitySections.healthcareProfessionalIds)
         || facilitySections.healthProfessionalsRelations.length
 
-    return areThereUnsavedChanges
+        return areThereUnsavedChanges
+    }
+    if (moderationScreenStore.activeScreen === ModerationScreen.EditHealthcareProfessional) {
```
theyokohamalife, line 202: We perform these active screen checks in a lot of places in the repo. At some point (and not necessarily this PR) we should write functions to make them a bit cleaner. IE `if (activeScreenIsEditHealthcareProfessional())`.

components/ModEditFacilityOrProfessionalTopbar.vue
```diff
@@ -201,18 +258,46 @@ const updateFacilityOrHealthcareProfessional = async () => {
         toast.success(t('modEditFacilityOrHPTopbar.facilityUpdatedSuccessfully'))
         return response
     }
+    // This makes the on click update the facility if the screen is EditFacility
+    if (moderationScreenStore.activeScreen === ModerationScreen.EditHealthcareProfessional) {
+        // This prevents us from sending a requested unnecessarily if the user has not made changes
+        if (!checkForUnsavedChanges) return
```
theyokohamalife, line 264: ♻️

components/ModEditFacilityOrProfessionalTopbar.vue
theyokohamalife: After reviewing this some more I realized that you are declaring `areThereUnsavedChanges` twice in the same function, but they represent very different values. One is if the facility has unsaved changes, and the other is if the healthcare professional has unsaved changes. These should be two distinct variables. Something like `facilityHasUnsavedChanges` and `healthcareProfessionalHasUnsavedChanges`.

stores/healthcareProfessionalsStore.ts
NabbeunNabi: I agree. I feel like that would be a good issue to add a utility function that checks an array passed with the primitive values that are held in that array and if not they key is undefined and that will make it partial

components/ModDashboardHealthProfessionalCard.vue
```diff
@@ -52,48 +55,47 @@
...
+                    >
+                        {{ $t("modHealthcareProfessionalCard.saveName") }}
+                    </button>
                 </div>
             </div>
             <div
-                v-if="!isHealthcareProfessionalReadyForRemoval(healthcareProfessional.id)"
+                v-if="!isHealthcareProfessionalReadyForRemoval(healthcareProfessional.id)
+                    && !healthcareProfessionalsStore.removedHealthcareProfessionalNames
+                        .find(name => name.locale === healthcareProfessionalNameByLocale.locale)"
```
theyokohamalife, line 98: We should move this logic out of the template and into the script.

components/ModDashboardHealthProfessionalCard.vue
```diff
@@ -103,7 +105,9 @@
                 />
             </div>
             <div
-                v-if="isHealthcareProfessionalReadyForRemoval(healthcareProfessional.id)"
+                v-if="isHealthcareProfessionalReadyForRemoval(healthcareProfessional.id)
+                    || healthcareProfessionalsStore.removedHealthcareProfessionalNames
+                        .find(name => name.locale === healthcareProfessionalNameByLocale.locale)"
```
theyokohamalife, line 110: ♻️

components/ModEditFacilityOrProfessionalTopbar.vue
theyokohamalife: I thought about this a bit more, and I think it would be better if you kept them as separate functions. Meaning, keep `facilityHasUnsavedChanges` and create a new `healthcareProfessionalHasChanges`. Then call the two functions in `hasUnsavedChanges`. The reason for this is because otherwise it becomes a big function performing multiple things, when we should try to aim for 1 purpose per function.

components/ModEditFacilityOrProfessionalTopbar.vue
NabbeunNabi: You have address this in #973

components/ModEditFacilityOrProfessionalTopbar.vue
NabbeunNabi: ♻️

ermish:
@NabbeunNabi for the UX, there's a couple concerns I've got. 
- the edit button location to me is confusing. The card feels like nice example rendering to display how their name would look on the card, but the edit and delete buttons should not be in it. Can we move them into a separate container?
- For the locale selector, the call to action is a bit strange. Am I adding a name per selected locale? How do I know which ones are set? I would recommend we could start with something like an array of name cards for each locale and they can add/remove one. This should work because most people don't have a name in more than 3 languages.

NabbeunNabi:
> @NabbeunNabi for the UX, there's a couple concerns I've got.
> 
> * the edit button location to me is confusing. The card feels like nice example rendering to display how their name would look on the card, but the edit and delete buttons should not be in it. Can we move them into a separate container?
> * For the locale selector, the call to action is a bit strange. Am I adding a name per selected locale? How do I know which ones are set? I would recommend we could start with something like an array of name cards for each locale and they can add/remove one. This should work because most people don't have a name in more than 3 languages.

@ermish updated. Will add another video with the update to the PR description

ermish:
Loving the updates! Looks like you got all the UX changes in. 
One more usability change: 
Can you make the fields section attached to the preview card? 
It doesn't look like they're the same item.

stores/healthcareProfessionalsStore.ts
```diff
@@ -62,61 +61,92 @@ export const useHealthcareProfessionalsStore = defineStore(
...
-            try {
-                return await graphQLClientRequestWithRetry(
-                    gqlClient.request.bind(gqlClient),
-                    updateHealthcareProfessionalGqlMutation,
-                    healthcareProfessional
+        async function updateHealthcareProfessional():
+        Promise<ServerResponse<Maybe<HealthcareProfessional>>> {
+            const serverResponse = { data: {} as Maybe<HealthcareProfessional>, errors: [] as ServerError[], hasErrors: false }
+
+            const facilitiesForRelationshipCreationArray = Array.from(selectedFacilities.value)
```
ermish, line 73: would an array vs Set make things easier?

ermish (changes requested):
ton of work in here! here's those comments I forgot to hit send on

stores/healthcareProfessionalsStore.ts
NabbeunNabi: It would! Can we do this in a separate issue and PR as a refactoring ticket? Or would it be better here? That would end up touching multiple files and add more complexity to this PR. 🤔

components/ModEditFacilityOrProfessionalTopbar.vue
```diff
@@ -145,7 +146,24 @@ function setSelectedId() {
     }
 }
 
-const facilityHasUnsavedChanges = () => {
+type ComparableArray = Insurance[] | Locale[] | Degree[] | Specialty[] | string[]
+
+//This is a helper function that sorts and compares that all the values in a primitive value array are equal
+const arraysAreEqual = (array1: ComparableArray,
```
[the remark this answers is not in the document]
NabbeunNabi: Moved and I am thinking about adding an issue to update this to check not only arrays of primitive types to make this even more utility.

components/ModDashboardHealthProfessionalCard.vue
```diff
@@ -1,188 +1,192 @@
...
-                            @click="() => setChangeToEditable(nameLocale.locale)"
+                            class="flex items-center justify-center w-6 h-6"
+                            @click="() => setChangeToEditable()"
                         >
-                            {{ $t("modHealthcareProfessionalCard.editName") }}
+                            <SVGPencilIcon />
                         </button>
                         <button
-                            v-if="isEditable && updateOnClick"
+                            v-if="isEditable && chosenLocaleIndex === 0"
```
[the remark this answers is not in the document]
NabbeunNabi: comment added near the button and near the prop for understanding

components/ModDashboardHealthProfessionalCard.vue
```diff
@@ -1,188 +1,192 @@
...
                         </span>
-                    </div>
-                    <span>
-                        ID:  {{ healthcareProfessional.id }}
-                    </span>
-                    <div>
-                        <div
-                            class="flex flex-wrap justify-start items-start self-end pt-2"
-                        >
+                        <div>
```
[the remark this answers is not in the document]
NabbeunNabi: ids activated!

stores/healthcareProfessionalsStore.ts
ermish: Good refactor issue ✌️

ermish (approved):
LGTM!

Issue #966, opened 2025-01-16
Fix typescript error for localization string `$t`


NabbeunNabi:
# Problem
Currently there is a typescript error displaying for files that contain `$t` for translatable strings. This was previously fixed but seems to have become a problem again. We want to fix this as we do not want any `typescript` errors within our codebase.

# Success
This problem is researched and fixed. There is clear understanding through comments and a well written PR as to why and how this happened so going forward we can make sure this problem doesn't happen again, or can be resolved quickly.

### Requirements

- [x] The cause of the error has been found and understood
- [x] The error has been fixed
- [x] There are comments in the code if applicable explaining how this happened so it can be prevented going forward. If comments are not appropriate in the code it is communicated through a clear and thorough PR description.

# Implementation notes
This problem started when we updated our dependency in this PR. It should be related to this package version upgrade.

#939 

You can checkout into the commit before this PR (by looking at the commits in the main branch) and build `.nuxt` to see that there are no errors and then checkout into the hash of the commit of the PR to see that the errors then show up.

PR #967, opened 2025-01-16
feat: upgrade Nuxt to 3.15.2

gminetoma:
Resolves [#966](https://github.com/ourjapanlife/findadoc-web/issues/966)

## 🔧 What changed

- Update Nuxt to the latest version
- Add a declaration file containing a patch for the issue

## 🔍 Context

The main issue was that Nuxt previously used the `@vue/runtime-core` namespace to augment the `vue` namespace. This caused conflicts with any package using `vue`.

As of Nuxt v3.13.0, they now use the `vue` namespace directly, which is the recommended approach. However, some packages, such as `@auth0/auth0-vue`, still rely on `@vue/runtime-core`, leading to compatibility issues.

Our solution involved creating a declaration file to patch this problem.

## 📖 Reference

https://nuxt.com/blog/v3-13#vue-typescript-changes

### DIFF
--- a/declarations.d.ts
+++ b/declarations.d.ts
@@ -1,0 +1,18 @@
+/* eslint-disable @typescript-eslint/no-empty-object-type */
+
+/*
+This is a patch for outdated packages that still use `@vue/runtime-core` to augment the `vue` namespace.
+We can remove this file once all packages update their code.
+Removing this might show TypeScript linting errors.
+https://nuxt.com/blog/v3-13#vue-typescript-changes
+*/
+
+import type {
+    ComponentCustomOptions as _ComponentCustomOptions,
+    ComponentCustomProperties as _ComponentCustomProperties
+} from 'vue'
+
+declare module '@vue/runtime-core' {
+    interface ComponentCustomProperties extends _ComponentCustomProperties {}
+    interface ComponentCustomOptions extends _ComponentCustomOptions {}
+}
--- a/components/ModDashboardHealthProfessionalCard.vue
+++ b/components/ModDashboardHealthProfessionalCard.vue
@@ -172,7 +172,7 @@ const healthcareProfessionalsStore = useHealthcareProfessionalsStore()
 const isHealthcareProfessionalReadyForRemoval = (id: string) =>
     facilitiesStore.facilitySectionFields.healthProfessionalsRelations
         .find(healthcareProfessionalRelation => healthcareProfessionalRelation.otherEntityId === id
-        && healthcareProfessionalRelation.action === RelationshipAction.Delete)
+          && healthcareProfessionalRelation.action === RelationshipAction.Delete)
 
 const removeHealthcareProfessional = (id: string = '0') => {
     if (props.healthcareProfessionalsRelatedToFacility && props.healthcareProfessionalsRelatedToFacility.includes(id)) {
--- a/components/ModEditFacilityOrProfessionalTopbar.vue
+++ b/components/ModEditFacilityOrProfessionalTopbar.vue
@@ -160,25 +160,25 @@ const facilityHasUnsavedChanges = () => {
 
     const areThereUnsavedFacilityChanges
         = facilityBeforeChange.nameEn !== facilitySections.nameEn
-        || facilityBeforeChange.nameJa !== facilitySections.nameJa
-        || facilityBeforeChange.contact.phone !== facilitySections.phone
-        || facilityBeforeChange.contact.website !== facilitySections.website
-        || facilityBeforeChange.contact.email !== facilitySections.email
-        || facilityBeforeChange.contact.address.postalCode !== facilitySections.postalCode
-        || facilityBeforeChange.contact.address.prefectureEn !== facilitySections.prefectureEn
-        || facilityBeforeChange.contact.address.cityEn !== facilitySections.cityEn
-        || facilityBeforeChange.contact.address.addressLine1En !== facilitySections.addressLine1En
-        || facilityBeforeChange.contact.address.addressLine2En !== facilitySections.addressLine2En
-        || facilityBeforeChange.contact.address.prefectureJa !== facilitySections.prefectureJa
-        || facilityBeforeChange.contact.address.cityJa !== facilitySections.cityJa
-        || facilityBeforeChange.contact.address.addressLine1Ja !== facilitySections.addressLine1Ja
-        || facilityBeforeChange.contact.address.addressLine2Ja !== facilitySections.addressLine2Ja
-        || facilityBeforeChange.contact.googleMapsUrl !== facilitySections.googlemapsURL
-        || facilityBeforeChange.mapLatitude.toString() !== facilitySections.mapLatitude
-        || facilityBeforeChange.mapLongitude.toString() !== facilitySections.mapLongitude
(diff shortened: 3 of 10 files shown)


theyokohamalife:
Nice work so far! ✨ 

Ideally, updating a major package such as Nuxt should be done in its own separate PR. Aside from yarn.lock, nuxt.config, and the declaration file, are any of the other changes reliant on us switching from `Nuxt 3.12.4` to `Nuxt 3.15.2`? If not, I would suggest removing them and sticking to the upgrade. If we need to roll it back for whatever reason, we wouldn't want to undo all of the other changes along with it.

That said, it appears that everything else is linting. 🧹  Were those made by running `yarn lint:fix`? If not, it may be a different formatter you have configured, and they shouldn't be checked in.

Perhaps I missed something, but when looking over the code I didn't see any changes specifically relating to i18n. I think a more accurate PR title would be something along the lines of `feat: upgrade Nuxt to 3.15.2`.

theyokohamalife (changes requested):
Nice work so far! ✨

Ideally, updating a major package such as Nuxt should be done in its own separate PR. Aside from yarn.lock, nuxt.config, and the declaration file, are any of the other changes reliant on us switching from Nuxt 3.12.4 to Nuxt 3.15.2? If not, I would suggest removing them and sticking to the upgrade. If we need to roll it back for whatever reason, we wouldn't want to undo all of the other changes along with it.

That said, it appears that everything else is linting. 🧹 Were those made by running yarn lint:fix? If not, it may be a different formatter you have configured, and they shouldn't be checked in.

Perhaps I missed something, but when looking over the code I didn't see any changes specifically relating to i18n. I think a more accurate PR title would be something along the lines of feat: upgrade Nuxt to 3.15.2.

gminetoma:
@theyokohamalife Thank you!

This is all related to updating Nuxt, and I have run `yarn lint:fix`.

[Nuxt v3.13 update reference](https://nuxt.com/blog/v3-13)

[Nuxt v3.15 update reference](https://nuxt.com/blog/v3-15)

theyokohamalife:
> @theyokohamalife Thank you!
> 
> This is all related to updating Nuxt, and I have run `yarn lint:fix`.
> 
> [Nuxt v3.13 update reference](https://nuxt.com/blog/v3-13)
> 
> [Nuxt v3.15 update reference](https://nuxt.com/blog/v3-15)

If that is the case then the first commit should be all of the upgrade-related changes. That includes the change you have to nuxt.config where you added `compatibilityDate: '2025-01-17'`. The commit message itself should probably be reworded from `fix: correct i18n translation function type` to something like `feat: upgrade Nuxt and update nuxt.config`. 

The linting changes _shouldn't be related_ to upgrading Nuxt—however, @ermish and I have differing opinions on whether to leave unrelated fixes in or not. If you want to keep them in, I suggest rewording the 2nd commit message from `chore: format code and update Nuxt configuration` to something like `chore: format code with linter`.

theyokohamalife:
LGTM  😎  but @ermish will likely also want to take a look at it since we're upgrading Nuxt.

nuxt.config.js
```diff
@@ -162,5 +162,6 @@ export default defineNuxtConfig({
             NUXT_USE_LOCAL_API: process.env.NUXT_USE_LOCAL_API
         }
     },
-    telemetry: false
+    telemetry: false,
+    compatibilityDate: '2025-01-17'
```
ermish, line 166: Let's add some comments around this so people can understand what it does for us

declarations.d.ts
```diff
@@ -0,0 +1,11 @@
+/* eslint-disable @typescript-eslint/no-empty-object-type */
+
+import type {
```
ermish, line 3: Let's add some comments around this so people can understand what it does for us. 
Especially so we know it's a temporary hack until it's upgraded

ermish (commented):
Added comment comments but otherwise looks good! Nice investigation 💪
I agree with @theyokohamalife here and actually think we should remove the linter fixes from this PR because this might we one we want to rollback as a whole if we have issues. But we wouldn't want to rollback the linter fixes.

nuxt.config.js
gminetoma: Do I really need to add a comment here? If you hover the mouse over `compatibilityDate`, the explanation will pop up.

image

gminetoma: I'm adding a comment anyway.

ermish: this is a good lesson point about comments! 
Comments that just say "what" something does are a waste of time. "Why" is where comments have value.
So here, just explaining compatibilityDate isn't relevant and I agree. 

But questions that are relevant to answer "why": Why we actually need to set it and when can I delete it? When should I update it?

ermish: I'm actually wondering if we can remove it now. Do we actually need to set it?

gminetoma: It's generally a good practice to set the `compatibilityDate` after a Nuxt upgrade. This property controls the behavior of some Nuxt modules and Nitro as of the specified date. By doing so, we can ensure that minor updates to these modules won't unexpectedly break our app (which is a possibility).

I didn't find many references to this beyond the Nuxt documentation, but this is what I understood based on it.

ermish (approved):
Looks good! Thanks for looking into it!

ermish:
@gminetoma looks like the types are failing CI now. Can you take a look?

ermish:
At a high level, @gminetoma whats the reason we wanted to upgrade nuxt? 
I'm fine with merging, but ina. Bigger port I would question why the upgrade is worth having to do a hack for vs just waiting for a better version update

gminetoma:
> @gminetoma looks like the types are failing CI now. Can you take a look?

You asked me to remove the linting commit, and that’s why it’s failing now.

> At a high level, @gminetoma whats the reason we wanted to upgrade nuxt?
I'm fine with merging, but in bigger project I would question why the upgrade is worth having to do a hack for vs just waiting for a better version update

This isn’t really a Nuxt problem at this stage. After Nuxt’s update, they’ve switched to the correct namespace, so the issues we were encountering before due to conflicts with older versions of Nuxt have been resolved. The current problem lies with older packages that are not yet following the Vue recommendations. The hack is not for Nuxt, but for ensuring compatibility with those outdated packages.

We were having problems, so we are solving them instead of waiting for someone to have the good will to do it.

ermish:
@gminetoma hmmm I'm not sure why, but the CI is only failing in this branch. Those linting errors aren't happening on other ones. I'm guessing the upgrade has new linting rules?

ermish:
hmm I tried triggering it again. If it's part of the upgrade, then it makes sense to include them

gminetoma:
@ermish 

It's strange I didn't touch the eslint file and if we go to the main branch and open one file mentioned at the print above, we can clearly see the eslint error. But if we run yarn lint there are no errors.

ermish (approved):
good to go! nice work figuring out the upgrade @gminetoma  🥳 
I think the linting is because the nuxt upgrade likely has eslint config changes that we inherit from which aren't very visible

ermish:
@theyokohamalife Can you re-review/approve?

theyokohamalife (approved):
LGTM

PR #971, opened 2025-01-17
fix: nuxt configuration file

### LINKED ISSUE #970: Fix: Nuxt configuration file
# Problem
The `nuxt.config.js` file, when written as a JavaScript file, does not throw any errors if the configuration object is incorrect. According to the Nuxt documentation, the current configuration object is set up incorrectly.

## Example

```TypeScript
export default defineNuxtConfig({
    //typescript settings
    typescript: {
        shim: false,
        tsConfig: { compilerOptions: { baseUrl: '.' } }
    },
    // Global page headers: https://go.nuxtjs.dev/config-head
    head: {
        titleTemplate: 'Health Services in Japan',
        title: SITE_TITLE,
        htmlAttrs: {
            lang: 'en'
        },
        meta: [
            {
                name: 'google-site-verification',
                content: '-iHcqQwxR-wiZU6Y4BKtcn79tUZJAcyWXJt7w7iobpY'
            },
            { charset: 'utf-8' },
            { name: 'viewport', content: 'width=device-width, initial-scale=1' },
            { hid: 'description', name: 'description', content: SITE_DESCRIPTION },
            { name: 'format-detection', content: 'telephone=no' },
...
```

According to the Nuxt documentation, the `head` property should be placed inside the `app` property. This is critical for our SEO configuration, and the current setup is negatively impacting our app. As shown below, there are no meta tags inside the `<head>` section.

Image

Image

## Reference

[https://nuxt.com/docs/api/nuxt-config](https://nuxt.com/docs/api/nuxt-config)

[https://nuxt.com/docs/getting-started/seo-meta](https://nuxt.com/docs/getting-started/seo-meta)

# Success
The file should have a `.ts` file extension, and the object should be properly formatted according to Nuxt's documentation.

### Requirements

- [x] Display the meta tags
- [x] `.ts` file extension

gminetoma:
Resolves #970 

## 🔧 What changed

- `nuxt.config.js` to `nuxt.config.ts`
- Solve app metadata
- Centralize TypeScript configuration in `nuxt.config.ts`
- Remove unnecessary Pinia `autoImports` configuration
- Remove invalid `target` property
- Remove `typescript.shim` (default is already `false`)

## 🔍 Context

The app metadata is now correctly placed in our build.

Nuxt generates the TypeScript configuration automatically, so
maintaining a custom `tsconfig.json` can lead to conflicts with Nuxt's
generated configuration, breaking features like auto-imports.

To ensure maximum compatibility, our `tsconfig.json` now only imports
the Nuxt-generated config file. Additionally, we are using the
`tsConfig` property in `nuxt.config.ts` to add our custom
configuration directly.

This change resolves an issue where Nuxt was struggling to auto-import
components and methods from modules. It also significantly reduces the
number of imports required in our files.

With this update, explicit imports are no longer needed for:
- Components
- Stores
- Utilities
- Vue Composition API (e.g., `ref`, `watch`, etc.)
- And more...

## 📖 References
https://nuxt.com/docs/api/nuxt-config
https://nuxt.com/docs/guide/concepts/typescript
https://nuxt.com/docs/guide/directory-structure/tsconfig
https://pinia.vuejs.org/api/@pinia/nuxt/interfaces/ModuleOptions.html#ModuleOptions
https://nuxt.com/docs/guide/concepts/auto-imports

## 📸 Screenshots

-   ### Before
Image

image

image

-   ### After
image

image

image


nuxt.config.ts
```diff
@@ -0,0 +1,170 @@
+import { defineNuxtConfig } from 'nuxt/config'
+
+const SITE_TITLE = 'Find a Doc, Japan!'
+const SITE_DESCRIPTION
+    = 'Health service information for the international community in Japan'
+
+export default defineNuxtConfig({
+    //typescript settings
+    typescript: {
```
ermish, line 9: I prefer we keep our tsconfig external from nuxt config for a few reasons:
- Better separation of concerns 
- We may move away from nuxt someday 
- We will lose typescript IDE intellisense around the config
- Other builders look for a tsconfig which we may move to (such as Bun for example)

ermish (changes requested):
Love the TS upgrade and fixes! 
Have one comment around tsconfig to look at

nuxt.config.ts
ermish: image
Here's how to have nuxt import our tsconfig instead

gminetoma: @ermish 

I understand the concern about not having our own `tsconfig.json`. I like the idea of separation of concerns and not being tied to Nuxt. However, the problem is that Nuxt creates its own `tsconfig.json` with import configurations and other settings. We can't extend this file into our `tsconfig` and import it inside the Nuxt configuration (circular dependency). 

So, we must decide which one to use. If we use our own file, we will lose all the TypeScript linting (the original problem). What we can do is manually copy the Nuxt `tsconfig` into ours.

gminetoma: Okay, I found the root cause of the TypeScript linting issue. Our `tsconfig` "include" was overwriting Nuxt's "include". We can resolve this by removing the conflicting part and sticking to our configuration file. Additionally, we don't need to import this file into `nuxt.config`.

gminetoma: @ermish 

The way I did now it's working. But I removed the included and exclude paths.

tsconfig.json
```diff
@@ -1,31 +1,116 @@
...
-        "utils/**/*",
-        "__tests__/*",
-        "typesgeneratorconfig.ts",
-    ],
-    "exclude": [
-        "node_modules"
-    ]
+  "extends": "./.nuxt/tsconfig.json",
+  "compilerOptions": {
+    /* Visit https://aka.ms/tsconfig to read more about this file */
```
ermish, line 4: Can we remove all the commented out parameters that aren't being set?

ermish (approved):
looks great! thanks for the improvements!

PR #975, opened 2025-01-18
refactor: change modal div to use v-if

theyokohamalife:
## 🔧 What changed
The `data-testid="submission-form-modal"` modal is contained in a div with a `v-show` attribute. This is causing CI to occasionally fail, because the test was not expecting the modal to appear in the DOM. This PR changes the `v-show` to `v-if` to prevent it from existing.

## 🧪 Testing instructions
1. Run the local dev server with `yarn dev`
2. Run `yarn cypress open` to start Cypress in the browser
3. Select the E2E tests, and select the moderationDashboard.cy.ts file
4. Confirm that the `should not display modal if user navigates back without making changes` passes


NabbeunNabi (approved):
💯  that mages sense why the test was flakey

theyokohamalife:
@NabbeunNabi I updated the PR description, because the changes in your PR actually came first. 🙇🏾‍♀️  I couldn't use `cy.get('[data-testid="submission-form-modal"]').should('not.be.visible')` because it would fail when the .get couldn't find the modal. 🤔 This change just works around that.

ermish (approved):
LGTM 👍

PR #981, opened 2025-01-24
refactor: extract login from Cypress tests

### LINKED ISSUE #797: Refactor the login on moderation into the `cypress.config.ts`
# Problem
Currently the login process in the tests for the Moderation Panel do not follow the recommended structure in the [Cypress Documentation](https://docs.cypress.io/guides/end-to-end-testing/auth0-authentication) for using OAuth. This could be slowing down our tests and it would be better to use the method in the documentation above.

# Success
The login process is refactored from our current usage in the tests to the usage in the documentation listed above. It is then used in the tests with all the tests for the moderation dashboard passing
### Requirements

- [x] Refactor login process according to the [Cypress Documentation](https://docs.cypress.io/guides/end-to-end-testing/auth0-authentication) 
- [x] Update the tests for the moderation dashboard using the refactored login process
- [x] Have all the tests passing with this new process

gminetoma:
Resolves #797

### ✨ Linting
Fixed Cypress linting

### 🧪 Test Environment
Added Yarn support for `.env.test` to securely manage environment
variables required for testing. The `.env.test` file includes:
- `AUTH0_DOMAIN`
- `AUTH0_USERNAME`
- `AUTH0_PASSWORD`

These variables are critical for the Auth0 login process. To ensure
security, only authorized individuals will have access to the
`.env.test` file, and it has been added to `.gitignore` to prevent
accidental commits of sensitive information.

### 🤖 Cypress
Add Cypress cy.login() custom command
- Introduce an utils file with the following functions:
  - `auth0Login`: Handles the Auth0 login routine
  - `hasOperation`: Checks the request query for a specific operation
  - `aliasQuery`: Adds an alias for an operation and stubs it
- Change file extensions to `.ts`
- Add Cypress plugin types to `tsconfig.json`
- Import Cypress plugins in `e2e.ts`

This introduces a custom `cy.login()` command for Cypress that
uses the Auth0 login routine to create a login session. This optimizes
testing by reducing the need for repetitive logins during tests.

Additionally, it adds utility functions to simplify the manipulation of
GraphQL requests. Furthermore, Cypress plugins are now imported in a
centralized `e2e.ts` file, eliminating the need to import them
individually in each test file.

### 📝 Cypress Tests
Simplify login process and enhance cypress.intercept usage
- Remove unnecessary plugin imports
- Directly import JSON fake data
- Use the new `cy.login()` command
- Separate each test suite into a different spec file

This refactor streamlines the login process for tests, improving both
speed and efficiency. Instead of using fixtures for fake data, JSON
files are now directly imported, as fixtures are unnecessary due to how
GraphQL requests are manipulated.

Cypress plugin imports have been centralized in the support file,
removing the need for redundant imports in individual test files.

Additionally, test suites have been separated into different spec files
for better organization and maintainability.

### 🎬 GitHub Actions
Updated the configuration to ensure GitHub Actions runs E2E tests 
using our custom command (`yarn test:e2e`) and GitHub secrets. 
This guarantees that the tests follow our defined configurations, 
ensuring consistency between local development and CI environments.

## 📑 References:
https://www.npmjs.com/package/eslint-plugin-cypress
https://www.npmjs.com/package/cypress-plugin-tab
https://docs.cypress.io/app/guides/network-requests#Working-with-GraphQL
https://docs.cypress.io/app/guides/authentication-testing/auth0-authentication
https://github.com/dmtrKovalenko/cypress-real-events#readme
https://github.com/cypress-io/github-action

### DIFF
--- a/.yarnrc.yml
+++ b/.yarnrc.yml
@@ -5,3 +5,6 @@ enableGlobalCache: false
 nodeLinker: node-modules
 
 yarnPath: .yarn/releases/yarn-4.4.0.cjs
+
+injectEnvironmentFiles:
+  - .env.test?
--- a/.yarnrc.yml
+++ b/.yarnrc.yml
@@ -5,6 +5,3 @@ enableGlobalCache: false
 nodeLinker: node-modules
 
 yarnPath: .yarn/releases/yarn-4.4.0.cjs
-
-injectEnvironmentFiles:
-  - .env.test?
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -2,7 +2,7 @@ import globals from 'globals'
 import eslintJsPlugin from '@eslint/js'
 import tseslint from 'typescript-eslint'
 import stylistic from '@stylistic/eslint-plugin'
-import pluginCypress from 'eslint-plugin-cypress'
+import pluginCypress from 'eslint-plugin-cypress/flat'
 import pluginVue from 'eslint-plugin-vue'
 import withNuxt from './.nuxt/eslint.config.mjs'
 
@@ -14,7 +14,6 @@ export default withNuxt(
             '.output/*',
             '.nuxt/*',
             'coverage/*',
-            'cypress/*',
             '.yarn/*',
             'typedefs/gqlTypes.ts'
         ]
@@ -36,7 +35,7 @@ export default withNuxt(
             '@typescript-eslint': tseslint.plugin,
             '@stylistic': stylistic
         },
-        ignores: ['./typeDefs/gqlTypes.ts', './typesgeneratorconfig.ts', 'cypress/**/*'],
+        ignores: ['./typeDefs/gqlTypes.ts', './typesgeneratorconfig.ts'],
         rules: {
             ...tseslint.configs.recommended,
             // TS specific rules
@@ -94,32 +93,12 @@ export default withNuxt(
             'vue/html-indent': ['error', 4]
         }
     },
-    // Linting for tests (cypress + pinia)
+    // Linting for Cypress (https://www.npmjs.com/package/eslint-plugin-cypress)
     {
-        languageOptions: {
-            globals: {
-                cy: true,
-                it: true,
-                describe: true,
-                context: true,
-                beforeEach: true,
-                before: true
-            }
-        },
-        files: ['cypress/e2e/*.ts', 'test/**/*'],
-        plugins: {
-            cypress: pluginCypress
-        },
+        files: ['test/cypress/**/*'],
+        ...pluginCypress.configs.recommended,
         rules: {
-            //this is to support chai chaining syntax
-            '@typescript-eslint/no-unused-expressions': 'off'
-        //     'cypress/no-assigning-return-values": "error",
-        //     "cypress/no-unnecessary-waiting": "error",
-        //     "cypress/assertion-before-screenshot": "warn",
-        //     "cypress/no-force": "warn",
-        //     "cypress/no-async-tests": "error",
-        //     "cypress/no-async-before": "error",
-        //     "cypress/no-pause": "error"
+            'cypress/no-unnecessary-waiting': 'off'
         }
     }
 )
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -15,6 +15,7 @@ export default withNuxt(
             '.nuxt/*',
             'coverage/*',
             '.yarn/*',
+            'test/vitest',
             'typedefs/gqlTypes.ts'
         ]
     },
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
         "test": "yarn test:pinia --run && yarn test:e2e",
         "test:pinia": "nuxi prepare && nuxi generate && vitest",
         "test:pinia:coverage": "vitest run --coverage",
-        "test:e2e": "cypress run --e2e --browser chrome --record --key 04a1fe54-0011-407b-b7d7-fcb9b2daf1e6"
+        "test:e2e": "cypress run --e2e"
     },
     "packageManager": "yarn@4.4.0",
     "engines": {
--- a/package.json
+++ b/package.json
@@ -17,11 +17,8 @@
         "test": "yarn test:pinia --run && yarn test:e2e",
         "test:pinia": "nuxi prepare && nuxi generate && vitest",
         "test:pinia:coverage": "vitest run --coverage",
-        "test:e2e": "NODE_ENV=test dotenv -e .env.test -o -- yarn test:e2e:start",
-        "test:e2e:start": "yarn test:env:echo && cypress run --e2e",
-        "test:e2e:open": "NODE_ENV=test dotenv -e .env.test -o -- yarn test:e2e:open:start",
-        "test:e2e:open:start": "yarn test:env:echo && cypress open",
-        "test:env:echo": "echo \n 🧪 $NODE_ENV environment loaded 🧪"
+        "test:e2e": "dotenv -e .env.test -o -- cypress open",
+        "test:e2e:ci": "dotenv -e .env.test -o -- cypress run --e2e"
     },
     "packageManager": "yarn@4.4.0",
     "engines": {
--- a/package.json
+++ b/package.json
@@ -17,7 +17,11 @@
         "test": "yarn test:pinia --run && yarn test:e2e",
         "test:pinia": "nuxi prepare && nuxi generate && vitest",
         "test:pinia:coverage": "vitest run --coverage",
-        "test:e2e": "cypress run --e2e"
+        "test:e2e": "NODE_ENV=test dotenv -e .env.test -o -- yarn test:e2e:start",
+        "test:e2e:start": "yarn test:env:echo && cypress run --e2e",
+        "test:e2e:open": "NODE_ENV=test dotenv -e .env.test -o -- yarn test:e2e:open:start",
+        "test:e2e:open:start": "yarn test:env:echo && cypress open",
+        "test:env:echo": "echo \n 🧪 $NODE_ENV environment loaded 🧪"
     },
     "packageManager": "yarn@4.4.0",
     "engines": {
@@ -59,6 +63,7 @@
         "autoprefixer": "^10.4.20",
         "cypress": "^13.13.2",
         "cypress-plugin-tab": "^1.0.5",
+        "dotenv-cli": "^8.0.0",
         "eslint": "^9.8.0",
         "eslint-plugin-cypress": "^3.4.0",
         "eslint-plugin-json": "^4.0.1",
--- a/test/cypress/support/commands.ts
+++ b/test/cypress/support/commands.ts
@@ -36,7 +36,7 @@ Cypress.Commands.add('login', () => {
         autoEnd: false
     })
 
-    log.snapshot('before')
+    log.snapshot('Connecting to auth0')
 
     cy.session(`auth0-${auth0UserName}`, () => {
         auth0Login()
@@ -49,6 +49,6 @@ Cypress.Commands.add('login', () => {
         cacheAcrossSpecs: true
     })
 
-    log.snapshot('after')
+    log.snapshot('Successfully logged in')
     log.end()
 })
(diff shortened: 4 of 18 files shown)


gminetoma:
@NabbeunNabi

.env.test
```diff
@@ -0,0 +1,3 @@
+AUTH0_DOMAIN="https://findadoc.jp.auth0.com/"
+AUTH0_USERNAME="findadoctest@proton.me"
+AUTH0_PASSWORD="vCnL5J8agHg6m2f"
```
NabbeunNabi, line 3: @gminetoma Did you mean to push this?

.yarnrc.yml
```diff
@@ -5,3 +5,6 @@ enableGlobalCache: false
 nodeLinker: node-modules
 
 yarnPath: .yarn/releases/yarn-4.4.0.cjs
+
+injectEnvironmentFiles:
+  - .env.test?
```
NabbeunNabi, line 10: Should this be changed to github secrets for the CI/CD?

NabbeunNabi (commented):
@gminetoma great work. Just a few initial questions

.yarnrc.yml
gminetoma: The way it is now (with the env file and the Cypress action config) is working because yarn is loading the env variables. This should be changed if don't have the env file anymore.

.env.test
gminetoma: Yes. We still need the variables somewhere to access the moderation page during the tests. The only way to hide this is to stub the login processes.

.yarnrc.yml
ermish: I feel like there's a better option here like importing through the yarn command in package.json or through the yaml file 🤔

gminetoma: @ermish
We could install a package like [dotenv](https://www.npmjs.com/package/dotenv) or [cross-env](https://www.npmjs.com/package/cross-env) to set the env file in a yarn command.

I think we can't use the yaml file because we still want to load the variables during the development while creating tests.

ermish (changes requested):

.yarnrc.yml
ermish: Those are fine to add 👍 
we're actively using both of them in our projects already

test/cypress/utils.ts
```diff
@@ -0,0 +1,51 @@
...
+// https://docs.cypress.io/api/commands/intercept#Interception-lifecycle
+export const requestHandler = (req: IncomingHttpRequest) => {
+    req.on('before:response', res => {
+        // force all API responses to not be cached
+        res.headers['cache-control'] = 'no-store'
+    })
+}
+
+// https://docs.cypress.io/app/guides/network-requests#Working-with-GraphQL
+export const aliasQuery = (req: IncomingHttpRequest, operation: string, responseBody: unknown) => {
```
ermish, line 32: What is this for? I can't tell from quickly reading it

test/cypress/utils.ts
```diff
@@ -0,0 +1,51 @@
...
+
+    req.alias = operation
+
+    req.reply({
+        statusCode: 200,
+        body: responseBody
+    })
+}
+
+export const hasOperation = (req: IncomingHttpRequest, operation) => {
```
ermish, line 48: Why do we need this?

test/cypress/utils.ts
```diff
@@ -0,0 +1,51 @@
...
+        cy.get('[data-action-button-primary]').click()
+    })
+
+    const baseUrl = `${Cypress.config().baseUrl}/`
+
+    cy.url().should('equal', baseUrl)
+}
+
+// https://docs.cypress.io/api/commands/intercept#Interception-lifecycle
+export const requestHandler = (req: IncomingHttpRequest) => {
```
ermish, line 24: Since this is only doing one purpose of clearing the cache, that would be a better function name here

test/cypress/support/index.d.ts
```diff
@@ -1,10 +1,7 @@
-export { }
+/// <reference types="cypress" />
 
-declare global {
-    namespace Cypress {
-        interface Chainable {
-            isInViewport(args?: string): Chainable<JQuery<HTMLElement>>
-            isNotInViewport(args?: string): Chainable<JQuery<HTMLElement>>
-        }
+declare namespace Cypress{
```
ermish, line 3: Can this type come from cypress types or an @types/cypress library?

test/cypress/support/commands.ts
```diff
@@ -23,3 +23,32 @@
...
+Cypress.Commands.add('login', () => {
+    const auth0UserName = Cypress.env('AUTH0_USERNAME')
+
+    const log = Cypress.log({
+        displayName: 'AUTH0_LOGIN',
+        message: [`🔐 Authenticating | ${auth0UserName}`],
+        autoEnd: false
+    })
+
+    log.snapshot('before')
```
ermish, line 39: Instead of "before" and "after", can be more explicit like
 "connecting to auth0"
And 
"Successfully logged in"

eslint.config.js
```diff
@@ -94,32 +94,12 @@ export default withNuxt(
...
-                beforeEach: true,
-                before: true
-            }
-        },
-        files: ['cypress/e2e/*.ts', 'test/**/*'],
-        plugins: {
-            cypress: pluginCypress
-        },
+        files: ['test/cypress/**/*'],
+        ...pluginCypress.configs.recommended,
```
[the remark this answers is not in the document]
ermish: Please confirm on this one

ermish (changes requested):
Added some comments

eslint.config.js
```diff
@@ -94,32 +94,12 @@ export default withNuxt(
             'vue/html-indent': ['error', 4]
         }
     },
-    // Linting for tests (cypress + pinia)
+    // Linting for Cypress (https://www.npmjs.com/package/eslint-plugin-cypress)
```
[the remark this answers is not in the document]
gminetoma: Currently, for Pinia tests, we are using Vitest. I'm already addressing its rules here:  
Issues: #982  
Commit: [commit](https://github.com/ourjapanlife/findadoc-web/commit/1456d6b6120e09e440abed371197b16472e7b00c)

eslint.config.js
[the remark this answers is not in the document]
gminetoma: We are not using any rules for Cypress right now. We are using the plugin only for the globals. We are overriding the rules as you can see there.

I would recommend activating the rules since they follow Cypress best practices. I tested the recommended rules, and we will have a lot to refactor if we decide to activate them.

Here are the plugin's recommended options:
```
Object.assign(plugin.configs, {
  recommended: {
    name: 'cypress/recommended',
    plugins: {
      cypress: plugin
    },
    rules: {
      'cypress/no-assigning-return-values': 'error',
      'cypress/no-unnecessary-waiting': 'error',
      'cypress/no-async-tests': 'error',
      'cypress/unsafe-to-chain-command': 'error',
    },
    languageOptions: {
      globals:
        commonGlobals,
      ...commonLanguageOptions
    }
  }
})
```

[Cypress recommended rules](https://www.npmjs.com/package/eslint-plugin-cypress#rules)  
[Cypress best practices](https://docs.cypress.io/app/core-concepts/best-practices)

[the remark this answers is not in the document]
ermish: I agree we should activate them 👍 let's do it in a separate PR though with the corresponding changes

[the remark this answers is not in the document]
ermish: I was thinking it would add the rules in addition to ours but I realized it's the config section and not the rules set

test/cypress/support/commands.ts
gminetoma: I don't know if you meant to ask this, but those snapshots are only for debugging.

To me, it doesn’t make much sense to do it this way. 'Before' the auth and 'After' makes more sense to me.

If this is your intention, I'll do it your way.

image

test/cypress/support/index.d.ts
gminetoma: We are using the Cypress namespace, and we don't have `@types/cypress`. We are extending the Cypress namespace with our custom command.  

[Types for custom commands](https://docs.cypress.io/app/tooling/typescript-support#Types-for-Custom-Commands)  

Sorry, I think I didn't understand your question. Could you elaborate more, please?

test/cypress/utils.ts
gminetoma: Check if the GraphQL operation is included in the request body

test/cypress/utils.ts
gminetoma: Set the alias for a request based on the operation name. After setting it, we can use cy.wait(@{alias}).

I'm following the [Cypress documentation](https://docs.cypress.io/app/guides/network-requests#Working-with-GraphQL) for working with GraphQL.

eslint.config.js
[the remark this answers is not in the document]
gminetoma: Do you want me to move all the linting part to another PR? Should I remove the commit?

test/cypress/utils.ts
gminetoma:
```
const preventRequestCache = (req: IncomingHttpRequest) => {
    req.on('before:response', res => {
        // force all API responses to not be cached
        res.headers['cache-control'] = 'no-store'
    })
}
```

test/cypress/support/commands.ts
gminetoma: @ermish  

Is your thumbs-up to proceed your way, or should we continue using Before and After? 😅

ermish: I like the "upgraded" one 😆 
It's just for logging, but we can have other logging messages that aren't just for auth0 so it's good to make it more specific

eslint.config.js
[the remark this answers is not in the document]
ermish: you got it. Looks good now, and yes let's do do a separate PR for the linting

test/cypress/utils.ts
ermish: thanks that helps

test/cypress/utils.ts
ermish: perfect 👌

package.json
```diff
@@ -17,7 +17,11 @@
         "test": "yarn test:pinia --run && yarn test:e2e",
         "test:pinia": "nuxi prepare && nuxi generate && vitest",
         "test:pinia:coverage": "vitest run --coverage",
-        "test:e2e": "cypress run --e2e --browser chrome --record --key 04a1fe54-0011-407b-b7d7-fcb9b2daf1e6"
+        "test:e2e": "NODE_ENV=test dotenv -e .env.test -o -- yarn test:e2e:start",
+        "test:e2e:start": "yarn test:env:echo && cypress run --e2e",
```
ermish, line 21: someone of a personal preference here, but I think we should avoid "subscripts" where possible for a few reasons:
-  While the reusability and separation is nice, there's no way to look at the list of scripts and know which ones should be run vs not.
- Too many scripts. When someone is new to the project, it can be intimidating. Especially when there's several that look almost the same. `e2e:start` vs `e2e` vs `e2e:open` for example
- Ideally (where possible) there should only be one script to start something. So basically I would combine `test:e2e` + `test:e2e:start` + `test:e2e:open` + `test:e2e:open:start`+`test:env:echo` all into a single `test:e2e` script. I get it can be long, but I think that's ok. It only matters when someone is trying to edit the script.

gminetoma: @ermish 

What do you think this way?:

```suggestion
        "test:e2e": "NODE_ENV=test dotenv -e .env.test -o -- bash -c 'echo 🧪 $NODE_ENV environment loaded 🧪 && cypress run --e2e'",
        "test:e2e:open": "NODE_ENV=test dotenv -e .env.test -o -- bash -c 'echo 🧪 $NODE_ENV environment loaded 🧪 && cypress open'"
```

ermish:
@gminetoma I just noticed the `yarn test:e2e:open` fails in windows
image
Bash is specific to unix. On the server side, we console with something like this `echo '🧶 Installing yarn dependencies...🧶'`

package.json
ermish: after talking in slack, we've updated the commands

ermish (approved):
Looks good now! Super excited for this one! Going to simplify our test flows

Issue #1006, opened 2025-02-06
Add the `/cypress` folder to the `.gitignore`


NabbeunNabi:
# Problem
When people run tests there are screenshots and logs added to this file. We do not want these files accidentally added and pushed to branches and then possibly merged to main.

# Success
The folder `/cypress` is added to the `.gitignore`

### Requirements

- [x] Add the file to the gitignore
- [x] Run tests that make them fail to make sure these files in the folder are not tracked

PR #1005, opened 2025-02-06
fix: add cypress file to gitignore file

Anissa3005:
Resolves #1006 
- [x] PR title follows [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/#summary) specs

## 🔧 What changed

This PR adds the Cypress file to .gitignore to prevent test snapshots from being unintentionally committed when tests fail.

## 🧪 Testing instructions

NA

PR #1007, opened 2025-02-06
fix: fix-flaky-test-in-CI/CD

### LINKED ISSUE #1000: Research and fix flaky test in CI/CD
# Problem
There is a test that occasionally is flaky and fails in CI/CD. It is hard to replicate locally so will require some research, but when it does fail it is the following test.

Test location --> [click here](https://github.com/ourjapanlife/findadoc-web/blob/d52fa4c25eabfcdd3a0742aade5bec9b5702de3c/test/cypress/e2e/moderationDashboard.cy.ts#L407)

Image

# Success
The reason for the test is found and explained well within the PR description and the flaky test is fixed
### Requirements

- [ ] The cause of the flaky test is research and understood
- [ ] The fix of the flaky test is implemented
- [ ] A clear explanation is given in the PR description

# Implementation notes
It would be good to understand how cypress works. I would look through the configuration files and the tests contained with the following folder:

[Cypress folder](https://github.com/ourjapanlife/findadoc-web/tree/main/test/cypress)

And read up on cypress especially the assertions being used in the [Cypress Documentation](https://docs.cypress.io/app/get-started/why-cypress)

prestonb91:
Resolves #1000 
- [ ] PR title follows [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/#summary) specs

## 📸 Screenshots

-   ### Before

Error found by Will shows that exiting out of editing without any changes sometimes still shows confirmation modal exists. 
Tests pass locally in headless mode and GUI. 

Potential causes and solutions to try:
-  DOM may be taking time to load or tests sometimes takes longer in GUI vs headless mode, so if flaky test occurs in GUI, test itself potentially could be OK just taking time to run so fix by including a timeout. Unlikely though if error occurred in headless mode, also not passing even after 10sec timeout
- may have similar issue of DOM elements persisting when should not due to cypress version [link](https://github.com/cypress-io/cypress/issues/25363) to look into
- adding a click in the test to simulate closing the modal [link](https://stackoverflow.com/questions/75161891/cypress-modal-doesnt-close)

-   ### After


theyokohamalife:
@prestonb91 I know this is still a draft, but I just wanted to comment that we don't want to check in those screenshots. It might be a good idea to add `cypress/screenshots` and `cypress/videos` to .gitignore. Maybe `cypress/downloads` too. That could be done in a separate PR that gets merged before this one.

NabbeunNabi:
> @prestonb91 I know this is still a draft, but I just wanted to comment that we don't want to check in those screenshots. It might be a good idea to add `cypress/screenshots` and `cypress/videos` to .gitignore. Maybe `cypress/downloads` too. That could be done in a separate PR that gets merged before this one.

Actually this was just addressed yesterday in a ticket where @Anissa3005 added that folder to the ginignore! 🎆 🍾

nuxt.config.ts
```diff
@@ -5,153 +5,160 @@ const SITE_DESCRIPTION
...
-            GOOGLE_MAPS_API_KEY: process.env.GOOGLE_MAPS_API_KEY,
-
-            ENABLE_MODERATION_PANEL: process.env.ENABLE_MODERATION_PANEL,
-
-            NUXT_USE_LOCAL_API: process.env.NUXT_USE_LOCAL_API
-        }
-    },
-    telemetry: false
-})
+  app: {
```
theyokohamalife, line 8: It's hard to tell what has actually been changed in this file, because the indentation has been changed, which makes it appear as if the whole file has been changed.

theyokohamalife:
> > @prestonb91 I know this is still a draft, but I just wanted to comment that we don't want to check in those screenshots. It might be a good idea to add `cypress/screenshots` and `cypress/videos` to .gitignore. Maybe `cypress/downloads` too. That could be done in a separate PR that gets merged before this one.
> 
> Actually this was just addressed yesterday in a ticket where @Anissa3005 added that folder to the ginignore! 🎆 🍾

Nice! Link to ticket?

gminetoma:
@theyokohamalife

The ticket: https://github.com/ourjapanlife/findadoc-web/pull/1005

prestonb91:
> @prestonb91 I know this is still a draft, but I just wanted to comment that we don't want to check in those screenshots. It might be a good idea to add `cypress/screenshots` and `cypress/videos` to .gitignore. Maybe `cypress/downloads` too. That could be done in a separate PR that gets merged before this one.

@theyokohamalife, understood about this for future!

NabbeunNabi:
> @theyokohamalife
> 
> The ticket: #1005

That is the PR: #1006 is the issue

NabbeunNabi:
@gminetoma @prestonb91 It happened!

It's go time

image

PR #1012, opened 2025-02-07
feat: add loader to ModEdithealthcareProfessionalSection component

### LINKED ISSUE #954: feat: show loading placeholder in ModEditHealthcareProfessionalSection
# Problem
We currently have no loading placeholder in the `ModEditHealthcareProfessionalSection.vue`. We do have a loading component, but it isn't in use when the data is being loaded in this section. 

# Success
The loader is shown on the page load for the path `moderation/edit-facility/:id`

### Requirements

- [x] The loader should appear when data is being loaded. 
- [ ] There should be no visual disruption when transitioning in/out. 
- [ ] It should make the experience of using the section clearer for the user. 
- [ ] A video is included in the PR to show the desired behavior

# Implementation notes
You can see in the [MainContentContainer.vue](https://github.com/ourjapanlife/findadoc-web/blob/main/components/MainContentContainer.vue) where the loader is being used.A

ConwayDecker:
Resolves #954 

## 🔧 What changed
added the loading component when opening a healthcare professional profile in moderation
## 🧪 Testing instructions
loading is usually too fast, so changing network speed may allow you to view the loading component
## 📸 Screenshots
-   ### After
https://github.com/user-attachments/assets/7214da0c-f3bc-4376-af49-3c3ccb95af34

### DIFF
--- a/components/ModEditHealthcareProfessionalSection.vue
+++ b/components/ModEditHealthcareProfessionalSection.vue
@@ -1,4 +1,5 @@
 <template>
+    <Loader />
     <div v-if="isHealthcareProfessionalInitialized">
         <div class="mod-healthcare-professional-section">
             <h2 class="mb-3.5 text-start text-primary-text text-2xl font-bold font-sans leading-normal">
@@ -272,6 +273,9 @@ let toast: ToastInterface
 const route = useRoute()
 
 const { t } = useI18n()
+
+const loadingStore = useLoadingStore()
+loadingStore.setIsLoading(true)
 
 const localesStore = useLocaleStore()
 const healthcareProfessionalsStore = useHealthcareProfessionalsStore()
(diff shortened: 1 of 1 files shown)


theyokohamalife:
Nit: Generally it's good practice to avoid using abbreviations in PR titles if it's something that may not be immediately obvious to other devs. What do you think of something like `feat: add loader to ModEdithealthcareProfessionalSection component`?

ermish (approved):
nice first PR @ConwayDecker! You raised a good point around the loader being too fast and we should fix this so the user experience is smooth. Typically you just add a fake delay to make it feel like it's doing something. 
For this PR, I think we can skip it and have a separate issue (cc @NabbeunNabi ) for making the loader show a min time.

ermish:
@ConwayDecker I've approved the PR, feel free to merge AFTER you've updated the title like @theyokohamalife mentioned

ConwayDecker:
I agree with @theyokohamalife that this title is better! If there's nothing else to change I guess it should be ready to merge @ermish . 
Thanks everyone!

ermish:
Adding you to the contributors list!
https://github.com/all-contributors please add @ConwayDecker for code

ermish:
https://github.com/all-contributors please add @ConwayDecker for code

PR #1025, opened 2025-02-09
fix: ESLint config

### LINKED ISSUE #982: fix: `eslint.config.js`
# Problem
Our `eslint.config.js` is misconfigured, causing some rules not to be applied. This is problematic as it leads to coding inconsistencies in our codebase.

# Success
Fix `eslint.config.js` and refactor all files after resolving the issue.

# Implementation Notes
The root of the problem is as follows:

Image

- The path is incorrect.
- The `tseslint.configs.recommended` is an array, so it's either in the wrong place or using the wrong property.

After correcting these issues, we will encounter numerous ESLint errors:

Image

Image

gminetoma:
✅ Resolves #982

## 🔧 What Changed  

We are now using the correct recommended TSESLint rules (`tseslint.configs.eslintRecommended.rules`).  

Previously, we were using the wrong plugin property: `tseslint.configs.recommended`. This includes not only the rules but the entire recommended plugin configuration.  

### Added Vitest ESLint Plugin and Fixed the Following Errors:  

- #### Styling  
  - `max-statements-per-line` turned off and removed `curly` ESLint rule.  

- #### `no-console`  
  - Updated this rule to only affect `console.log`.  
  - `console.warn`, `console.error` and `console.info` can still be used.  
  - In the future, this rule should block all `console` commands once a logger is added.  

- #### `no-underscore-dangle`  
  - Added ESLint ignore and renamed variables.  

- #### `no-shadow`  
  - Renamed conflicting variables.  

- #### `consistent-return`  
  - Removed the rule due inconsistencies across the codebase.

- #### `no-alert`  
  - Added `eslint-disable-next-line no-alert`

### Notes:  
I didn't remove the `require-atomic-updates` warning. This needs further investigation—it may just be a false alarm.

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -31,14 +31,17 @@ export default withNuxt(
                 }
             }
         },
-        files: ['test**/*.ts', './**/*.ts', './**/*.js'],
+        files: ['**/*.{ts,js}'],
         plugins: {
             '@typescript-eslint': tseslint.plugin,
             '@stylistic': stylistic
         },
         ignores: ['./typeDefs/gqlTypes.ts', './typesgeneratorconfig.ts'],
         rules: {
-            ...tseslint.configs.recommended,
+            // Current eslintRecommended rules (https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/src/configs/eslint-recommended-raw.ts)
+            // More information about the plugin (https://typescript-eslint.io/users/configs#eslint-recommended)
+            ...tseslint.configs.eslintRecommended.rules,
+
             // TS specific rules
             '@typescript-eslint/no-shadow': 'error',
             '@typescript-eslint/no-unused-vars': 'error',
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -52,7 +52,6 @@ export default withNuxt(
             'no-unused-vars': 'off',
             'block-scoped-var': 'error',
             complexity: ['error', { max: 40 }],
-            curly: 'error',
             'dot-location': ['error', 'property'],
             'dot-notation': ['error', { allowPattern: '^[a-z]+(_[a-z]+)+$' }],
             'no-alert': 'error',
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -52,7 +52,6 @@ export default withNuxt(
             'no-unused-vars': 'off',
             'block-scoped-var': 'error',
             complexity: ['error', { max: 40 }],
-            'consistent-return': 'error',
             curly: 'error',
             'dot-location': ['error', 'property'],
             'dot-notation': ['error', { allowPattern: '^[a-z]+(_[a-z]+)+$' }],
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -36,7 +36,7 @@ export default withNuxt(
             '@typescript-eslint': tseslint.plugin,
             '@stylistic': stylistic
         },
-        ignores: ['./typeDefs/gqlTypes.ts', './typesgeneratorconfig.ts'],
+        ignores: ['typeDefs/gqlTypes.ts', 'typesgeneratorconfig.ts', 'i18n/checkLocaleKeys.js'],
         rules: {
             // Current eslintRecommended rules (https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/src/configs/eslint-recommended-raw.ts)
             // More information about the plugin (https://typescript-eslint.io/users/configs#eslint-recommended)
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -3,6 +3,7 @@ import eslintJsPlugin from '@eslint/js'
 import tseslint from 'typescript-eslint'
 import stylistic from '@stylistic/eslint-plugin'
 import pluginCypress from 'eslint-plugin-cypress/flat'
+import pluginVitest from '@vitest/eslint-plugin'
 import pluginVue from 'eslint-plugin-vue'
 import withNuxt from './.nuxt/eslint.config.mjs'
 
@@ -15,7 +16,6 @@ export default withNuxt(
             '.nuxt/*',
             'coverage/*',
             '.yarn/*',
-            'test/vitest',
             'typedefs/gqlTypes.ts'
         ]
     },
@@ -103,6 +103,16 @@ export default withNuxt(
         ...pluginCypress.configs.recommended,
         rules: {
             'cypress/no-unnecessary-waiting': 'off'
+        }
+    },
+    // Linting for Vitest (https://github.com/vitest-dev/eslint-plugin-vitest)
+    {
+        files: ['test/vitest/**/*'],
+        ...pluginVitest.configs.recommended,
+        ...pluginVitest.configs.env,
+        rules: {
+            '@typescript-eslint/no-unused-expressions': 'off',
+            'no-unused-expressions': 'off'
         }
     }
 )
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -170,7 +170,9 @@ export default withNuxt(
                     ignoreTemplateLiterals: true,
                     ignoreRegExpLiterals: true
                 }
-            ], // be friendly to laptops
+            ],
+            '@stylistic/max-statements-per-line': 'off',
+            // be friendly to laptops
             '@stylistic/padding-line-between-statements': 'off', //we can choose newlines after variable declarations
             '@stylistic/no-multiple-empty-lines': ['error', { max: 1, maxEOF: 1 }],
             '@stylistic/object-curly-newline': ['error', { consistent: true }],
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -65,7 +65,8 @@ export default withNuxt(
             ],
             'vars-on-top': 'off',
             yoda: ['error', 'never', { exceptRange: true }],
-            'no-console': 'error', // we should use the logger instead
+            // Change to a logger in the future
+            'no-console': ['error', { allow: ['warn', 'error', 'info'] }],
             // Vue specific rules
 
             // Stylistic Issues and Opinions
--- a/stores/facilitiesStore.ts
+++ b/stores/facilitiesStore.ts
@@ -47,27 +47,27 @@ export const useFacilitiesStore = defineStore(
                 .find((facility: Facility) => facility.id === facilityId)
         }
 
-        function initializeFacilitySectionValues(facilityData: Facility | undefined) {
-            if (!facilityData) return
+        function initializeFacilitySectionValues(data: Facility | undefined) {
+            if (!data) return
 
-            facilitySectionFields.nameEn = facilityData.nameEn
-            facilitySectionFields.nameJa = facilityData.nameJa
-            facilitySectionFields.phone = facilityData?.contact?.phone
-            facilitySectionFields.email = facilityData?.contact?.email || undefined
-            facilitySectionFields.website = facilityData?.contact?.website || undefined
-            facilitySectionFields.postalCode = facilityData.contact?.address.postalCode
-            facilitySectionFields.prefectureEn = facilityData?.contact?.address?.prefectureEn
-            facilitySectionFields.cityEn = facilityData?.contact?.address?.cityEn
-            facilitySectionFields.addressLine1En = facilityData?.contact?.address?.addressLine1En
-            facilitySectionFields.addressLine2En = facilityData?.contact?.address?.addressLine2En
-            facilitySectionFields.prefectureJa = facilityData?.contact?.address?.prefectureJa
-            facilitySectionFields.cityJa = facilityData?.contact?.address?.cityJa
-            facilitySectionFields.addressLine1Ja = facilityData?.contact?.address?.addressLine1Ja
-            facilitySectionFields.addressLine2Ja = facilityData?.contact?.address?.addressLine2Ja
-            facilitySectionFields.googlemapsURL = facilityData?.contact?.googleMapsUrl
-            facilitySectionFields.healthcareProfessionalIds = facilityData.healthcareProfessionalIds
-            facilitySectionFields.mapLatitude = facilityData.mapLatitude.toString()
-            facilitySectionFields.mapLongitude = facilityData.mapLongitude.toString()
+            facilitySectionFields.nameEn = data.nameEn
+            facilitySectionFields.nameJa = data.nameJa
+            facilitySectionFields.phone = data?.contact?.phone
+            facilitySectionFields.email = data?.contact?.email || undefined
+            facilitySectionFields.website = data?.contact?.website || undefined
+            facilitySectionFields.postalCode = data.contact?.address.postalCode
+            facilitySectionFields.prefectureEn = data?.contact?.address?.prefectureEn
+            facilitySectionFields.cityEn = data?.contact?.address?.cityEn
+            facilitySectionFields.addressLine1En = data?.contact?.address?.addressLine1En
+            facilitySectionFields.addressLine2En = data?.contact?.address?.addressLine2En
+            facilitySectionFields.prefectureJa = data?.contact?.address?.prefectureJa
+            facilitySectionFields.cityJa = data?.contact?.address?.cityJa
+            facilitySectionFields.addressLine1Ja = data?.contact?.address?.addressLine1Ja
+            facilitySectionFields.addressLine2Ja = data?.contact?.address?.addressLine2Ja
+            facilitySectionFields.googlemapsURL = data?.contact?.googleMapsUrl
+            facilitySectionFields.healthcareProfessionalIds = data.healthcareProfessionalIds
+            facilitySectionFields.mapLatitude = data.mapLatitude.toString()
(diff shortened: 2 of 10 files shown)


eslint.config.js
```diff
@@ -62,7 +62,8 @@ export default withNuxt(
             ],
             'vars-on-top': 'off',
             yoda: ['error', 'never', { exceptRange: true }],
-            'no-console': 'error', // we should use the logger instead
+            // Change to a logger in the future
```
NabbeunNabi, line 65: Do you want to write a proposal and research what logger can be implemented. Give a few choices as to what you found and the benefits to each. You can use the template in the `gitbook`. @ermish does that sound good?

stores/facilitiesStore.ts
```diff
@@ -47,27 +47,29 @@ export const useFacilitiesStore = defineStore(
...
-            facilitySectionFields.addressLine1Ja = facilityData?.contact?.address?.addressLine1Ja
-            facilitySectionFields.addressLine2Ja = facilityData?.contact?.address?.addressLine2Ja
-            facilitySectionFields.googlemapsURL = facilityData?.contact?.googleMapsUrl
-            facilitySectionFields.healthcareProfessionalIds = facilityData.healthcareProfessionalIds
-            facilitySectionFields.mapLatitude = facilityData.mapLatitude.toString()
-            facilitySectionFields.mapLongitude = facilityData.mapLongitude.toString()
+        function initializeFacilitySectionValues(data: Facility | undefined) {
+            if (!data) {
+                return
+            }
```
NabbeunNabi, line 53: The new linting rules want this styling? I like it as long as it's consistent. Also probably more readable for people learning a codebase. 😎

stores/locationsStore.ts
```diff
@@ -55,7 +55,7 @@ async function queryFacilities(): Promise<Facility[]> {
         return result?.facilities ?? []
     } catch (error) {
         console.error(`Error getting facilities for dropdown: ${JSON.stringify(error)}`)
-        alert('Error getting data! Please contact our support team by clicking the bottom right link on the page!')
+        console.error('Error getting data! Please contact our support team by clicking the bottom right link on the page!')
```
NabbeunNabi, line 58: Nice! This alert would cause a 500 on `yarn dev:localserver` if the backend wasn't running

stores/localeStore.ts
```diff
@@ -17,7 +17,7 @@ export const useLocaleStore = defineStore('locale', () => {
     }
 
     function formatLanguageCodeToSimpleText(selectedLocale: string) {
-        const language = localeDisplayOptions.find(locale => locale.code === selectedLocale)
+        const language = localeDisplayOptions.find(l => l.code === selectedLocale)
```
NabbeunNabi, line 20: Is the one letter variable based on a conflict that you mentioned? Is there any way to keep it as locale or name it something more readable in order for other developers quickly looking at the function to know what l stands for?

stores/moderationSubmissionsStore.ts
```diff
@@ -111,7 +111,7 @@ export const useModerationSubmissionsStore = defineStore(
         async function approveSubmission() {
             if (!selectedSubmissionId.value) {
                 console.error('Unable to approve submission. Submission ID is required.')
-                return
+                return undefined
```
NabbeunNabi, line 114: I think this is good since we like to help people who don't know code at all to join the codebase. This will help solidify some knowledge of JavaScript

utils/handleScroll.ts
```diff
@@ -7,7 +7,9 @@ export interface SectionInformation {
 
 export const handleScroll
 = (sectionDetailsObject: SectionInformation[], isScrolling: Ref<boolean> = ref(false), activeSection: Ref<string> = ref('')) => {
-    if (isScrolling.value) return
+    if (isScrolling.value) {
+        return
```
NabbeunNabi, line 11: @gminetoma why does this not require `return undefined`?

NabbeunNabi (commented):
@gminetoma Great work.

Left a few comments. A couple questions and then I will passed it off to Phil

eslint.config.js
```diff
@@ -31,14 +31,14 @@ export default withNuxt(
...
+        files: ['**/*.{ts,js}'],
         plugins: {
             '@typescript-eslint': tseslint.plugin,
             '@stylistic': stylistic
         },
         ignores: ['./typeDefs/gqlTypes.ts', './typesgeneratorconfig.ts'],
         rules: {
-            ...tseslint.configs.recommended,
+            ...tseslint.configs.eslintRecommended.rules,
             // TS specific rules
```
NabbeunNabi, line 42: Also maybe explain in your PR description why this needed changing and add a comment above it as to what that is even doing. No one would know what either do without doing the research you did

stores/facilitiesStore.ts
gminetoma: I didn't change any rules besides the `no-console`

stores/localeStore.ts
gminetoma: Solved!

utils/handleScroll.ts
gminetoma: Because it's returning only undefined.

https://eslint.org/docs/latest/rules/consistent-return

eslint.config.js
NabbeunNabi: Or @ermish do you already have one you like?

utils/handleScroll.ts
NabbeunNabi: I see. I'll leave it up to @ermish  then whether he would want you to change this rule to not matter. I said earlier I like it explicitly doing it for newcomers; however in this context it makes it inconsistent throughout the codebase. So wondering if changing our use of the rule to ignore it if undefined is not explicitly stated is better. I usually like consistency throughout. So either all returns need undefined or all don't.

NabbeunNabi (approved):
@ermish On to you. I tagged you in a couple of my comments

stores/moderationSubmissionsStore.ts
NabbeunNabi: @gminetoma @ermish The more I think about it I think maybe removing this and having it not explicitly stated undefined is better.

It is like a semicolon. Unnecessarily added text. Probably better to learn that JavaScript returns undefined when just return is written

utils/handleScroll.ts
gminetoma: I also prefer consistency. Typing `return` at the end of a function feels weird, and typing `return undefined` feels even weirder.  

I understand that this rule is meant to prevent forgetting to return something, but it seems more annoying than helpful.

NabbeunNabi:
> @ermish On to you. I tagged you in a couple of my comments

Actually I'm going to use my judgment to say I want to have return there but not have to explicitly say I defined. So not yet to you

NabbeunNabi (changes requested):
Actually, I changed my mind. Sorry to do that. Could you see if it is possible to update the config to not require explicitly stating undefined. I am fine with there being a return. But adding undefined seems extra, like a semicolon and we don't use those

utils/handleScroll.ts
NabbeunNabi: I'm glad we got rid of that

NabbeunNabi (approved):
Ok @ermish now it is all yours

eslint.config.js
ermish: We already have a specific one to use. It's with Grafana / Prometheus

stores/localeStore.ts
```diff
@@ -9,15 +9,15 @@ export type LocaleDisplay = {
 }
 
 export const useLocaleStore = defineStore('locale', () => {
-    const enUsLocale = localeDisplayOptions.find(o => o.code === Locale.EnUs) as LocaleDisplay
+    const enUsLocale = localeDisplayOptions.find(currentLocale => currentLocale.code === Locale.EnUs) as LocaleDisplay
```
ermish, line 12: Is this an error from the linter around single letter vars?

I definitely love well worded variables so it's clear, but I would say that these were fine before since it was a simple operation and easier to read the short line.

stores/locationsStore.ts
ermish: Wait. This was intentionally an alert because it's saying our whole site has crashed and we want to tell the user. We could replace this with a modal as an alternative, but console won't work here. 

We also have a console error in the line before to give more dev details which this is now duplicating

stores/searchResultsStore.ts
```diff
@@ -100,7 +100,7 @@ async function queryProfessionals(searchSpecialty?: Specialty, searchLanguage?:
         return professionalsSearchResult
     } catch (error) {
         console.error(`Error getting professionals: ${JSON.stringify(error)}`)
-        alert('Error getting data! Please contact our support team by clicking the bottom right link on the page!')
```
ermish, line 103: Also should keep this one

stores/searchResultsStore.ts
```diff
@@ -135,7 +135,7 @@ async function queryFacilities(healthcareProfessionalIds: string[], searchCity?:
         return locationFilteredSearchResults
     } catch (error) {
         console.error(`Error getting facilities: ${JSON.stringify(error)}`)
-        alert('Error getting data! Please contact our support team by clicking the bottom right link on the page!')
```
ermish, line 138: And this one

ermish: @NabbeunNabi  We can make a separate issue for moving these alerts to nicer user modal messages

utils/formValidations.ts
```diff
@@ -168,34 +168,41 @@ export function validateFloat(float: string): boolean {
 }
 
 export function validateUsername(name: string): boolean {
-    if (name.length < 1)
+    if (name.length < 1) {
```
ermish, line 171: Let's put back the rule that lets these be shorter

utils/handleScroll.ts
ermish: +1 don't want this rule. I've never seen a codebase using everything undefined lol 😂

utils/handleServerErrorMessaging.ts
```diff
@@ -13,7 +13,7 @@ export function handleServerErrorMessaging(
         }
         toast.error((t('serverErrorMessages.genericErrorMessage')))
         console.error((t('serverErrorMessages.genericErrorMessage')))
-        console.info(t('serverErrorMessages.errorCodeMessagingNeeded'))
```
ermish, line 16: I think this one should be info since we already have the error right before it

ermish (changes requested):
Human linting of the PR complete!

eslint.config.js
```diff
@@ -101,6 +103,28 @@ export default withNuxt(
...
+            vitest: pluginVitest
+        },
+        rules: {
+            ...pluginVitest.configs.recommended.rules,
+            '@typescript-eslint/no-unused-expressions': 'off',
+            'no-unused-expressions': 'off'
+        },
+        settings: {
+            vitest: {
+                typecheck: true
```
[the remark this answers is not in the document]
gminetoma: This feature is for Vitest if type checking is enabled in its config.  
Since it's not, I'll remove it.  

https://github.com/vitest-dev/eslint-plugin-vitest?tab=readme-ov-file#enabling-with-type-testing  
https://vitest.dev/guide/testing-types

i18n/checkLocaleKeys.js
```diff
@@ -1,13 +1,14 @@
+/* eslint-disable no-console */
 import fs from 'fs'
 import path from 'path'
 import { fileURLToPath } from 'url'
 
 // Convert import.meta.url to __dirname equivalent
-const __filename = fileURLToPath(import.meta.url)
-const __dirname = path.dirname(__filename)
+const fileName = fileURLToPath(import.meta.url)
```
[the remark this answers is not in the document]
gminetoma: Should I keep it or remove it? These names are not available in ES modules.  

We have rules that block the usage of "_" for variable naming.  
Even if we were running a `.cjs` script, we shouldn't modify them.  

https://stackoverflow.com/questions/75834128/why-are-filename-and-dirname-not-available-in-es-modules

i18n/checkLocaleKeys.js
```diff
@@ -1,13 +1,14 @@
+/* eslint-disable no-console */
```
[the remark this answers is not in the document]
gminetoma: Should I apply it to each `console.log` line then?

i18n/checkLocaleKeys.js
```diff
@@ -28,7 +29,7 @@ const checkLocaleKeys = () => {
 
     //insert in development
     findAndInsertMissingKeysAndValuesInNonEnFiles(
-        localeFilesWithoutEnFileName, enFileContent, referenceKeys, insertMissingKeysAndValues
+        localeFilesWithoutEnFileName, enFileContent, referenceKeys
```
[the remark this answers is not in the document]
gminetoma: I removed it because the function in question is in the global scope.  
It doesn't make sense to pass them to a function when we can use them directly.

stores/searchResultsStore.ts
gminetoma: I'm bringing the alerts back. Should I remove the rule or ignore it per line?

stores/facilitiesStore.ts
gminetoma: @ermish  

My intention was to apply the correct recommended rules since `...tseslint.configs.recommended` was only being applied in the rules, which was denying all the rules in the block.  

I just fixed it, but if this is not what was intended, we can remove it.

utils/formValidations.ts
gminetoma: I didn't touch this rule. It was being applied 8 months ago when ESLint was working.  

image

gminetoma: I didn't modify any styling rules. This rule comes from here:  

image  

image  

We can override it if you approve.

stores/localeStore.ts
gminetoma: @NabbeunNabi Will linting this one!

stores/facilitiesStore.ts
gminetoma: I didn't touch this rule. It was being applied 8 months ago when ESLint was working.  

image

gminetoma: I didn't modify any styling rules. This rule comes from here:  

image  

image  

We can override it if you approve.

i18n/checkLocaleKeys.js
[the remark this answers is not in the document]
ermish: They're well known names in the JavaScript/node space. The linting is from new rules added in this PR that are throwing the error so we should just override the rule. 
I think the rule can be ok if it allows exceptions for any other reserved words/special cases

i18n/checkLocaleKeys.js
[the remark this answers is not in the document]
ermish: Yes for all the ones that are intentional 👍

i18n/checkLocaleKeys.js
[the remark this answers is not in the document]
ermish: Ah makes sense. 👍

stores/facilitiesStore.ts
ermish: Let's add an override and use the simpler way 👍

Sorry if I wasn't clear, I meant even though you didn't add any specific rules, this PR still does enable rules that were not being linted before it.

utils/formValidations.ts
ermish: Even though you didn't directly add it, the PR still did 😆 
Let's go ahead and change the curly override rule to allow it

stores/facilitiesStore.ts
gminetoma: Should I make this a rule, or can we leave it up to the developer to decide? What format do you prefer? 
```
if (condition) return

if (condition) {
 return
}
```

i18n/checkLocaleKeys.js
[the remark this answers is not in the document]
gminetoma: I added this file to the ignore list since it doesn't follow a bunch of our rules.  
I also reverted all its changes.

stores/searchResultsStore.ts
NabbeunNabi: OK! I think it of course good for users. But the developers who just start too think they broke something possibly. So a modal with the same contact our development team message displayed in another location would be good?

I think this is on auth

@ermish

stores/facilitiesStore.ts
ermish: Leave it to the dev 👍
I think both are acceptable and readability depends on case by case

gminetoma: Removed the `curly` rule

stores/searchResultsStore.ts
ermish: hai, I think that would work @NabbeunNabi

ermish (approved):
LGTM! Linting complete!

i18n/checkLocaleKeys.js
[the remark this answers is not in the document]
ermish: this should be fine since it's a file used for building only

PR #1026, opened 2025-02-11
feat: refactor arraysAreEqual utility function to also handle non-primitive variables

### LINKED ISSUE #998: Improve our utility function to check array equality
# Problem
Currently we have a utility function that checks the equality of arrays, but it only works for primitive data types. We would want this to work for any type of array that is being passed.

# Success
This array utility function has been updated with tests to prove that it can take primitive data types of a non sorted array and test for equality, by first sorting and then checking for equality, or an array holding non-primitive data types.

### Requirements

- [ ] Array utility function is updated to work for primitive and non primitive types
- [ ] Tests are written to show that it will return false if they do not match and true if they do match for primitive data types
- [ ] Tests are written to show that it will return false if they do not match and true if they do match for non-primitive data types

# Implementation notes
The function is located in the code [here](https://github.com/ourjapanlife/findadoc-web/blob/d52fa4c25eabfcdd3a0742aade5bec9b5702de3c/utils/arrayUtils.ts#L11)

Testing should be added to a new file [here](https://github.com/ourjapanlife/findadoc-web/tree/main/test/vitest) and run with 

```
yarn test:pinia
```

We might want to separate out the pinia store tests and utils test within vitest

vororo98:
✅ Resolves #998

## 🔧 What changed
arraysAreEqual in utils/arraysUtils.ts can now sort and compare arrays that contain objects or arrays. It does this by turning the values into strings and then comparing them. Tests were also added in a new file test/vitest/arrayUtils.spec.ts that checks for primitive arrays as well as non-primitive.

## 🧪 Testing instructions
"yarn test:pinia" runs the test

### DIFF
--- a/test/vitest/arrayUtils.spec.ts
+++ b/test/vitest/arrayUtils.spec.ts
@@ -1,11 +1,7 @@
-import { setActivePinia, createPinia } from 'pinia'
 import { expect } from 'chai'
 import { arraysAreEqual } from '@/utils/arrayUtils'
 
 describe('arraysAreEqual', () => {
-    beforeEach(() => {
-        setActivePinia(createPinia())
-    })
 
     it('should sort and find two arrays with the same primitive values to be equal', () => {
         expect(arraysAreEqual([1,2,3], [3,1,2])).to.be.true
--- a/test/vitest/arrayUtils.spec.ts
+++ b/test/vitest/arrayUtils.spec.ts
@@ -1,0 +1,46 @@
+import { setActivePinia, createPinia } from 'pinia'
+import { expect } from 'chai'
+import { arraysAreEqual } from '@/utils/arrayUtils'
+
+describe('arraysAreEqual', () => {
+    beforeEach(() => {
+        setActivePinia(createPinia())
+    })
+
+    it('should sort and find two arrays with the same primitive values to be equal', () => {
+        expect(arraysAreEqual([1,2,3], [3,1,2])).to.be.true;
+    })
+
+    it('should sort and find two arrays with different primitive values to be not equal', () => {
+        expect(arraysAreEqual([1,2,3], [4,5,6])).to.be.false;
+    })
+
+
+    it('should sort and find two arrays with the same object values to be equal', () => {
+        const arr1 = [{name: "Tony", profession: "tester"}, {name: "Sara", profession: "coder"}];
+        const arr2 = [{name: "Sara", profession: "coder"}, {name: "Tony", profession: "tester"}];
+
+        expect(arraysAreEqual(arr1, arr2)).to.be.true;
+    })
+
+    it('should sort and find two arrays with different object values to be not equal', () => {
+        const arr1 = [{name: "Tony", profession: "tester"}, {name: "Sara", profession: "coder"}];
+        const arr2 = [{name: "Max", profession: "Truck Driver"}, {name: "Helen", profession: "History Professor"}];
+
+        expect(arraysAreEqual(arr1, arr2)).to.be.false;
+    })
+
+    it('should sort and find two arrays with the same array values to be equal', () => {
+        const arr1 = [[1,2,3], [4,5,6]];
+        const arr2 = [ [4,5,6], [1,2,3]];
+
+        expect(arraysAreEqual(arr1, arr2)).to.be.true;
+    })
+
+    it('should sort and find two arrays with different array values to be not equal', () => {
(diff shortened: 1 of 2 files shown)


test/vitest/arrayUtils.spec.ts
```diff
@@ -0,0 +1,46 @@
+import { setActivePinia, createPinia } from 'pinia'
+import { expect } from 'chai'
+import { arraysAreEqual } from '@/utils/arrayUtils'
+
+describe('arraysAreEqual', () => {
+    beforeEach(() => {
+        setActivePinia(createPinia())
+    })
```
NabbeunNabi, line 8: I think you took this from the other files. But since this isn't testing a `Pinia` store you can remove line one imports and lines 6-8. The tests should still pass

utils/arrayUtils.ts
```diff
@@ -8,17 +8,53 @@ export function removeDuplicates<T>(inputArray: T[]): T[] {
...
+    const sortedArray2 = [...array2].sort(compareElements)
 
     // Compare each element in the sorted arrays
-    return sortedArray1.every((value, index) => value === sortedArray2[index])
+    for (let i = 0; i < sortedArray1.length; i++) {
+        if (canonicalize(sortedArray1[i]) !== canonicalize(sortedArray2[i])) {
+            return false
+        }
+    }
+
```
NabbeunNabi, line 28: Nice job! I usually prefer array methods but in this case I think this makes it clearer and more readable and can exit easily once something is not the same.

utils/arrayUtils.ts
```diff
@@ -8,17 +8,53 @@ export function removeDuplicates<T>(inputArray: T[]): T[] {
...
+    return true
+}
+
+function compareElements<T>(ele1: T, ele2: T): number {
+    const ele1Str = canonicalize(ele1)
+    const ele2Str = canonicalize(ele2)
+    if (ele1Str < ele2Str) return -1
+    if (ele1Str > ele2Str) return 1
+    return 0
+}
```
NabbeunNabi, line 38: I can understand what this is doing, but could you leave a comment above it explain what it is doing?

utils/arrayUtils.ts
```diff
@@ -8,17 +8,53 @@ export function removeDuplicates<T>(inputArray: T[]): T[] {
...
+
+//This function handles different data types and returns them as canonicalized strings
+function canonicalize<T>(value: T): string {
```
NabbeunNabi, line 41: Similar to how you verbally explained it to me yesterday, could you leave a comment to make it clear why you chose this function name and what this is doing. You can leave comments above the conditionals too if you think that will help. 

Just as a note. There is a character restriction per line on the linter so for multi line comments you can start it with 
```
/* Your
multi
line
comment */
```

NabbeunNabi (changes requested):
@vororo98 This is awesome work. Just a few comments on things you can add and change. Thanks for the hard work!

utils/arrayUtils.ts
```diff
@@ -8,17 +8,53 @@ export function removeDuplicates<T>(inputArray: T[]): T[] {
...
+    if (value === null) return 'null'
+    if (value === undefined) return 'undefined'
+
+    if (typeof value !== 'object') {
+        return JSON.stringify(value)
+    }
+
+    if (Array.isArray(value)) {
+        return '[' + value.map(canonicalize).join(',') + ']'
+        //return '[' + (value as Array<unknown>).map(canonicalize).join(',') + ']'
```
NabbeunNabi, line 51: @vororo98 I also mean to add you can delete this as it looks like a remnant of something you were trying

test/vitest/arrayUtils.spec.ts
vororo98: Got it!

utils/arrayUtils.ts
vororo98: Commented!

utils/arrayUtils.ts
vororo98: Good catch, deleted

NabbeunNabi (approved):
@vororo98 Great job! On to @ermish

NabbeunNabi:
@vororo98 I approved it, but you will need to go to your branch and pull the changes that were merged with the button I clicked. From there you will need to run `yarn lint:fix` and push a commit. That can be something like 

`chore: run linter to match codebase styling`

ermish (approved):
Looks great! nice work!

PR #1033, opened 2025-02-12
fix: update workflow to autogenerate PR with type differences

### LINKED ISSUE #1032: Fix workflow action for auto generating frontend types from backend to see if there is a difference
# Problem
There is a bug where even though there are differences in the backend types and frontend types that this workflow is ending on an error code instead of making a PR and committing the updated types.

# Success
The workflow is fixed, manually triggered, and you can see that it has made a PR since there is an updated type in the backend seen [here](https://github.com/ourjapanlife/findadoc-server/blob/27f7f3987ffcfeed10cb50367c4bd74f93998e62/src/typeDefs/gqlTypes.ts#L509)
### Requirements

- [ ] Workflow fixed and run
- [ ] Workflow makes PR from bot that includes the type above

NabbeunNabi:
✅ Resolves #1032 

## 🔧 What changed
Before this `cron` job was running but wasn't actually making a PR and exiting with an error code of 1 as seen in the before screenshot. Now it will run the script and if there are differences make a PR successfully as seen in #1034 automatically. After testing this was moved back to a `cron` job.

This uses this workflow action --> https://github.com/peter-evans/create-pull-request/tree/v5/
A PR will not be created if it isn't ahead of `main` as seen in the screenshot below.
## 📸 Screenshots

-   ### Before
image
image

-   ### After

image

image

### DIFF
--- a/.github/workflows/generate-gql-types.yml
+++ b/.github/workflows/generate-gql-types.yml
@@ -6,8 +6,9 @@ on:
     - cron: '0 0 * * MON'
 
 permissions:
-  contents: write  # Allows for committing and creating pull requests
+  contents: write
   actions: write
+  pull-requests: write
 
 jobs:
   generate-gql-types:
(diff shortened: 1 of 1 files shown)


.github/workflows/generate-gql-types.yml
```diff
@@ -2,12 +2,12 @@ name: Generate and Commit GQL Types
 
 on:
   schedule:
-    # Run the workflow every Monday at midnight (UTC)
```
ermish, line 5: We should add this comment back!

.github/workflows/generate-gql-types.yml
```diff
@@ -18,41 +18,34 @@ jobs:
...
-          git config --global user.name "github-actions[bot]"
-          git config --global user.email "github-actions[bot]@users.noreply.github.com"
-          git add typedefs/
-          git commit -m "chore: update gql types"
-          git push
 
-      # Create a pull request if changes were committed
       - name: Create Pull Request
-        if: failure()
         uses: peter-evans/create-pull-request@v5
```
ermish, line 39: Would be good to add a comment that this will only create a PR on if there's diffs.

ermish (changes requested):
autogenerated response: please resolve comments

ermish (approved):
Processing complete: GitHub Review CI passed

PR #1060, opened 2025-02-21
feat: enable pinia store access in cypress

### LINKED ISSUE #1068: Enable our cypress tests to access our pinia stores for better testing and e2e testing
# Problem
Since March 2024 we have been trying to implement the cypress tests to access pinia stores for more robust e2e testing but it has been difficult to access them in a way that was secure without making it available globally in production.

# Success
Pinia stores are added to the window for testing only when AUT (Application under testing)
### Requirements

- [x] Pinia stores accessible in cypress but only in testing and CI/CD
- [x] Tests pass locally when run and tested
- [x] Tests pass in CI/CD when run and tested
- [x] Types are correctly accessed.
- [x] A use case of how the store can be used within a testing file has been added so others can reference it if need be

NabbeunNabi:
✅ Resolves #1068

## 🔧 What changed
Users do not need to add an environmental variable and the tests run and pass locally and pass in CI/CD.
We can now test user interaction updating the store and making sure the values stay the types we want. Getting closer to true e2e testing!

## 🧪 Testing instructions
- You can run tests locally to see them pass within the `moderationEditFacility.cy.ts` file.
- You can also see below that the tests pass within CI/CD.

## 📸 Screenshots

-   ### After
image
image


plugins/expose-pinia.client.ts
```diff
@@ -0,0 +1,52 @@
...
+import { useSpecialtiesStore } from '~/stores/specialtiesStore'
+import { useSubmissionStore } from '~/stores/submissionStore'
+
+export default defineNuxtPlugin(nuxtApp => {
+    const config = useRuntimeConfig()
+
+    if (process.env.NODE_ENV === 'development' || process.env.NODE_ENV === 'test' || config.public.NUXT_PUBLIC_LOAD_STORES) {
+    /* Create the stores using the app's Pinia instance.
+    This must be done for every store. So if new stores are created for testing they need to be added here */
+        const authStore = useAuthStore(nuxtApp.$pinia as Pinia)
```
ermish, line 22: before we merge this, I'm wondering if getting `useStore` to be shared might be a simpler option with less upkeep?

ermish (changes requested):
left some comments

plugins/expose-pinia.client.ts
NabbeunNabi: @ermish removed the `development` part of the conditional cuz it is only necessary for testing environment which cypress sets to `test` and the CI/CD `env` variable. Now there is zero upkeep in both places

ermish (approved):
Access Enabled!

PR #1061, opened 2025-02-21
feat: add loading component to modeditfacilitysection

### LINKED ISSUE #957: feat: show loading placeholder in ModEditFacilitySection
# Problem
We currently have no loading placeholder in the `ModEditFacilitySection.vue`. We do have a loading component, but it isn't in use when the data is being loaded in this section. On refresh we need to fetch data from the server. For visual for users to understand that loading is happening we want to display the loader.

# Success
The loader is shown on the page load for the path `moderation/edit-facility/:id`
### Requirements

- [ ] The loader should appear when data is being loaded. 
- [ ] There should be no visual disruption when transitioning in/out. 
- [ ] It should make the experience of using the section clearer for the user. 
- [ ] A video is included in the PR to show the desired behavior

# Implementation notes
You can see in the [MainContentContainer.vue](https://github.com/ourjapanlife/findadoc-web/blob/main/components/MainContentContainer.vue) where the loader is being used.A

ljbroersen:
✅ Resolves #957 

## 🔧 What changed

Added a loading component when you open a facility profile in the moderation panel.

## 🧪 Testing instructions

I throttled the network speed to slow 4G since I could not catch the loader on video otherwise.

## 📸 Screenshots

-   ### After

https://github.com/user-attachments/assets/a6f69cdc-8341-47ee-99eb-3bdc9ac5baaf


NabbeunNabi (approved):
@ljbroersen Looks great to me

ermish:
Looks great!
Future ticket idea:  One improvement we could do for the user experience (UX) is to show the loader for at least 2 secs. (it can be a fake delay, but it makes it feel smoother instead of the jumpiness) 
We don't need to do it in this PR, but would be a nice upgrade

ermish (approved):
Nice! Thanks @ljbroersen !

PR #1062, opened 2025-02-23
fix: enable locale search by all locale display options then save locale code

### LINKED ISSUE #1053: fix: update ModEditHealthcareProfessionalSection to search by language name instead of locale code
# Problem
Currently ModEditHealthcareProfessionalSection with ModSearchBar searches by locale code instead of the full language name. 
This means when typing in i.e. English, it will only show up if "en" is typed, but "eng" will not display any results. 

Image
Image

# Success
In moderation panel healthcare professionals, should be able to search languages and show results with their full name instead of locale code. 

### Requirements

- [ ] When searching language, results show based on language name instead of locale code. 
- [ ] Still saves the locale code to the spoken locales as that is what the server needs

# Implementation notes
Related files:
[ModSearchBar.vue](https://github.com/ourjapanlife/findadoc-web/blob/main/components/ModSearchBar.vue)
[ModEditHealthcareProfessionalSection.vue](https://github.com/ourjapanlife/findadoc-web/blob/main/components/ModEditHealthcareProfessionalSection.vue)
[healthcareProfessionalStore.ts](https://github.com/ourjapanlife/findadoc-web/blob/main/stores/healthcareProfessionalsStore.ts)

prestonb91:
✅ Resolves #1053 

## 🔧 What changed
- ModEditHealthcareProfessionalSection only searched by filtering through language codes
- This prevented users from searching with the full locale name text (i.e. English) or to search by the local language (i.e. 日本語 (Japan)
- Updated handleLocaleInputChange to filter across code, simple text, display text based on input value, then map the locale code to filtered items so that it still saves the locale code

## 📸 Screenshots
-   ### Before 
Locale-search-bug-1
Locale-search-bug-2

-   ### After
https://github.com/user-attachments/assets/abf7b85b-eb99-4ce7-bd16-0d77812e5412


NabbeunNabi (changes requested):
This is a good idea.

We do have a store that has functions to deal with locale values.  It has the simple text and code. Could you make a reusable store function with this type of logic where you can input a language and it will use the actually language and then return the code so you can set the value that way and put a comment above it.

Here is the file. https://github.com/ourjapanlife/findadoc-web/blob/main/stores/localeStore.ts

Great start!

prestonb91:
@NabbeunNabi thanks for the feedback. 
I refactored to create a reusable function getLocaleByInput in the localStore and adjusted the ModEditHealthcareProfessionalSection to use it. 
Manually tested and it searches by display options and saves like the original commit. 
https://github.com/ourjapanlife/findadoc-web/pull/1062/commits/34ad046f4a52846fe024bbce7de4db98a9d91d75

Wanted to confirm if this is what you were looking for based on your comments?

stores/localeStore.ts
```diff
@@ -30,7 +30,26 @@ export const useLocaleStore = defineStore('locale', () => {
               ?? []
     }
 
-    return { locale, localeDisplayOptions, mvpLocaleDisplayOptions, setLocale, formatLanguages, formatLanguageCodeToSimpleText }
+    // Takes the input value and searches all locale display options to return the locale code
+    function getLocaleByInput(inputValue: string) {
+        const filteredLocales = localeDisplayOptions
```
NabbeunNabi, line 35: Great job! I would think along with the comment you can name the function `getLocaleCodeBySimpleTextInput` or something similar to this for the name to know I am extracting the code by the language name.

NabbeunNabi (approved):
@prestonb91 just one thing to change on my end but I'm sure you will finish by the time @ermish reviews it.

So on to him!

stores/localeStore.ts
prestonb91: Okay! I updated it to “getLocaleByNameInput"
Just to clarify, didn’t name it get by simpleText because this will also look up by code and displayText. 
Reason was, for example, if it was only simple text, it would not pick up Japan locale unless typed out in kanji, same with cases of korean or hebrew, etc where all simpleText is not in English text. 
So to make the search more robust, it will pick up the locale with its code or simpleText or displayText.

stores/localeStore.ts
```diff
@@ -30,7 +30,26 @@ export const useLocaleStore = defineStore('locale', () => {
               ?? []
     }
 
-    return { locale, localeDisplayOptions, mvpLocaleDisplayOptions, setLocale, formatLanguages, formatLanguageCodeToSimpleText }
+    // Takes the input value and searches all locale display options to return the locale code
+    function getLocaleByNameInput(inputValue: string) {
+        const filteredLocales = localeDisplayOptions
+            .filter(localeDisplay =>
+                localeDisplay.code.toLowerCase().replace('_', '').includes(inputValue)
```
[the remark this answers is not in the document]
prestonb91: @ermish thank you for the feedback and have just made the changes!

ermish (approved):
Looks great! Nice work @prestonb91 :shipit:

PR #1074, opened 2025-03-07
chore: update ModMainContent with codebases styling and new screen check

NabbeunNabi:
✅ Resolves #[Issue number]

## 🔧 What changed
The `ModMainContent.vue` component was not following our codebase's styling to use mostly if statements only and was not using our updated active screen check. Now it does.

### DIFF
--- a/components/ModMainContent.vue
+++ b/components/ModMainContent.vue
@@ -1,16 +1,16 @@
 <template>
     <div class="h-full overflow-hidden">
-        <div v-if="screenStore.activeScreen === ModerationScreen.Dashboard">
+        <div v-if="moderationScreenStore.dashboardScreenIsActive()">
             <ModSubmissionListContainer />
         </div>
         <div
-            v-else-if="screenStore.activeScreen === ModerationScreen.EditSubmission"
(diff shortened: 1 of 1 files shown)


components/ModMainContent.vue
```diff
@@ -51,21 +51,27 @@ const selectedIdFromModSubmissionList = computed(() => route.params.id as string
 const setActiveScreenBasedOnRoute = async () => {
     await nextTick()
     if (routePathForModerationScreen.value.includes('edit-submission') && selectedIdFromModSubmissionList.value) {
-        screenStore.setActiveScreen(ModerationScreen.EditSubmission)
+        moderationScreenStore.setActiveScreen(ModerationScreen.EditSubmission)
```
ermish, line 54: should this be a switch statement? Do you think we'll add any more options?

NabbeunNabi: Yes! I think that is a great idea. We will be adding at least two more. Let's have @ConwayDecker refactor this within the issue he is working on now to be a switch statement

ermish (approved):
No else's or but's about it. Let's merge

PR #1081, opened 2025-03-07
fix: extend visibility loading component for improved UX

### LINKED ISSUE #1073: Extend the time that the loading component is visible to improve UX
# Problem
Based on Phil's [comment](https://github.com/ourjapanlife/findadoc-web/issues/1061#issuecomment-2697911781_) in #1061.

Right now, the loading component is only visible for a incredibly short amount of time, which results in a not-so-great UX. To improve this, we would like the loader to be longer visible (for at least 2 seconds).

# Success
On pages where the loading component is shown on refresh, the loader should be visible to the user for at least 2 seconds.

### Requirements

- [ ] The loading component should be visible for at least 2 seconds.
- [ ] After those 2 seconds are over, show the page as usual.

# Implementation notes
- For this issue, please don't throttle your network speed.
- The changes have to be made within the loading component.

ljbroersen:
✅ Resolves #1073 

## 🔧 What changed
The visibility of the loading component has been increased to 2 seconds to improve UX.

However, I also added a video of an 1 second delay, since I think 2 seconds are maybe a little bit too long. I would like to have someone's opinion on that.

## 🧪 Testing instructions
On the moderation panel, please choose a facility/healthcare professional which entry you want to edit. Then, while the edit page loads, the loading component will be visible. If you want to test it again, just refresh the page. Also, don't throttle your network speed.

## 📸 Video

https://github.com/user-attachments/assets/7fd3220b-70fd-4ded-88b7-584dd7cdbd16
1 second delay

https://github.com/user-attachments/assets/6bee24d2-4e2b-4467-863f-6b769ca3ab1f
2 seconds delay

### DIFF
--- a/stores/loadingStore.ts
+++ b/stores/loadingStore.ts
@@ -3,18 +3,12 @@ import { ref } from 'vue'
 
 export const useLoadingStore = defineStore('loading', () => {
     const isLoading: Ref<boolean> = ref(false)
-    let timeout: ReturnType<typeof setTimeout> | null = null
 
-    function setIsLoading(loadingStatus: boolean) {
-        if (loadingStatus) {
-            isLoading.value = true
+    function setIsLoading(newValue: boolean, duration: number = 0) {
+        isLoading.value = newValue
 
-            if (timeout) clearTimeout(timeout)
-
-            timeout = setTimeout(() => {
-                isLoading.value = false
-                timeout = null
-            }, 2000)
+        if (newValue && duration > 0) {
+            setTimeout(() => (isLoading.value = false), duration)
         }
     }
 
--- a/stores/loadingStore.ts
+++ b/stores/loadingStore.ts
@@ -2,7 +2,7 @@ import { defineStore } from 'pinia'
 import { ref } from 'vue'
 
 export const useLoadingStore = defineStore('loading', () => {
-    const isLoading = ref(false)
+    const isLoading: Ref<boolean> = ref(false)
 
     function setIsLoading(newValue: boolean) {
         if (newValue) {
--- a/stores/loadingStore.ts
+++ b/stores/loadingStore.ts
@@ -1,11 +1,14 @@
 import { defineStore } from 'pinia'
-import { ref, type Ref } from 'vue'
+import { ref } from 'vue'
 
 export const useLoadingStore = defineStore('loading', () => {
-    const isLoading: Ref<boolean> = ref(false)
+    const isLoading = ref(false)
 
     function setIsLoading(newValue: boolean) {
-        isLoading.value = newValue === true ? true : false
+        if (newValue) {
+            isLoading.value = true
+            setTimeout(() => (isLoading.value = false), 2000)
+        }
     }
 
     return { isLoading, setIsLoading }
--- a/stores/loadingStore.ts
+++ b/stores/loadingStore.ts
@@ -3,11 +3,18 @@ import { ref } from 'vue'
 
 export const useLoadingStore = defineStore('loading', () => {
     const isLoading: Ref<boolean> = ref(false)
+    let timeout: ReturnType<typeof setTimeout> | null = null
 
-    function setIsLoading(newValue: boolean) {
-        if (newValue) {
+    function setIsLoading(loadingStatus: boolean) {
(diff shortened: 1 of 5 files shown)


ermish (changes requested):
nice addition! Added a comment

stores/loadingStore.ts
```diff
@@ -1,11 +1,14 @@
...
+import { ref } from 'vue'
 
 export const useLoadingStore = defineStore('loading', () => {
     const isLoading: Ref<boolean> = ref(false)
 
     function setIsLoading(newValue: boolean) {
-        isLoading.value = newValue === true ? true : false
+        if (newValue) {
+            isLoading.value = true
+            setTimeout(() => (isLoading.value = false), 2000)
```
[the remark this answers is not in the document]
ljbroersen: I added an optional parameter to the setIsLoading function.

stores/loadingStore.ts
```diff
@@ -1,11 +1,21 @@
...
-import { ref, type Ref } from 'vue'
+import { ref } from 'vue'
 
 export const useLoadingStore = defineStore('loading', () => {
     const isLoading: Ref<boolean> = ref(false)
+    let timeout: ReturnType<typeof setTimeout> | null = null
 
-    function setIsLoading(newValue: boolean) {
-        isLoading.value = newValue === true ? true : false
+    function setIsLoading(loadingStatus: boolean) {
```
NabbeunNabi, line 8: @ljbroersen I think in this case @ermish might have meant an optional parameter to is loading where you can also pass `ms` as a second parameter and it equals zero unless provided a different value. So originally it is set to zero unless someone else puts an explicit ms number in there

NabbeunNabi (changes requested):

stores/loadingStore.ts
ljbroersen: Something like this then?

```
    function setIsLoading(newValue: boolean, duration: number = 0) {
        isLoading.value = newValue

        if (newValue && duration > 0) {
            setTimeout(() => (isLoading.value = false), duration)
        }
    }
```

NabbeunNabi: Something exactly like this.

And don't forget to add a test to make sure the behavior works as intended

NabbeunNabi:
@ljbroersen just converted to draft until I approve so @ermish doesn't look at it

stores/loadingStore.ts
ljbroersen: I've added a test. Could you check if this makes sense? If I run `yarn test:pinia` all tests, including the one below, pass.

``` 
it('should update isLoading to false after 2000ms', async () => {
        vi.useFakeTimers()
        const loadingStore = useLoadingStore()

        loadingStore.setIsLoading(true, 2000)
        expect(loadingStore.isLoading).to.be.true
        await vi.advanceTimersByTimeAsync(2000)

        loadingStore.setIsLoading(false)
        expect(loadingStore.isLoading).to.be.false

        vi.useRealTimers()
    })
```

test/vitest/loadingStore.spec.ts
```diff
@@ -21,5 +22,19 @@ describe('LoadingStore', () => {
...
+        const loadingStore = useLoadingStore()
+
+        loadingStore.setIsLoading(true, 2000)
+        expect(loadingStore.isLoading).to.be.true
+        await vi.advanceTimersByTimeAsync(2000)
+
+        loadingStore.setIsLoading(false)
+        expect(loadingStore.isLoading).to.be.false
+
+        vi.useRealTimers()
```
NabbeunNabi, line 37: Sure makes sense to me. I would just add a comment above it for anyone new on the codebase as to what `advanceTimersByTimeAsync` does

NabbeunNabi (approved):
@ljbroersen Looks good to me. Just add a comment and onto @ermish. I was wondering if we should throw an error if someone tries to set a timeout for the fake loading delay over a certain duration to not let developers do that. But That might be overkill and I don't know if that is common practice

test/vitest/loadingStore.spec.ts
ljbroersen: Added a comment!

PR #1089, opened 2025-03-22
fix: remove global variable declaration of env for cypress

NabbeunNabi:
✅ Resolves #[Issue number]

## 🔧 What changed
Before the global variable for auth env variables was being declared within the `cypress.config.ts`. Now it was removed since it caused CI/CD to fail and was unnecessary for tests
## 🧪 Testing instructions
CI/CD passes now

## 📸 Screenshots

-   ### Before
image

### DIFF
--- a/cypress.config.ts
+++ b/cypress.config.ts
@@ -1,15 +1,4 @@
-/* eslint-disable @typescript-eslint/no-namespace */
 import { defineConfig } from 'cypress'
-
-declare global {
-    namespace NodeJS {
-        interface ProcessEnv {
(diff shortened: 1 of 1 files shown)


global.d.ts
```diff
@@ -7,4 +7,11 @@ declare global {
         // Declare the type for the Pinia store instance injected by Nuxt into `window` for testing purposes
         $pinia: Pinia
     }
+    namespace NodeJS {
```
ermish, line 10: This could use a comment about why we need these

ermish (approved):
Looks properly configured

PR #1107, opened 2025-04-10
fix: add the trash can icon back for edit facility

### LINKED ISSUE #1106: Add active screen of edit facility to moderation page
# Problem
When changing the active screen check a conditional was removed which got rid of the trash can button for deleting a healthcare professional off of a specific screen.

# Success
The trash can and functionality is back
### Requirements

- [x] The trash can icon is back
- [x] The functionality is back

NabbeunNabi:
✅ Resolves #1106

## 🔧 What changed
When changing the past way to check moderation screen check this got removed which got rid of the functionality for a moderator to finish

## 🧪 Testing instructions
It is visible now

## 📸 Screenshots

-   ### Before
- 
image

-   ### After
image
image
image

### DIFF
--- a/components/ModDashboardHealthProfessionalCard.vue
+++ b/components/ModDashboardHealthProfessionalCard.vue
@@ -92,7 +92,9 @@
                 </div>
                 <div
                     v-if="!isHealthcareProfessionalReadyForRemoval(healthcareProfessional.id)
-                        && moderationScreenStore.editSubmissionScreenIsActive()"
+                        && moderationScreenStore.editSubmissionScreenIsActive()
+                        || moderationScreenStore.editFacilityScreenIsActive()
+                        && !isHealthcareProfessionalReadyForRemoval(healthcareProfessional.id)"
                     id="remove-related-healthcare-professional-to-facility"
                     class="flex w-8 items-center justify-center
                     cursor-pointer font-bold text-secondary text-sm self-start p-1"


theyokohamalife (approved):
~~We run the risk of this and other elements randomly disappearing due to the lack of tests that they're visible when the conditionals are met. ~~ Tests are our best defense to make sure that UI elements aren't accidentally removed, especially if they're necessary for performing certain functions (in this case removing a healthcare professional).

With the moderation panel revamp coming up, it's even more important that we're testing pages and components for specific functionality, so nothing gets broken, forgotten, or lost.

In this case, we should have a test for deleting a healthcare professional from a facility. Regardless of whatever icons we use now or in the future, that's something we'd expect to be able to do while looking at a facility.

NabbeunNabi:
> ~We run the risk of this and other elements randomly disappearing due to the lack of tests that they're visible when the conditionals are met.~ Tests are our best defense to make sure that UI elements aren't accidentally removed, especially if they're necessary for performing certain functions (in this case removing a healthcare professional).
> 
> With the moderation panel revamp coming up, it's even more important that we're testing pages and components for specific functionality, so nothing gets broken, forgotten, or lost.
> 
> In this case, we should have a test for deleting a healthcare professional from a facility. Regardless of whatever icons we use now or in the future, that's something we'd expect to be able to do while looking at a facility.

@theyokohamalife I agree one hundred percent. I want to see if a volunteer would like to learn and head up and lead the testing revamp and we make the testing more robust to avoid any problems for users especially with more non moderation user interaction coming up

PR #1130, opened 2025-04-16
fix: avoid race condition healthcareProfessionalStore.ts

### LINKED ISSUE #1129: fix: make update of facilitiesRelationships atomic to prevent ESLint warning
# Problem
Currently, facilitiesRelationsToSelectedHealthcareProfessional.value is modified progressively (using .push) inside an async function (updateHealthcareProfessional).
This creates a possible race condition where the ref’s state might be overwritten if there are multiple updates

Image

# Success
The array of relationships has to be built safely without any potential race condition issues.

### Requirements

- [ ]  Avoid reading and writing facilitiesRelationsToSelectedHealthcareProfessional.value in the same async scope

# Implementation notes
Replace createFacilitiesRelationArray() (which currently pushes directly into the ref) with a pure function that returns a Relationship
Use .map() to build the list of relationships in updateHealthcareProfessional, then assign the array to facilitiesRelationsToSelectedHealthcareProfessional.value

Ghezz9:
✅ Resolves #1129
- [x] PR title follows
- [x] @NabbeunNabi has been selected for review

## 🔧 What changed
This PR fixes a race condition in the logic that builds `facilitiesRelationsToSelectedHealthcareProfessional` inside the `updateHealthcareProfessional` function.

Previously, the code used `.forEach()` combined with `ref.value.push(...)`, which incrementally mutated a reactive array inside an async context. This approach was unsafe  because lead to race conditions, especially when `.value` was read and written across separate operations.

Replaced `.forEach()` and `.push()` with a single `.map(...)` assignment to ensure atomic update of the array.

Removed all direct mutations like .push() or repeated .value access.

Additionally, the code was further updated to avoid reading `.value` after await statements, which previously still caused ESLint to raise warnings.

## 📸 Screenshots

-   ### Before
V.1
image
image

-   ### After
V.1
image
image

V.2 (No Es-Lint error)
image
image
image


stores/healthcareProfessionalsStore.ts
```diff
@@ -88,19 +88,19 @@ export const useHealthcareProfessionalsStore = defineStore(
...
             const updateHealthcareProfessionalInput: MutationUpdateHealthcareProfessionalArgs = {
                 id: selectedHealthcareProfessionalId.value,
                 input: {
                     acceptedInsurance: healthcareProfessionalSectionFields.acceptedInsurance,
                     degrees: healthcareProfessionalSectionFields.degrees,
-                    facilityIds: facilitiesRelationsToSelectedHealthcareProfessional.value.length
-                      > 0
-                        ? facilitiesRelationsToSelectedHealthcareProfessional.value
-                        : undefined,
+                    facilityIds: relations.length > 0 ? relations : undefined,
```
NabbeunNabi, line 103: @Ghezz9 If you do it this way you can get rid of `facilitiesRelationsToSelectedHealthcareProfessional` on line 52 and 53. These are not necessary anymore. And I checked through the project and is not being used within a component so you can remove the export on line 240. You can double check that it isn't used in other files as well. Since the relations are all contained within the update function it is fine to have it just be a variable here and you can clean up the unused code and variables. Let me know if you have any questions.

stores/healthcareProfessionalsStore.ts
```diff
@@ -88,19 +88,19 @@ export const useHealthcareProfessionalsStore = defineStore(
 
             const facilitiesForRelationshipCreationArray = selectedFacilities.value
 
+            let relations: Relationship[] = []
```
NabbeunNabi, line 91: based on my other comment you could leave this as relations or use the more descriptive name of the unnecessary variable here with `facilitiesRelationsToSelectedHealthcareProfessional`. I think in this case maybe updating this one with the descriptive name and removing the other one might be helpful

NabbeunNabi (changes requested):
@Ghezz9 Great work. Looks like you fixed the problem. Just a couple clean up things mentioned

stores/healthcareProfessionalsStore.ts
Ghezz9: I've also deleted each ref .value = [].
Now if we use a local variable we just need to assign 
image

I didn't see any race condition after last commit.
Thankyou

theyokohamalife:
Since the PR is probably good to go after implementing the changes @NabbeunNabi suggested, I have a few small suggestions on how to improve the PR description.

It's really good to be descriptive in your PR, but you should avoid using versioning to describe your commits. It's better and easier for reviewers to understand if you edit the PR description to reflect the final solution you came up with.

Another way to improve readability is to wrap your code in code blocks. Variable or function names like `facilitiesRelationsToSelectedHealthcareProfessional` stand out more and become easier to read. For long code blocks, you should always use backticks to format the code like this:

```ts
const someLongBlockOfCode = "Look at how pretty I am!"
```

Try to copy/paste the code into code blocks like these instead of using screenshots. Screenshots are usually reserved for UI changes.

NabbeunNabi (approved):
@Ghezz9 NICE!

theyokohamalife (approved):
LGTM but I think you can shorten your PR description to just include the final changes that were made. 🙏🏾

stores/healthcareProfessionalsStore.ts
```diff
@@ -89,18 +87,21 @@ export const useHealthcareProfessionalsStore = defineStore(
...
+
             const updateHealthcareProfessionalInput: MutationUpdateHealthcareProfessionalArgs = {
                 id: selectedHealthcareProfessionalId.value,
                 input: {
                     acceptedInsurance: healthcareProfessionalSectionFields.acceptedInsurance,
                     degrees: healthcareProfessionalSectionFields.degrees,
-                    facilityIds: facilitiesRelationsToSelectedHealthcareProfessional.value.length
-                      > 0
-                        ? facilitiesRelationsToSelectedHealthcareProfessional.value
+                    facilityIds: facilitiesRelationsToSelectedHealthcareProfessional.length > 0
```
ermish, line 103: tip: you can just check `.length` without the 0 check
```suggestion
                    facilityIds: facilitiesRelationsToSelectedHealthcareProfessional.length
```

ermish (approved):
looks great! nice work @Ghezz9 🚀 
I added a small nit that is optional if you want to fix before merging

PR #1131, opened 2025-04-17
fix: position doctors modal correctly

### LINKED ISSUE #1077: bug: the doctor modal is not positioned properly
# Problem
The modal isn't positioned properly (top left corner justified) near the results. Instead it's in the center screen. this is bad because it's not visually associated near the results list that was clicked. 

Image

# Implementation notes
Note #1078 and #1079 are associated

### LINKED ISSUE #1078: bug: modal background is not properly positioned
# Problem
The background blur on the modal when you click on a result in desktop mode doesn't stretch properly to the map container. 
It should be the full width and height. It should also only z-index over the map container and not the whole page.

Zugim:
✅ Resolves #1077 Resolves #1078

## 🔧 What changed
The doctor modal is now positioned correctly to the left of the search results. It's z-index no longer interferes with the search results or top navigation bar but still blocks interaction with the map unless the close button is pressed. Works in Firefox, Chrome and Safari.

## 🧪 Testing instructions
Pull this branch to your local machine and select a search result from the sidebar, the doctor modal should now display correctly over the map.

## 📸 Screenshots

-   ### Before
image

-   ### After
image

-   ### Before (Mobile)
image

-   ### After (Mobile)
image


NabbeunNabi (approved):
Awesome! Great work! It looks much better

theyokohamalife:
I confirmed that it looks good on desktop for Safari and Firefox. However, I noticed that it's still off center for mobile (tested with iPhone 14 pro max). Since #1077 isn't specific to desktop, could you include a fix for mobile in this PR as well?
Screenshot 2025-04-17 at 9 27 43 PM

Zugim:
Sure I can try and get it working on mobile devices too. Just to confirm in the case of mobile devices should the modal cover the entire screen? As on mobile the map is pretty small so limiting the modal to the map is probably not going to look great

Zugim:
Doctor modal should now also display correctly on mobile devices. Below screenshot is using Firefox devtools on an iPhone 12/13 + pro

image

NabbeunNabi:
@Zugim Looks good. BTW just a nit this is a fix since it was for a bug. Not a feat. So your PR title should probably say `fix`

theyokohamalife (approved):
LGTM 🚢

NabbeunNabi (changes requested):
@Zugim Is there a way actually to make it so that the modal does not cover the map above it at all. And also change the size of the text. It is a bit squeezed to the side

ermish:
to @NabbeunNabi 's point earlier, in the ticket #1077, we actually don't want it to be centered. It was originally in the top left hand corner. The main reason is that if it's in the center, it blocks the  map pin that they are currently looking at and they can't see / select other ones. Do you want to move it in this PR or we could have the positioning in a separate PR since this already is much better :)

Zugim:
Sorry I somehow missed these requests. I'll get on this. I think I understand roughly how you want it

Zugim:
@ermish were you envisioning the left aligned modal being flush up against the left nav or do we still want some padding? I was thinking flush against the left nav and filling the viewport height might look good

Zugim:
The modal no longer covers the map on desktop views, instead it is aligned to the left much like in Google Maps
image

On mobile the modal should now display full screen consistently. I think this is probably best on smaller screens as when the modal is open the visible portion of the map is too small to be helpful to the user.
 
image

theyokohamalife:
@Zugim Hey! I see that you requested review. Is this meant to still be a draft?

PR #1168, opened 2025-05-07
fix: fix track leftnavbar

Ghezz9:
✅ Resolves #[Issue number]

## 🔧 What changed

This PR fixes the issue where the active section highlighting in both `ModSubmissionLeftNavBar.vue` and `ModHealthcareProfessionalLeftNavBar.vue` was not updating during manual scroll. Previously, clicking on a nav button worked (via `scrollToSectionOfForm()`), but scrolling the form manually did not update the `activeSection`, causing a mismatch in UI highlighting.

## 📸 Screenshots

-   ### Before

-   ### After

https://github.com/user-attachments/assets/bf6f9bc4-7fc2-4266-81ee-f7df69c8825a

### DIFF
--- a/components/ModSubmissionLeftNavbar.vue
+++ b/components/ModSubmissionLeftNavbar.vue
@@ -10,7 +10,7 @@
                 :class="{ 'bg-secondary': activeSection === ModSubmissionLeftNavbarSectionIDs.ContactInformation,
                           'bg-primary-inverted': activeSection !== ModSubmissionLeftNavbarSectionIDs.ContactInformation }"
                 class="w-full py-4 my-2 text-sm text-start pl-2 rounded border-b-2 border-tertiary-bg"
-                @click="scrollToSectionOfForm(ModSubmissionLeftNavbarSectionIDs.ContactInformation)"
+                @click="handleNavClick(ModSubmissionLeftNavbarSectionIDs.ContactInformation)"
             >
                 {{ $t("modPanelSubmissionLeftNavbar.contactInformation") }}
             </button>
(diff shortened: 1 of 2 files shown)


Ghezz9:
As you can see from the video, the section tracker is not working as accurately as I’d like. I believe the issue lies with the `thresholdValue: number = 0.70` parameter in the IntersectionObserver. I tried tweaking the value and while increasing or decreasing it improves accuracy for some section IDs, it makes it worse for others.
One possible solution could be to set a different threshold for each section individually.
However, this approach seems unreliable and not good because if the layout will changes or new sections are added, we would need to manually adjust each threshold again. This doesn't feel like a scalable or maintainable solution.

NabbeunNabi:
@Ghezz9 is there a way to calculate or get the height of the section. And the write a formula that calculates the threshold based on the height of that element with the screen size with [Window.innerHeight](https://developer.mozilla.org/en-US/docs/Web/API/Window/innerHeight)

ermish:
@NabbeunNabi I'll review this one after you've read through it. You're familiar with a lot of the scrolling details

NabbeunNabi (changes requested):
I mentioned in a comment you maybe missed but here is what I wrote

@Ghezz9 is there a way to calculate or get the height of the section. And the write a formula that calculates the threshold based on the height of that element with the screen size with [Window.innerHeight](https://developer.mozilla.org/en-US/docs/Web/API/Window/innerHeight)

NabbeunNabi:
@Ghezz9 Are you still working on this. Might not be necessary anymore if we end up redesigning the moderation panel with @ljbroersen

PR #1191, opened 2025-05-15
refactor: call init onMounted in index and moderation pages

### LINKED ISSUE #1190: Init auth on dev and testing to avoid possible async problems
# Problem
because the `init` function in the `authStore.ts` file is async, we need to await it when using it on dev and testing in the index page and moderation page. If not it could lead to being logged out unintentionally when in development.

# Success
Init is called on mount in the two routes and exported from authStore
### Requirements

- [x] The `init` function instead of being called in the store is exported from the `authStore.ts`
- [x] The `init` function is called `onMounted` in the `index.vue` page
- [x] The `init` function is called `onMounted` in the `moderation.vue` page

NabbeunNabi:
✅ Resolves #1190

## 🔧 What changed
The `init` function being used for testing the `authStore` due to not being await as @ermish was worried about, was leading to some unintended side effects within development.

This moves to it being called in `index.vue` and `moderation.vue` to have it be awaited on mount. This will allow users to consistently test the moderation in dev while not having problems within the production environment

## 🧪 Testing instructions
Run the server locally on dev and log in. You can access both pages when logged in with the data and also it will show you logged in when you visit again assuming the cookie has not been removed

### DIFF
--- a/pages/index.vue
+++ b/pages/index.vue
@@ -21,9 +21,13 @@
 
 <script setup lang="ts">
 import { useScreenOrientation } from '~/utils/useScreenOrientation'
-import { useModalStore } from '~/stores/modalStore'
+import { useAuthStore } from '#imports'
 
-useModalStore()
+const authStore = useAuthStore()
 
 const { isPortrait } = useScreenOrientation()
+
+onMounted(async () => {
+    await authStore.init()
+})
 </script>
--- a/pages/moderation.vue
+++ b/pages/moderation.vue
@@ -74,6 +74,8 @@ const router = useRouter()
 const doesTheUserHaveAccess: Ref<boolean> = ref(true)
 
 const checkIfUserIsLoggedIn = async () => {
+    // init the auth to see if we are in dev/testing mode or prod
+    await authStore.init()
     // This promise is here to make the Suspense component work.
     // It doesn't do anything, but <Suspense> requires an awaited setup method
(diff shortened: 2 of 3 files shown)


pages/index.vue
```diff
@@ -21,9 +21,13 @@
 
 <script setup lang="ts">
 import { useScreenOrientation } from '~/utils/useScreenOrientation'
-import { useModalStore } from '~/stores/modalStore'
```
theyokohamalife, line 24: Why is the `useModalStore` removed here? Is it not being used?

pages/moderation.vue
```diff
@@ -74,6 +74,8 @@ const router = useRouter()
 const doesTheUserHaveAccess: Ref<boolean> = ref(true)
 
 const checkIfUserIsLoggedIn = async () => {
+    // init the auth to see if we are in dev/testing mode or prod
```
theyokohamalife, line 77:
```suggestion
    // Inits the auth to see if we are in dev/testing mode or prod.
```

This is a small nit. We should determine and stick to our conventions for leaving comments in code. Like if there should always be a space after the `//` and if it should start with an uppercase and end with a period. You can see the inconsistencies in the two comments below this one. Each one is different.

theyokohamalife (commented):
Looks good! Just left a question and small comment.

ermish:
closing as this will be replaced by #1183

PR #1196, opened 2025-05-15
feat: add limit to fetch 100 submissions

NabbeunNabi:

## 🔧 What changed
We added a filter to send the limit to the backend so that way it does not default to fetching 20 submissions. This will now fetch 100 submissions so moderators for now can approve or reject the submissions we have while we update the backend and frontend to include pagination.


theyokohamalife (approved):
🚢

PR #1201, opened 2025-05-15
fix: add event default to prevent default action

NabbeunNabi:
✅ Resolves #[Issue number]

## 🔧 What changed
The default browser action was causing weird side events for events when people were clicking if a modal was ever opened.
This prevents that from happening

## 📸 Screenshots

-   ### Before

-   ### After


ermish (approved):
always use prevention

PR #1202, opened 2025-05-15
fix: fix ref not being reset to false after submission update

NabbeunNabi:

## 🔧 What changed
Our `resetModalRefs` function that resets all the parts of the top bar and the modal was missing one of the values that needs to be reset to false. This has been fixed so the watch function that always watches when the dom changes does not auto update

Also the watch only looks at the necessary value of `updatingSubmissionFromTopBar`


theyokohamalife (approved):
🚢

PR #1210, opened 2025-05-17
feat: integrate pagination for facilities

### LINKED ISSUE #1039: Add pagination on the Moderation Dashboard
# Problem
Write now we are currently displaying all the submissions, facilities, and healthcare professionals depending on what is selected. We want the user to have a list of 20 and there to be pagination with what total there is and how many they see. (i.e. 1-20 out of 49)

# Success
Pagination is added for the displayed data on the `ModSubmissionListContainer.Vue` with a total count at the bottom and which ones are being displayed for 20 each page.

### Requirements

- [ ] Add a count total at the bottom depending on if you are viewing submissions, facilities, or healthcare professionals
- [ ] Add pagination so that way you only see 20 or the final amount. If seeing the final amount there is only a left arrow to go back. If displaying 1-20 you only see the right arrow
- [ ] Tests are adding so that this functionality exists in cypress as if a user is accessing this page and trying to go through the pagination

# Implementation notes
- Path to the [`ModSubmissionListContainer.Vue`](https://github.com/ourjapanlife/findadoc-web/blob/main/components/ModSubmissionListContainer.vue)
- [Cypress test files](https://github.com/ourjapanlife/findadoc-web/blob/main/test/cypress/e2e/moderationDashboard.cy.ts)
- If the logic in this PR gets too long, feel free to break it up into 3. One for each facilities, healthcare professionals, and submissions

Ghezz9:
✅ Resolves #1039

## 🔧 What changed
This PR integrates the pagination component into the facilities submission list, using the existing `ModPagination.vue` component that is already available in the base branch.
Only the `ModSubmissionListContainer.vue` file is modified in this PR. All other related components (pagination logic, SVG icons) already exist in `davide/mod-dashboard-pagination`.

this PR has a base branch that reffear to this PR #1236

## 📸 Screenshots

-   ### Before

-   ### After

image


NabbeunNabi (changes requested):
@Ghezz9 This looks like the same as the previous one I reviewed. If the changes are all the same and are relevant to the same component and ticket without a lot of changes you can put this into one PR. Describe it well. And the PR would be to introduce pagination on the Dashboard since the functionality for display is the same for facilities/submissions/and healthcare professionals. You are not changing much in then except for displaying them. If it was different functionality for each I would break it up. But it looks that these have the same changes

components/ModSubmissionListContainer.vue
```diff
@@ -108,15 +116,19 @@
 </template>
 
 <script setup lang="ts">
-import { computed, onMounted } from 'vue'
+import { computed, onMounted, ref, type Ref } from 'vue'
 import { SelectedModerationListView, useModerationSubmissionsStore } from '~/stores/moderationSubmissionsStore'
 import { useHealthcareProfessionalsStore } from '~/stores/healthcareProfessionalsStore'
 import { useFacilitiesStore } from '~/stores/facilitiesStore'
+import ModPagination from '~/components/ModPagination.vue'
```
NabbeunNabi, line 123: You should not need to import the component. They should be auto imported. The ModSearchBar for some reason has been having a problem where it isn't. But yours should be. I would check and take out the import

NabbeunNabi (changes requested):

PR #1215, opened 2025-05-17
refactor: centralize Japanese prefecture lists

### LINKED ISSUE #1208: Move Japanese prefecture lists into one file and refactor components using said lists
# Problem
Every time a component needs the lists of prefectures in Japan, we are hardcoding these lists into that file (e.g.: `ModEditFacilitySection.vue`, `ModEditSubmissionForm.vue`). This is not efficient.

Image

# Success
Abstract the lists of prefectures in Japan and refactor all of the components that have them to point to the file the lists are moved to.

### Requirements

- [ ] Move both `listPrefectureJapanEn` and `listPrefectureJapanJa` collectively into a new file
- [ ] Refactor all components that currently have these lists hardcoded

Ghezz9:
✅ Resolves #1208

## 🔧 What changed

`ModEditFacilitySection.vue`, `ModEditSubmissionForm.vue` now import the two list of japan prefecture from `/stores/prefectureList`

## 📸 Screenshots

-   ### Before

-   ### After

Nothing different

image

As you can see before and after there is still one "bug" Prefecture Eng field is empty before choosing one differente to Prefecture Jap


NabbeunNabi (changes requested):
@Ghezz9 good improvement so not to reuse the locations of this any time we will need it in a dropdown list.

Have you checked to make sure this is the only place it is used in the whole website?

Also stores are for pinia stores as a folder. This should more appropriately be in a utils file or part of a store.

We already have a `locationStore.ts` so I would suggest moving it there and exported it from the store. Then initializing the store within the files necessary and then using the variable.

stores/locationsStore.ts
```diff
@@ -6,6 +6,26 @@ import { useLoadingStore } from './loadingStore.js'
...
+export const listPrefectureJapanJa: string[] = [
+    '北海道', '青森県', '岩手県', '宮城県', '秋田県', '山形県', '福島県',
+    '茨城県', '栃木県', '群馬県', '埼玉県', '千葉県', '東京都', '神奈川県',
+    '新潟県', '富山県', '石川県', '福井県', '山梨県', '長野県', '岐阜県',
+    '静岡県', '愛知県', '三重県', '滋賀県', '京都府', '大阪府', '兵庫県',
+    '奈良県', '和歌山県', '鳥取県', '島根県', '岡山県', '広島県', '山口県',
+    '徳島県', '香川県', '愛媛県', '高知県', '福岡県', '佐賀県', '長崎県',
+    '熊本県', '大分県', '宮崎県', '鹿児島県', '沖縄県'
+]
+
```
NabbeunNabi, line 28: @Ghezz9 Let's move this to the bottom of the file similar to how it is here

https://github.com/ourjapanlife/findadoc-web/blob/85aecf5db3f9e39dcbd66e483549af7cd91a3595/stores/localeStore.ts#L56

That we put most constants within store files at the bottom

NabbeunNabi (changes requested):

NabbeunNabi (approved):
LGTM onto @ermish

theyokohamalife (commented):
Off to a good start! The [ModCreateFacilitySection component](https://github.com/ourjapanlife/findadoc-web/blob/85aecf5db3f9e39dcbd66e483549af7cd91a3595/components/ModCreateFacilitySection.vue#L311) also has lists that need to be refactored.

theyokohamalife (approved):
Thank you! 😎

PR #1233, opened 2025-05-19
refactor: change facility section of form for editing submission

### LINKED ISSUE #1232: Update `ModEditFacilitySection.vue` to conditionally render parts not necessary for `ModEditSubmissionForm.vue`
# Problem
Right now there are certain parts of the facility professional section that should not be displayed within the edit submission form.

# Success
The section is updated with logic that is only necessary for the parts in the parts in each section so it can be used on two different screens

### Requirements

- [x] Logic is updated to conditionally render correct sections in each active screen

NabbeunNabi:
✅ Resolves #1232 

## 🔧 What changed
Before this section was only prepped for facilities that already existed in our database. Now it conditionally renders necessary sections for the facility but that should be displayed differently in the `ModEditSubmissionForm.vue` there when that form ends up using this section

## 📸 Screenshots
- No UI/UX changes on this form


components/ModEditFacilitySection.vue
```diff
@@ -1,10 +1,11 @@
 <template>
-    <Loader />
+    <Loader v-if="moderationScreenStore.editFacilityScreenIsActive()" />
     <div v-if="isFacilitySectionInitialized">
         <div
             class="mod-facility-section"
         >
             <h1
+                v-if="moderationScreenStore.editFacilityScreenIsActive()"
```
theyokohamalife, line 8: Why are we adding this to the `h1` instead of the parent `div`? 🤔

theyokohamalife (approved):
LGTM

PR #1236, opened 2025-05-19
feat: add modPagination components

Ghezz9:
✅ Resolves[#1039](https://github.com/ourjapanlife/findadoc-web/issues/1039)

## 🔧 What changed
I've created the component `ModPagination.vue` and add a pagination for `Submissions` `Facilities` `Healthcare Professionals`

## 📸 Screenshots

-   ### Before

-   ### After

finalFac

es 5
es 1
image


components/ModPagination.vue
```diff
@@ -0,0 +1,48 @@
+<template>
+    <div class="flex items-center justify-center gap-4 mt-4">
+        <!-- Left arrow -->
+        <button
+            v-if="totalPages >0 "
```
NabbeunNabi, line 5:
```suggestion
            v-if="totalPages"
```

components/ModPagination.vue
```diff
@@ -0,0 +1,48 @@
...
+            :disabled="currentPage === 1 || totalPages === 0"
+            class="w-10 h-10 flex items-center justify-center rounded bg-gray-400 disabled:opacity-30"
+            @click="emit('update:currentPage', currentPage - 1)"
+        >
```
NabbeunNabi, line 9: I am debating stylistically how this should look. @Ntlja or @sylentwulf81 do either of you have any opinions on how pagination should look. Especially if it is something we could reuse? Do the arrows show just disabled or muted colors? Is it better to have a first, last and then arrows for previous and next?

components/ModPagination.vue
```diff
@@ -0,0 +1,48 @@
...
+            <SVGSVGLeftArrow class="w-4 h-4 text-white" />
```
NabbeunNabi, line 10: Only use `SVG` once in the described import so `SVGLeftArrow`

components/ModPagination.vue
```diff
@@ -0,0 +1,48 @@
...
+            {{ $t("modPagination.page") }} {{ currentPage }} {{ $t("modPagination.of") }} {{ totalPages }}
+        </p>
+        <!-- Right arrow -->
+        <button
+            v-if="totalPages >0 "
+            :disabled="currentPage === 1 || totalPages === 0"
+            class="w-10 h-10 flex items-center justify-center rounded bg-gray-400 disabled:opacity-30"
+            @click="emit('update:currentPage', currentPage + 1)"
+        >
+            <SVGSVGRightArrow class="w-4 h-4 text-white" />
```
NabbeunNabi, line 26: ♻️

components/ModPagination.vue
```diff
@@ -0,0 +1,48 @@
...
+        </button>
+    </div>
+</template>
+
+<script setup lang="ts">
+import { computed } from 'vue'
+import SVGSVGLeftArrow from '~/assets/icons/left-arrow-pagination.svg'
+import SVGSVGRightArrow from '~/assets/icons/right-arrow-pagination.svg'
```
NabbeunNabi, line 34: ♻️

i18n/locales/cn.json
```diff
@@ -420,5 +420,9 @@
     "firstName": "First name",
     "middleName": "Middle name",
     "nameLocale": "Name locale"
+  },
+  "modPagination": {
+        "page": "Page",
+        "of": "of"
```
NabbeunNabi, line 426: Instead of having multiple key value pairs you can actually inject a variable into the translatable string. 

Here are the docs for you https://vue-i18n.intlify.dev/guide/essentials/syntax

i18n/locales/de.json
```diff
@@ -420,5 +420,9 @@
     "firstName": "First name",
     "middleName": "Middle name",
     "nameLocale": "Name locale"
+  },
+  "modPagination": {
+        "page": "Page",
+        "of": "of"
```
NabbeunNabi, line 426: ♻️

i18n/locales/en.json
```diff
@@ -411,5 +411,9 @@
         "firstName": "First name",
         "middleName": "Middle name",
         "nameLocale": "Name locale"
+    },
+    "modPagination": {
+        "page": "Page",
+        "of": "of"
```
NabbeunNabi, line 417: ♻️

i18n/locales/fr.json
```diff
@@ -424,5 +424,9 @@
     "firstName": "First name",
     "middleName": "Middle name",
     "nameLocale": "Name locale"
+  },
+  "modPagination": {
+        "page": "Page",
+        "of": "of"
```
NabbeunNabi, line 430: ♻️

i18n/locales/it.json
```diff
@@ -411,5 +411,9 @@
     "firstName": "First name",
     "middleName": "Middle name",
     "nameLocale": "Name locale"
+  },
+  "modPagination": {
+        "page": "Pagina",
+        "of": "di"
```
NabbeunNabi, line 417: ♻️

i18n/locales/ja.json
```diff
@@ -421,5 +421,9 @@
     "firstName": "First name",
     "middleName": "Middle name",
     "nameLocale": "Name locale"
+  },
+  "modPagination": {
+        "page": "Page",
+        "of": "of"
```
NabbeunNabi, line 427: ♻️

i18n/locales/pt.json
```diff
@@ -420,5 +420,9 @@
     "firstName": "First name",
     "middleName": "Middle name",
     "nameLocale": "Name locale"
+  },
+  "modPagination": {
+        "page": "Page",
+        "of": "of"
```
NabbeunNabi, line 426: ♻️

i18n/locales/ru.json
```diff
@@ -420,5 +420,9 @@
     "firstName": "First name",
     "middleName": "Middle name",
     "nameLocale": "Name locale"
+  },
+  "modPagination": {
+        "page": "Page",
+        "of": "of"
```
NabbeunNabi, line 426: ♻️

i18n/locales/tl.json
```diff
@@ -411,5 +411,9 @@
     "firstName": "First name",
     "middleName": "Middle name",
     "nameLocale": "Name locale"
+  },
+  "modPagination": {
+        "page": "Page",
+        "of": "of"
```
NabbeunNabi, line 417: ♻️

i18n/locales/vi.json
```diff
@@ -411,5 +411,9 @@
     "firstName": "First name",
     "middleName": "Middle name",
     "nameLocale": "Name locale"
+  },
+  "modPagination": {
+        "page": "Page",
+        "of": "of"
```
NabbeunNabi, line 417: ♻️

NabbeunNabi (changes requested):

components/ModPagination.vue
sylentwulf81: Without having time to sit and look through everything, my preference would be something like: 

<<  < 1 2 ... # >  >>

If we have three or less total pages, we show clickable page numbers along with to first, to last, and next / prev. 

If no next, grey out 'next' but leave it presented. Same behaviour for 'prev'. 

If at last page, grey out 'last' but leave it presented. Same behaviour for 'first'.

Here's where I'm open to discussion - I was thinking if we have more than three pages, we show the user the page they're on, +/- 1. So if the user is on page 2, they would see clickables for page 1, 2, 3. The only caveat is how we present the total number of pages to the user cleanly. 

I could see us having many, many pages at some point which might get unwieldy if the user doesn't know how many pages there are total. Some sites simply don't show total number of pages if the number is unwieldy or doesn't add value to the user experience. 

My kneejerk answer is show the user current page +/- 1 or 2, present navigation arrows, and worry about a cleaner fix if and when we have double digit or greater pages to manage. 

Also to answer your question more specifically, disabled should be default, since we're dealing with a broad user base that may not intuitively grasp non-standard colors. Grey is kindof the universal 'disabled' or 'non-interactive' indication color. Also, in keeping with our previous thinking around accessibility, grey is visible to all color-blindness spectrums, we'd have to do some digging for each color we add (assuming you mean the arrow colors would follow the theming?) to see if it's broadly visible.

sylentwulf81: To my eye this looks good, and I think the average user would find it intuitive, but I might be too close to it since it was my idea. @NabbeunNabi ? Thoughts?

NabbeunNabi: @sylentwulf81 @Ghezz9 I like it a lot

components/ModSubmissionListContainer.vue
```diff
@@ -34,7 +34,9 @@
                         :to="`/moderation/edit-submission/${submission.id}`"
                         class="grid grid-cols-subgrid col-span-4 bg-primary-text-muted p-1 hover:bg-primary"
                     >
-                        <span class="text-start">{{ index + 1 }}</span>
+                        <span class="text-start">
+                            {{ (currentSubmissionsPage - 1) * submissionsPerPage + index + 1 }}
```
NabbeunNabi, line 38: Could you write a function with a good name to call here so we know what we are doing exactly here for readability?

components/ModSubmissionListContainer.vue
```diff
@@ -62,7 +70,9 @@
                         :to="`/moderation/edit-facility/${facility.id}`"
                         class="grid grid-cols-subgrid col-span-4 p-1 hover:bg-primary"
                     >
-                        <span class="text-start">{{ index + 1 }}</span>
+                        <span class="text-start">
+                            {{ (currentFacilitiesPage - 1) * facilitiesPerPage + index + 1 }}
```
NabbeunNabi, line 74: ♻️ use function here

components/ModSubmissionListContainer.vue
```diff
@@ -91,7 +107,9 @@
                         :to="`/moderation/edit-healthcare-professional/${healthcareProfessional.id}`"
                         class="grid grid-cols-subgrid col-span-4 p-1 hover:bg-primary"
                     >
-                        <span class="text-start">{{ index + 1 }}</span>
+                        <span class="text-start">
+                            {{ (currentHealthcarePage - 1) * healthcarePerPage + index + 1 }}
```
NabbeunNabi, line 111: ♻️ use function here

components/ModSubmissionListContainer.vue
```diff
@@ -100,6 +118,12 @@
                     </NuxtLink>
                 </div>
             </div>
+            <ModPagination
+                :current-page="currentHealthcarePage"
+                :total-items="totalHealthcareProfessionals"
+                :items-per-page="healthcarePerPage"
+                @update:current-page="val => currentHealthcarePage = val"
```
NabbeunNabi, line 125: You could probably here also call a function that says specifically what this is

components/ModPagination.vue
```diff
@@ -0,0 +1,124 @@
...
+
+        for (let i = start; i <= end; i++) {
+            pages.push(i)
+        }
+
+        if (props.currentPage < totalPages.value - 1) {
+            if (props.currentPage < totalPages.value - 2) pages.push('...')
+            pages.push(totalPages.value)
+        }
+    }
```
NabbeunNabi, line 119: Can you change the for loop to a map and then set the pages to that value.

Also rework the logic that returns out instead of using nested ifs. If this is unclear let me know.

So for example you could do
```
if (props.currentPage < totalPages.value - 1) {
            if (props.currentPage < totalPages.value - 2) pages.push('...')
            pages.push(totalPages.value)
        }
        ```
        As two ifs with the `&&` operator. Also It seems like there are a lot of cases that maybe you could use current page to make a switch case instead of ifs for readability

**Review by NabbeunNabi — CHANGES_REQUESTED** [review]

File: components/ModPagination.vue [hunk 21]
```diff
@@ -0,0 +1,119 @@
...
+import SVGFirstPage from '~/assets/icons/first-page-pagination.svg'
+import SVGLastPage from '~/assets/icons/last-page-pagination.svg'
+
+const props = defineProps<{
+    currentPage: number
+    totalItems: number
+    itemsPerPage: number
+}>()
+
+const dots = '...'
```
NabbeunNabi, line 82: `ellipsisToShowMoreContent`

NabbeunNabi (approved):
@Ghezz9 just one small change and then @ermish can take a look. He will be more nitpicky about using push possibly

components/ModPagination.vue
```diff
@@ -0,0 +1,119 @@
+<template>
+    <div class="flex items-center justify-center gap-2 mt-4">
+        <!-- First page -->
+        <button
+            :disabled="currentPage === 1"
+            class="w-10 h-10 flex items-center justify-center rounded bg-gray-400 disabled:opacity-30 disabled:cursor-not-allowed"
```
ermish, line 6: instead of `bg-gray` and `text-white` let's use our theme colors. 
Otherwise, the gray won't look right when we use a dark theme or high-contract them for example

components/ModPagination.vue
```diff
@@ -0,0 +1,119 @@
...
+            @click="goToPage(1)"
+        >
+            <SVGFirstPage class="w-4 h-4 text-white" />
+        </button>
+
+        <!-- Prev page -->
+        <button
+            :disabled="currentPage === 1"
+            class="w-10 h-10 flex items-center justify-center rounded bg-gray-400 disabled:opacity-30 disabled:cursor-not-allowed"
```
ermish, line 15: ♻️

ermish: there's some other occurences, so I'll just leave the rest to you :)

components/ModPagination.vue
```diff
@@ -0,0 +1,119 @@
...
+    }
+}
+
+const visiblePages = computed(() => {
+    const total = totalPages.value
+    const current = props.currentPage
+    const pages: (number | string)[] = []
+
+    if (total <= 3) {
+        return Array.from({ length: total }, (_, i) => i + 1)
```
ermish, line 103: this is a bit confusing to read what this does at first glance. Can you improve the readability a bit?

components/ModPagination.vue
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    if (current > 2) pages.push(1)
```
ermish, line 106: also for these. Why are we pushing pages if current is changing?
you can use a variable to help explain why like
```
const thisIsTheReasonWhy = current > 3
if(thisIsTheReasonWhy)....
```

ermish:
At a high level, I think this design will be fine. 
Personally, I think just a left and right button along with `1-50/225 results` would be a bit cleaner and simpler. This is similar to how gmail pagination works for example  (but I understand you already built all of this out) 

Whether you keep this design or change it, I think the total results would be a valuable addition here. Could you make an issue for this?

ermish (changes requested):
this is great! Added some comments

components/ModPagination.vue
```diff
@@ -0,0 +1,119 @@
+<template>
```
ermish, line 1: I just realized this is the old file. Let's delete this one and we're good to go!

PR #1239, opened 2025-05-20
fix: reduce method only runs when submissions are present

### LINKED ISSUE #1238: bug: moderation dashboard not loading when backtracking from create or edit pages
# Problem
When navigating back from a create or edit page in moderation this error appears in the console:

Image

It comes from a function which attempts to sort the submission counts and then display them in the left navbar. When it does this to an empty Submissions array it causes the above error and doesn't allow the page to change back to the moderation dashboard.

# Success
The error no longer appears and navigating back to the dashboard is possible.

# Implementation notes
Only running the function causing the error when the array of Submissions has submissions in it may prevent the error.

ConwayDecker:
✅ Resolves #1238 

## 🔧 What changed
A conditional has been added to the createCountsForSubmissionListView which prevents it from running if the provided array of Submissions is empty. Optional chaining operators and nullish coalescing operators have also been added to where the sorted submission counts appear to prevent potential TypeScript errors or blank spaces if they are undefined.

## 📸 Screenshots

-   ### Before
https://github.com/user-attachments/assets/fde7a1a6-c61c-4fc8-857b-4f6c3aafc0ae

-   ### After
https://github.com/user-attachments/assets/4551b327-445e-4f7e-9a98-8b6cb9685c64

### DIFF
--- a/components/ModDashboardLeftNavbar.vue
+++ b/components/ModDashboardLeftNavbar.vue
@@ -94,7 +94,17 @@ const updateSelectedDashboardView = (event: Event) => {
 }
 
 function createCountsForSubmissionListView(submissions: Submission[]) {
-    return submissions.reduce((statusCountObject, submission) => {
+    const submissionStatusCounts = {
+        forReviewCount: 0,
+        approvedCount: 0,
+        rejectedCount: 0
+    }
+
+    if (!submissions.length) {
+        return submissionStatusCounts
+    }
+
+    submissions.reduce((statusCountObject, submission) => {
         const isNewSubmission = !submission.isRejected && !submission.isApproved && !submission.isUnderReview
         if (submission.isUnderReview) {
             statusCountObject.forReviewCount += 1
@@ -109,11 +119,8 @@ function createCountsForSubmissionListView(submissions: Submission[]) {
             statusCountObject.forReviewCount += 1
         }
         return statusCountObject
-    }, {
-        forReviewCount: 0,
-        approvedCount: 0,
-        rejectedCount: 0
-    })
+    }, submissionStatusCounts)
+    return submissionStatusCounts
 }
 
 const autofillStatusCount = computed(() => createCountsForSubmissionListView(moderationSubmissionsStore.submissionsData))


components/ModDashboardLeftNavbar.vue
```diff
@@ -94,26 +94,28 @@ const updateSelectedDashboardView = (event: Event) => {
...
-        if (isNewSubmission) {
-            statusCountObject.forReviewCount += 1
-        }
-        return statusCountObject
-    }, {
-        forReviewCount: 0,
-        approvedCount: 0,
-        rejectedCount: 0
-    })
+    if (submissions.length > 0) {
```
NabbeunNabi, line 97: This is a good start! I would think for easier readability and for it to defaulting to the count as that is the behavior we want when there are submissions to just check if there is not a length and if so return. Otherwise return the reduce.

Something like 

```
if (!submissions.length) return 0
```

NabbeunNabi (changes requested):

components/ModDashboardLeftNavbar.vue
ConwayDecker: I added `if (!submissions.length) return undefined` instead of returning 0 because the references of `autofillStatusCount` in the template showed a possible error if it returns as 0 (and then tries to access rejectedCount or other properties). With undefined it will default to 0.

NabbeunNabi: Ahhh yes.

I actually meant to remove the 0.

Because you have the default parts up top.

It would need to return that object in the reducer.

Should just be undefined. Nice one

components/ModDashboardLeftNavbar.vue
```diff
@@ -94,6 +94,7 @@ const updateSelectedDashboardView = (event: Event) => {
 }
 
 function createCountsForSubmissionListView(submissions: Submission[]) {
+    if (!submissions.length) return undefined
     return submissions.reduce((statusCountObject, submission) => {
```
NabbeunNabi, line 98: @ConwayDecker i know this was something I did. But it is more common I have heard from people to set the map or something to a variable such as `countedSubmissions` then return that.

It can make the reduce more readable as to what it actually is instead of returning it directly.

Do you mind doing that here and atoning for past Will's unreadableness

NabbeunNabi (changes requested):

NabbeunNabi (changes requested):
Made a comment about the function. Almost there @ConwayDecker

theyokohamalife (approved):
Nice! ✨

PR #1241, opened 2025-05-22
fix: avoid race condition `facilitiesStore.ts`

### LINKED ISSUE #1186: Make update of `healthProfessionalsRelations` atomic to prevent ESLint warning
# Problem
The data in the `facilitySectionFields.healthProfessionalsRelations` array might be updated slowly or from different parts of the app. If `updateFacility()` runs before this data finishes updating, it might use an old or incomplete version of it. This can cause healthcare professionals that should be connected to a facility to be missed or saved incorrectly.

Image

# Success
The array of relationships has to be built safely without any potential race condition issues.

### Requirements

- [ ] Avoid reading and writing `facilitySectionFields.healthProfessionalsRelations` in the same async scope.

ljbroersen:
✅ Resolves #1186 

## 🔧 What changed
This PR updates the logic inside the `updateFacility()` function to unconditionally clear `facilitySectionFields.healthProfessionalsRelations` after a successful facility update.

By resetting the array once the update is confirmed, the code avoids potential race conditions and ensures a clean state for the next user interaction.

### DIFF
--- a/stores/facilitiesStore.ts
+++ b/stores/facilitiesStore.ts
@@ -11,6 +11,7 @@ import type { DeleteResult, Facility,
     Relationship } from '~/typedefs/gqlTypes'
 import { gqlClient, graphQLClientRequestWithRetry } from '~/utils/graphql'
 import type { ServerResponse } from '~/typedefs/serverResponse'
+import { arraysAreEqual } from '~/utils/arrayUtils'
 
 export const useFacilitiesStore = defineStore(
     'facilitiesStore',
@@ -214,7 +215,10 @@ export const useFacilitiesStore = defineStore(
                 selectedFacilityData.value = serverResponse.data
                 initializeFacilitySectionValues(serverResponse.data)
 
-                if (facilitySectionFields.healthProfessionalsRelations === healthProfessionalRelationsBeforeMutation) {
+                if (arraysAreEqual(
+                    facilitySectionFields.healthProfessionalsRelations,
+                    healthProfessionalRelationsBeforeMutation
+                )) {
                     facilitySectionFields.healthProfessionalsRelations = []
                 }
             }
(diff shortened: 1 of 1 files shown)


NabbeunNabi (changes requested):
This field is non existent on a facility for our types. We only use it on the frontend to track if users pick a HP to add or remove from a facility. Is there a way to guarantee that it is set back to an empty array when no errors come back and remove this race condition. I am not quite sure how this will consistently clear the array.

You can see [here](https://github.com/ourjapanlife/findadoc-web/blob/85aecf5db3f9e39dcbd66e483549af7cd91a3595/typedefs/gqlTypes.ts#L106) that is is only a type on the frontend we add but not part of the response or anything from the backend. We just want to make sure the relationships we send back are cleared.

So these relationships are calculated based on the healthcare professional ids and if we add or remove them we create these relations within the array to send to the backend. Once we get a response back without errors we need to always clear the array

You can see it is necessary in the [mutation](https://github.com/ourjapanlife/findadoc-web/blob/85aecf5db3f9e39dcbd66e483549af7cd91a3595/typedefs/gqlTypes.ts#L493) though and that is why we create it.

Maybe I do not understand exactly what your function is doing?

NabbeunNabi:
@ljbroersen have you updated this? If so don't hesitate to ping me for re review. If you haven't then kindly disregard until addressed

stores/facilitiesStore.ts
```diff
@@ -209,11 +211,20 @@ export const useFacilitiesStore = defineStore(
...
+
+                // compare the healthProfessionalRelations array from before and after updating the facility
+                const postUpdateRelations = facilitySectionFields.healthProfessionalsRelations
+                const hasUserChangedRelations
+                    = postUpdateRelations.length !== preUpdateRelations.length
+                      || !postUpdateRelations.every((relation, i) => relation === preUpdateRelations[i])
+
+                if (!hasUserChangedRelations) {
+                    facilitySectionFields.healthProfessionalsRelations = []
+                }
```
NabbeunNabi, line 227: This is where my confusion is for reference

NabbeunNabi (approved):
@ermish time

PR #1249, opened 2025-05-26
fix: doctor website link functionality

### LINKED ISSUE #1244: Fix website for doctor link to be clickable in a new tab
# Problem
Currently when looking at a healthcare professional's information there is no way to click on the website. It does not look like a hyperlink and is not working correctly. For UX we should make it easy for them to click on the website and have it open a new tab. If you inspect the element it shows the tab but you cannot in fact click the link

# Success
The healthcare professional's website if present is a hyperlink that opens a new tab with the styling of a hyperlink for the user.
### Requirements

- [x] Fix hyperlink for healthcare professional's website when available
- [x] hyperlink opens the website when present in a new tab
- [x] The link is stylized so people can understand that it is a clickable link

# Implementation notes

Image

As you can see it shows but you cannot in fact click the hyperlink to go to the website unless you are in the developers console

ConwayDecker:
✅ Resolves #1244 

## 🔧 What changed
- A `preventDefault` in closeOnOutsideClick.ts ha been removed; this is what was preventing the doctor link to work correctly.
- Attributes which allow the link to open in a new tab, provide a bit of security, and styling so it looks like a clickable link.
## 🧪 Testing instructions
Click on the links in the modal.
## 📸 Screenshots

-   ### Before
image
-   ### After
image

### DIFF
--- a/utils/closeOnOutsideClick.ts
+++ b/utils/closeOnOutsideClick.ts
@@ -9,7 +9,6 @@ export const vCloseOnOutsideClick = {
         let isOpen = false
 
         el.clickOutsideEvent = event => {
-            event.preventDefault()
             const eventTarget: EventTarget | null = event.target
             const isClickInsideElement = (el == eventTarget || el.contains(eventTarget as Node))
             if (isOpen && !isClickInsideElement) {
--- a/components/SearchResultDetails.vue
+++ b/components/SearchResultDetails.vue
@@ -72,6 +72,9 @@
                     <a
                         v-if="website"
                         :href="website"
+                        target="_blank"
+                        rel="noopener noreferrer"
+                        class="underline text-blue"
                     >{{ website }}</a>
                 </div>
                 <div class="phone flex my-4">


utils/closeOnOutsideClick.ts
```diff
@@ -9,7 +9,6 @@ export const vCloseOnOutsideClick = {
         let isOpen = false
 
         el.clickOutsideEvent = event => {
-            event.preventDefault()
```
NabbeunNabi, line 12: This makes sense as to why it would have not been working. Did you make sure this does not affect anything else going on with the modal or hamburger menu? @ConwayDecker

ConwayDecker: I checked out the modal and hamburger menu on both main and this branch; it looks like everything works! Moreover, I found a bug in the hamburger menu "Contact" link since that was being blocked on main.
Now everything seems to work as intended on this branch!

NabbeunNabi (approved):
Looks great then!

@ermish

NabbeunNabi:
@ConwayDecker I would do some research on UI/UX. I think for now it is probably to go back to the blue underline as that is what users are used to and expect for hyperlinks

theyokohamalife (approved):
LGTM!

NabbeunNabi (changes requested):
Sorry. Maybe I wasn't clear,
It should always be underlined. Whether hovered or not

ermish (approved):
looks good! 
After merging, let's double check the modal clickoutside still works 👍

PR #1257, opened 2025-06-02
feat: introduce snapshot testing

### LINKED ISSUE #1166: Introduce snapshot testing on the frontend
# Problem
We have a lot of tests that are just testing for the existence of UI elements. The UI can iterate constantly and we will have to delete and update tests over and over again. It is more common now to use snapshot testing to log the diffs within the UI. This will allow code reviewers to see the diffs and then decide whether they are intentional with the PR description.

We should introduce this slowly and focus on one component so let's start on the `about.vue` page. 
And have a github action that will leave a comment for the diff.

# Success
Snapshot testing is added to the project with `about.vue` page. This has a github action that comments diffs and one member is removed so that way it can be seen that there is a diff

### Requirements

- [ ] Snapshot testing is added successfully
- [ ] The about us page has a member removed and the diffs are shown within the testing
- [ ] A github action logs the changes in a comment for code reviewers

# Implementation notes

https://jestjs.io/docs/snapshot-testing

Ghezz9:
✅ Resolves #1166 

## 🔧 What changed

Lisa and me wrote the first snapshot test, which will create a stamp (photo)of the page `about.vue`. After this we wrote a Github Actions called `snapshot-diff.yml`. If a page changes, and you forget to update the snapshot, the test in GitHub Actions will fail.

❗`package.json` **has also been updated!**

### Creating a snapshot test
In the `test/jest/snapshot` folder, please create a file with the format `[page/component you want a snapshot from].jest.test.ts`. Inside you will need the usual `describe`, `it` blocks, and a `wrapper` variable that includes a `mount` function. Inside the `mount` function, you will mention which page you want to use (the page has to be an `import` at the top of the file) and if there is anything else that Jest will need to correctly make a snapshot. In the case of Find a Doc, that would be the `i18n` plugin.
At the end, after the `mount` function, you need to write an `expect`, which should match the snapshot. (`expect(wrapper.element).toMatchSnapshot()`)

Snapshots are stored in the `test/jest/snapshot/__snapshots__` folder.

### Running the snapshot testing suite
Running the testing suite is quite easy, just run `yarn test:snapshot` in the terminal.

If you want to update the snapshots, run `yarn test:snapshot:update`.

## 🧪 Testing instructions & 📸 Screenshots
A snapshot is made of `about.vue` with all the members visible.
image
Changing the amount of members visible with the help of `data.members.slice(0, 3)`, and then running `yarn test:snapshot` again, gives you this result:
image
"Snapshot -380" means that compared to the snapshot previously made, the current version of `about.vue` has -380 lines deleted, and no new ones got added.
If you run `yarn test:snapshot:update`
image
And then, if you run `yarn test:snapshot` again, it passes, since it now matches the updated version.
image


ljbroersen:
> ✅ **Snapshot Test Passed** — Nothing in the UI changed.

This one should not be passing, since the snapshot doesn't match the updated `about.vue`

NabbeunNabi (approved):
awesome work @ljbroersen and @Ghezz9 . Onto @ermish

package.json
```diff
@@ -78,7 +88,9 @@
         "postcss-html": "^1.7.0",
         "start-server-and-test": "^2.0.8",
         "tailwindcss": "^3.4.7",
-        "typescript": "5.5.4",
+        "ts-jest": "^29.3.4",
+        "ts-node": "^10.9.2",
```
ermish, line 92: why did this one get added? I think we can remove `ts-node`

test/jest/support/index.d.ts
```diff
@@ -0,0 +1 @@
+/// <reference types="jest" />
```
ermish, line 1: I don't think we need this file 🤔

test/jest/tsconfig.jest.json
```diff
@@ -0,0 +1,8 @@
+{
+  "compilerOptions": {
+    "composite": true,
+     "noEmit": false,
+    "types": ["jest", "node"]
```
ermish, line 5: we might not need node types here for a web project

ermish (changes requested):
looks awesome! just a few things to check

package.json
ljbroersen: I checked and we actually need it. If I remove the package, this is the error that will appear when I run `yarn test:snapshot`:
``` Error: Jest: Failed to parse the TypeScript config file
Error: Jest: 'ts-node' is required for the TypeScript configuration files. Make sure it is installed
Error: Cannot find package 'ts-node' imported from
```

ermish (approved):
nice work yall!

PR #1286, opened 2025-06-14
refactor: move mod components into separate folder

### LINKED ISSUE #1284: Change file structure for moderation components to have their own file
# Problem
We have enough Moderation components where it makes sense that they would be given their own file for cleanliness.

# Success
Change where the components are for moderation in their own folder and make sure the auto import is working
### Requirements

- [ ] Update Nuxt to auto import components based only on component name.
- [ ] Give the moderation components their own file
- [ ] Make sure ModSearchBar is also autoimported as before it was not

# Implementation notes
https://nuxt.com/docs/guide/directory-structure/components#component-names

NabbeunNabi:
✅ Resolves #1284 

## 🔧 What changed
We were getting a lot of components in one file so it seemed like it was the right time to move all the mod components for the mod panel into their own folder. It was outnumbering the original components and is good for developers to separate where they are working. Also the `nuxt.config` was updated based on the docs to not use the path as part of the component since we prefer to just name our components well. You can also see that the `ModSearchBar` import bug was fixed with this change

## 🧪 Testing instructions
No testing changes as CI/CD uses these components and the tests will pass.

To manually test it you can checkout to the branch and run everything to make sure that it is still running properly in all the pages


ermish (approved):
folders are good. sometimes

PR #1287, opened 2025-06-15
fix:  prevent the display of "undefined undefined" as the practitioner name

### LINKED ISSUE #1262: Practitioner modal showing undefined undefined for romaji name
# Problem
For some practitioners when their modal is displayed over the map for their romanized name is displaying undefined undefined. 

Image

# Success
Modals should display all practitioner's names in romaji.
### Requirements

- [ ] utility function for checking presence of English romaji name created
- [ ] modals no longer display undefined undefined

# Implementation notes
After talking to will about this we think it is because the mock data is missing romaji for some practitioners. We should create a utility function which checks to make sure the practitioner name array contains an English romaji name

KSmiyazawa:
✅ Resolves #1262 

## 🔧 What changed

- Refactored and extracted name formatting logic:
   - Moved the logic for building full practitioner names into a new utility function: `formatHealthcareProfessionalName` in `utils/nameUtils.ts`.
   - The function checks if both `firstName` and `lastName` exist before constructing the full name; otherwise, it returns null.
   - Uses logical fallbacks (`||`) to choose between English and Japanese names, and defaults to an empty string if both are missing.
- Updated `SearchResultDetails.vue`:
   - Replaced inline name formatting logic with the new utility function.

## 🧪 Testing instructions

1. Under **Doctors Nearby**, click on a practitioner to open the modal.
2. Verify that the practitioner's name is displayed correctly:
     - If both `firstName` and `lastName` exist for the selected locale, show the full name.
     - If the name is missing in the selected locale, fallback to the other locale’s name.
     - If both are missing, show an empty string (not "undefined undefined").
3. Switch locales:
     - Ensure names update accordingly based on available data.
     - Confirm fallback behavior works as expected when a name is missing in the selected locale.

## 📸 Screenshots

-   ### Before
image

-   ### After
image

### DIFF
--- a/components/SearchResultDetails.vue
+++ b/components/SearchResultDetails.vue
@@ -124,6 +124,7 @@ import { useLocaleStore } from '~/stores/localeStore.js'
 import { useSpecialtiesStore } from '~/stores/specialtiesStore.js'
 import { useSearchResultsStore } from '~/stores/searchResultsStore'
 import { Locale } from '~/typedefs/gqlTypes.js'
+import { formatHealthcareProfessionalName } from '~/utils/nameUtils'
 
 const { t } = useI18n()
 
@@ -132,31 +133,13 @@ const localeStore = useLocaleStore()
 const specialtiesStore = useSpecialtiesStore()
 
 const healthcareProfessionalName = computed(() => {
-    const englishName
-        = resultsStore.$state.activeResult?.professional.names.find(
-            n => n.locale === Locale.EnUs
-        )
-    const japaneseName
-        = resultsStore.$state.activeResult?.professional.names.find(
-            n => n.locale === Locale.JaJp
-        )
-
-    const englishFullName = englishName?.firstName && englishName?.lastName 
-        ? `${englishName.firstName} ${englishName.lastName}`
-        : null
-    const japaneseFullName = japaneseName?.firstName && japaneseName?.lastName
-        ? `${japaneseName.lastName} ${japaneseName.firstName}`
-        : null
-        
-    switch (localeStore.locale.code) {
-        case Locale.EnUs:
-            return englishFullName || japaneseFullName || ''
-        case Locale.JaJp:
-            return japaneseFullName || englishFullName || ''
-        default:
-            return englishFullName || japaneseFullName || ''
-    }
-})
+    const name = formatHealthcareProfessionalName(
+        resultsStore.$state.activeResult?.professional.names,
+        localeStore.locale.code as Locale
+    )
+    return name;
+}
+)
 const healthcareProfessionalDegrees = computed(() => {
     const healthcareProfessionalDegreesText
         = resultsStore.$state.activeResult?.professional.degrees.join(', ')
--- a/components/SearchResultDetails.vue
+++ b/components/SearchResultDetails.vue
@@ -141,16 +141,20 @@ const healthcareProfessionalName = computed(() => {
             n => n.locale === Locale.JaJp
         )
 
-    const englishFullName = `${englishName?.firstName} ${englishName?.lastName}`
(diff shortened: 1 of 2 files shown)


utils/nameUtils.ts
```diff
@@ -0,0 +1,27 @@
...
+        ? `${japaneseName.lastName} ${japaneseName.firstName}`
+        : null
+
+    switch (preferredLocale) {
+        case Locale.EnUs:
+            return englishFullName || japaneseFullName || ''
+        case Locale.JaJp:
+            return japaneseFullName || englishFullName || ''
+        default:
+            return englishFullName || japaneseFullName || ''
```
NabbeunNabi, line 25: I wonder if there is something better to display to let the user know to contact us that there is an issue 🤔

NabbeunNabi (changes requested):
This is a great start. but how can we also add some validation within moderation to ensure that the moderators include at least a romaji aka `EnUs` name when accepting a submission or updating locale names since it is our default

utils/nameUtils.ts
KSmiyazawa: Added 

> Something went wrong on our end, please let us know so we can resolve the issue as soon as possible

as suggested

KSmiyazawa:
@NabbeunNabi I added validation before submission approval that displays a toast error message if the Romaji name is missing.
image

components/SearchResultDetails.vue
```diff
@@ -138,27 +139,13 @@ const localeStore = useLocaleStore()
...
-    switch (localeStore.locale.code) {
-        case Locale.EnUs:
-            return englishFullName ? englishFullName : japaneseFullName
-        case Locale.JaJp:
-            return japaneseFullName ? japaneseFullName : englishFullName
-        default:
-            return englishFullName ? englishFullName : japaneseFullName
-    }
+    const name = formatHealthcareProfessionalName(
+        resultsStore.$state.activeResult?.professional.names,
```
NabbeunNabi, line 143: I know this is how it was before, but since we have updated the way we use imports to use the store function to access the state like above could you try that?

components/moderation-panel/ModEditSubmissionForm.vue
```diff
@@ -316,6 +317,20 @@ const validateHealthcareProfessionalFields = () => {
     return areAllFieldsValid
 }
 
+const validateHealthcareProfessionalRomajiName = () => {
```
NabbeunNabi, line 320: 💯 nice

NabbeunNabi (changes requested):
Almost there @KSmiyazawa

components/SearchResultDetails.vue
```diff
@@ -138,27 +139,13 @@ const localeStore = useLocaleStore()
 const specialtiesStore = useSpecialtiesStore()
```
NabbeunNabi, line 139: Like here @KSmiyazawa

NabbeunNabi (approved):
Looks good now!

Onto @ermish

components/moderation-panel/ModEditSubmissionForm.vue
```diff
@@ -316,6 +317,20 @@ const validateHealthcareProfessionalFields = () => {
...
+    const healthcareProfessionalFields = healthcareProfessionalsStore.healthcareProfessionalSectionFields
+    const names = healthcareProfessionalFields.names
+
+    // Check if the healthcare professional has a name in Romaji (Latin characters)
+    const hasANameInRomaji = names.some(healthcareProfessionalName => {
```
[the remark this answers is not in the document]
KSmiyazawa: It's just checking whether a Romaji name already exists among the existing ones. But yes, I could also just check if there's a name in the `en_US` locale.

PR #1299, opened 2025-06-17
feat: updated specialties available and added top 10

ermish:
## 🔧 What changed
3 highlights
- Our specialty display names were not very user friendly, so I've updated the text to match the wording most people would search for, such as "Dentist" instead of "Dentistry" 💯 
- Updated our specialties to be translatable! 🥳 Also added all language translations 
- Added top 10 specialties to be used by our new on-boarding flow ✨ 

## 🧪 Testing instructions
Test the specialties dropdown works with updated values


NabbeunNabi (commented):
@ermish just one question and something to think about.

Though I know that they won't display the filters I'm pretty sure are listed by specialties

PR #1307, opened 2025-06-20
fix(style): change styling for modal on main page

NabbeunNabi:
Remove pink background on main page and use our own utility for screen orientation

Mod panel
image

Main page. Path will need to be updated in @ermish ui redesign if he changes the page to `/search`
image


theyokohamalife (approved):
Confirmed in the Netlify preview that the new (but temporary) styling works. Much appreciated! 
Screenshot 2025-06-20 at 2 36 34 PM

PR #1308, opened 2025-06-20
chore: add internal medicine to localization

theyokohamalife:
✅ Resolves #[Issue number]

## 📸 Screenshots

-   ### Before

-   ### After


ermish:
> > @theyokohamalife I noticed the localizations aren't translated into the languages. (good case for having AI help translate with windsurf or cursor) Do you want to do this before merging?
> 
> @ermish This is something I was discussing with @NabbeunNabi that we'd like to automate, because the majority of the fields are missing localization. Will merge this as is for now so the Japanese doctors will see it this weekend, however we should look into free AI translations via API for NPOs.

love it! definitely a great AI automation use case 🥳

PR #1315, opened 2025-06-20
refactor: update screen orientation to be reactive

NabbeunNabi:
## 🔧 What changed
Before this was not reactive so you had to refresh the page. Now it changes as it goes

### DIFF
--- a/utils/useScreenOrientation.ts
+++ b/utils/useScreenOrientation.ts
@@ -1,17 +1,20 @@
-import { ref, onMounted } from 'vue'
+import { ref, computed, onMounted, onUnmounted } from 'vue'
 
 export const useScreenOrientation = () => {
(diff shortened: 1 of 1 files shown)


utils/useScreenOrientation.ts
```diff
@@ -1,17 +1,20 @@
...
+        window.addEventListener('resize', updateWidth)
+    })
+    onUnmounted(() => {
+        window.removeEventListener('resize', updateWidth)
     })
 
-    return {
-        isPortrait
-    }
+    const isPortrait = computed(() => width.value <= 768)
```
ermish, line 17: this isn't really a true portrait vs landscape. It's more like desktop vs tablet sizing. Can we change this to be width < height? Can do in another PR

PR #1324, opened 2025-06-22
fix: prevent `UnsavedChanges` modal from appearing when deleting a facility

### LINKED ISSUE #1274: Add `DeleteConfirmation` modal instead of `UnsavedChanges` modal when deleting a facility
# Problem
Right now, if you delete a facility, instead of showing the `DeleteConfirmation` modal, the `UnsavedChanges` modal appears. Which does not make sense since once you delete a facility, you cannot go back and change anything about the entry anymore, since it is already deleted from the database (and you will get a HTTP 500 error).

# Success
When successful, instead of the `UnsavedChanges` modal, the `DeleteConfirmation` modal should appear.
### Requirements

- [ ] When clicking the delete button, the `DeleteConfirmation` modal should appear.
- [ ] Then, when clicking the 'Delete Forever' button, you will be redirected to the facilities overview page.
- [ ] The `UnsavedChanges` modal should not appear anymore.

ljbroersen:
✅ Resolves #1274 

## 🔧 What changed
Before, if you would delete a facility, the `UnsavedChanges` modal would appear after the `DeleteConfirmation` modal. Which makes no sense, because the facility is already deleted.
Now, 
1. `deleteFacilityOrHealthcareProfessional()` runs, and you confirm your choice on the `DeleteConfirmation` modal
2. The ref `originalFacilityRefsValue` turns to `undefined`-- all memory of the now deleted facility gets wiped
3. `facilityHasUnsavedChanges()` runs, but there is nothing to compare with anymore, resulting in the `UnsavedChanges` modal being skipped 
4. You return back to the facility overview page

### DIFF
--- a/components/moderation-panel/ModEditFacilityOrProfessionalTopbar.vue
+++ b/components/moderation-panel/ModEditFacilityOrProfessionalTopbar.vue
@@ -340,6 +340,8 @@ const deleteFacilityOrHealthcareProfessional = async () => {
             handleServerErrorMessaging(response.errors, toast, t)
             return response
         }
+        // Wipe the data of the now deleted facility from the ref to prevent the UnsavedChanges modal from triggering
+        originalFacilityRefsValue.value = undefined
 
         toast.success(t('modEditFacilityOrHPTopbar.facilityDeletedSuccessfully'))
         // Redirect to the dashboard since the facility no longer exists


NabbeunNabi (approved):
@ljbroersen looks good. small and might. While you are at it. Can you see if this happens for healthcare professionals too since it seems like it could do the same thing since the logic is similar. If so do you mind making a PR for that. Do not need an issue since it is so small unless you do not have time and want someone else to work on it. But if you do and it needs to be fixed just write the context in the PR description

ermish (approved):
nice improvement!

PR #1330, opened 2025-06-23
fix: fix formHasUnsavedChange

### LINKED ISSUE #1320: Prevent `formHasUnsavedChanges()` to trigger when leaving `ModEditSubmissionForm.vue` when nothing changed
# Problem
Right now, if you open a submission, decide you don't want to edit anything after all and click the back button of your browser, the `formHasUnsavedChanges()` function in `ModEditSubmissionForm.vue` gets triggered. A modal will appear that is meant to confirm that you have unsaved changes and if you are sure you want to leave the page. But there are no unsaved changes, so the modal should not appear in the first place.

This issue also makes the test `should not display modal if user navigates back without making changes` in `moderationEditSubmissionModal.cy.ts` flaky.

Image

# Success
If you click the back button in your browser, without editing anything on the submission page, the modal should not appear and you instantly get redirected back to the submission overview.

### Requirements

- [ ] `formHasUnsavedChanges()` should not be triggered when no form on the page has any changes
- [ ] The test `should not display modal if user navigates back without making changes` should pass without any issues 

# Implementation notes
If you do the same with a Healthcare Professional or a Facility (open a HP/Facility to edit, decide against it and click the back button), the modal does not appear and it works smoothly.

Ghezz9:
✅ Resolves [#1320](https://github.com/ourjapanlife/findadoc-web/issues/1320)

## 🔧 What changed

Initially, the primary problem we observed was an immediate "form has unsaved changes" prompt upon loading a submission, despite no user interaction. This behavior suggested that either the form's initial snapshot was not being taken correctly, or it was being created at the wrong time, leading to a perceived discrepancy with the actual form state. Through debugging, it became clear the issue stemmed from subtle type and data value inconsistencies between how the form fields were initialized and how the baseline snapshot for comparison was constructed.

Now we standardized the default/fallback values across both the form field initialization and the snapshot reconstruction.

image

Here is how i create the snapshot, my question is when a submission is created/opened the only fields that are not empty is usually this 3, am i saying something wrong??

Esentially this issues abou "form has unsaved changes" not work correctly only when you enter on a submission and you don't click update/save/save and exit.
I've tested more things like enter write a name save go out and enter again for seeing if it's still displayed and yes it works. Nothing changed how to works because i've only focused inside `initializeSubmissionFormValues`.

## 📸 Screenshots

-   ### Before

-   ### After

https://drive.google.com/file/d/1dxkF8yc0DIYGzxB1OXz-cHA91oQ0rU2V/view?usp=sharing


components/moderation-panel/ModEditSubmissionForm.vue
```diff
@@ -406,6 +407,8 @@ function initializeSubmissionFormValues(submissionData: Submission | undefined)
     = submissionData?.healthcareProfessionals?.[0]?.specialties ?? []
     healthcareProfessionalSections.spokenLanguages
     = submissionData?.spokenLanguages ?? []
+
+    console.log(healthcareProfessionalSections)
```
NabbeunNabi, line 411: 🤯

components/moderation-panel/ModEditSubmissionForm.vue
```diff
@@ -367,6 +367,7 @@ function initializeSubmissionFormValues(submissionData: Submission | undefined)
                 ))
             .filter((hp): hp is NonNullable<typeof hp> => hp !== undefined)
 
+    console.log(facilitySectionFields)
```
NabbeunNabi, line 370: 🤯

components/moderation-panel/ModEditSubmissionForm.vue
```diff
@@ -424,19 +427,34 @@ const formHasUnsavedChanges = () => {
...
-        || submissionBeforeChangesComparison?.facility?.nameJa !== facilitySectionFields.nameJa
-        || submissionBeforeChangesComparison?.facility?.contact?.phone !== facilitySectionFields.phone
-        || (submissionBeforeChangesComparison?.facility?.contact?.email
-          && submissionBeforeChangesComparison?.facility?.contact?.email !== facilitySectionFields.email)
-        || (submissionBeforeChangesComparison?.facility?.contact?.website
-          && submissionBeforeChangesComparison?.facility?.contact?.website !== facilitySectionFields.website)
-        || submissionBeforeChangesComparison?.facility?.contact?.address?.postalCode !== facilitySectionFields.postalCode
-        || !arraysAreEqual(submissionBeforeChangesComparison?.facility?.healthcareProfessionalIds,
-                           facilitySectionFields.healthcareProfessionalIds)
+    const isFacilitySectionEmpty = (
```
NabbeunNabi, line 430: Is this only a check if it is empty? Maybe you and @KSmiyazawa can join forces on this and her PR that does something similar in terms of sending data back so that way the work is going together

NabbeunNabi (changes requested):

components/moderation-panel/ModEditSubmissionForm.vue
KSmiyazawa: @NabbeunNabi @Ghezz9 I believe this issue would be resolved once my PR gets merged.

NabbeunNabi: Put it in draft for now and see if you and @KSmiyazawa can work on her rebasing together so we can get that in soon either today or tomorrow if possible

components/moderation-panel/ModEditSubmissionForm.vue
```diff
@@ -456,6 +458,72 @@ function initializeSubmissionFormValues(submissionData: Submission | undefined)
...
+            snapshotBase = {
+                ...submissionData,
+                healthcareProfessionalName: submissionData.healthcareProfessionalName ?? '',
+                googleMapsUrl: submissionData.googleMapsUrl ?? '',
+                spokenLanguages: submissionData.spokenLanguages ? [...submissionData.spokenLanguages] : []
+            }
+        } else {
+            return
+        }
+
```
NabbeunNabi, line 475: Flip logic with if not submission data return and then you can make the snapshot base a const after the conditional

components/moderation-panel/ModEditSubmissionForm.vue
```diff
@@ -456,6 +458,72 @@ function initializeSubmissionFormValues(submissionData: Submission | undefined)
     = submissionData?.healthcareProfessionals?.[0]?.specialties ?? []
     healthcareProfessionalSections.spokenLanguages
     = submissionData?.spokenLanguages ?? []
+
+    nextTick(() => {
+        let snapshotBase: Submission
```
NabbeunNabi, line 463: snapshot is the name of testing. Snapshot base of what?

Better variable name not that I can think what it is right now

components/moderation-panel/ModEditSubmissionForm.vue
```diff
@@ -456,6 +458,72 @@ function initializeSubmissionFormValues(submissionData: Submission | undefined)
...
+            acceptedInsurance: [...healthcareProfessionalSections.acceptedInsurance],
+            additionalInfoForPatients: healthcareProfessionalSections.additionalInfoForPatients || undefined,
+            degrees: [...healthcareProfessionalSections.degrees],
+            specialties: [...healthcareProfessionalSections.specialties],
+            spokenLanguages: [...healthcareProfessionalSections.spokenLanguages],
+            facilityIds: [...healthcareProfessionalSections.facilityIds],
+            __typename: 'HealthcareProfessionalSubmission' as const
+        }]
+
+        snapshotBase.facility = reconstructedFacility
```
NabbeunNabi, line 521: Reconstructed means what in this context?

If a comment makes more sense cuz it actually reconstructs you can do that

NabbeunNabi (changes requested):
@Ghezz9 great start!

Just a few comments and you are almost there. I also watched your video but you didn't show that it still has the behavior when you do make a change the modal pops up 🍝

components/moderation-panel/ModEditSubmissionForm.vue
Ghezz9: Solved

components/moderation-panel/ModEditSubmissionForm.vue
```diff
@@ -456,6 +458,74 @@ function initializeSubmissionFormValues(submissionData: Submission | undefined)
...
     = submissionData?.spokenLanguages ?? []
+
+    nextTick(() => {
+        if (!submissionData) {
+            return
+        }
+
+        // This constant holds the initial state of the submission form values.
+        // It serves as the baseline to detect if any unsaved changes have occurred
+        const snapshotBase: Submission = {
```
NabbeunNabi, line 469: Based on your comment wouldn't it be better to call this `initialSubmissionFormValueS apshot` or something along those lines.

I like comments as I originally suggested adding them but variables need to be updated and changed as code gets refactored. Comments may not be updated or removed or changed. So we want to make sure the variable name is the first comment and the comment is the fallback of understanding

components/moderation-panel/ModEditSubmissionForm.vue
```diff
@@ -456,6 +458,74 @@ function initializeSubmissionFormValues(submissionData: Submission | undefined)
...
+            ...submissionData,
+            healthcareProfessionalName: submissionData.healthcareProfessionalName ?? '',
+            googleMapsUrl: submissionData.googleMapsUrl ?? '',
+            spokenLanguages: submissionData.spokenLanguages ? [...submissionData.spokenLanguages] : []
+        }
+
+        // Reconstruct the FacilitySubmission object from the current form field values
+        // (facilitySectionFields) instead of using submissionData.facility directly.
+        // This ensures that we capture the latest edited state for comparison or submission.
```
NabbeunNabi, line 478: ♻️

components/moderation-panel/ModEditSubmissionForm.vue
```diff
@@ -456,6 +458,74 @@ function initializeSubmissionFormValues(submissionData: Submission | undefined)
...
+        const reconstructedFacility: FacilitySubmission = {
```
NabbeunNabi, line 479:
```suggestion
        const reconstructedFacilitySubmission: FacilitySubmission = {
```

components/moderation-panel/ModEditSubmissionForm.vue
```diff
@@ -456,6 +458,74 @@ function initializeSubmissionFormValues(submissionData: Submission | undefined)
...
+                },
+                __typename: 'Contact' as const
+            },
+            healthcareProfessionalIds: [...facilitiesStore.facilitySectionFields.healthcareProfessionalIds],
+            __typename: 'FacilitySubmission' as const
+        }
+
+        // Reconstruct the HealthcareProfessionalSubmission object from the current form field values
+        // (healthcareProfessionalSections) instead of using submissionData.healthcareProfessional directly.
+        // This ensures that we capture the latest edited state for comparison or submission.
```
NabbeunNabi, line 510: ♻️

components/moderation-panel/ModEditSubmissionForm.vue
```diff
@@ -456,6 +458,74 @@ function initializeSubmissionFormValues(submissionData: Submission | undefined)
...
+        const reconstructedHealthcareProfessionals: HealthcareProfessionalSubmission[] = [{
```
NabbeunNabi, line 511:
```suggestion
        const reconstructedHealthcareProfessionalSubmission: HealthcareProfessionalSubmission[] = [{
```
I removed the s as it mat ch ed the type and follows the variable understanding first concept

NabbeunNabi (changes requested):

Ghezz9:
[fix: normalized Latitude/longitude because they were flaky]

This commit i did because esentially the test doesn't pass cause another parameters (maybe) it's used on fake it was different, it seems both are number but this problem about unsaved changes happened because the first check confronted `0 !== ''` or `'' !== null` ecc.. so I thought it was better to normalize Logitudie/Latidude.

It works

NabbeunNabi (changes requested):
Almost there @Ghezz9

components/moderation-panel/ModEditSubmissionForm.vue
```diff
@@ -469,16 +541,62 @@ const hasFacilityChanges = (submissionBeforeChangesComparison: Submission | unde
...
+        // If the value is a string and is empty or consists only of whitespace, treat it as 0.
+        if (typeof value === 'string' && value.trim() === '') {
+            return 0
+        }
+
+        // Attempt to parse the value as a float.
+        // If parsing results in NaN (e.g., for non-numeric strings or undefined/null after type assertion),
+        // default to 0 to ensure a numeric comparison.
+        return parseFloat(value as string) || 0
+    }
```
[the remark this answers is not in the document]
Ghezz9: I've done a screen recorder where i used `.toString()` 

https://drive.google.com/file/d/155ThpXh6AY5jafcaYQJupdzo9Shm5LsF/view?usp=sharing

Say to me if i did mistake on the comparison, but i think is more like how we receive that value from backend or change how we initialize the form

image

NabbeunNabi (approved):
Go @ermish go

components/moderation-panel/ModEditSubmissionForm.vue
```diff
@@ -456,30 +456,133 @@ function initializeSubmissionFormValues(submissionData: Submission | undefined)
...
+
+        /* This constant holds the initial state of the submission form values.
+        *  It serves as the baseline to detect if any unsaved changes have occurred
+        */
+        const initialSubmissionFormValueSnapshot: Submission = {
+            ...submissionData,
+            healthcareProfessionalName: submissionData.healthcareProfessionalName ?? '',
+            googleMapsUrl: submissionData.googleMapsUrl ?? '',
+            spokenLanguages: submissionData.spokenLanguages ? [...submissionData.spokenLanguages] : []
+        }
```
ermish, line 473: you can use `as Submission` here to cast the type as `Submission`

components/moderation-panel/ModEditSubmissionForm.vue
```diff
@@ -456,30 +456,133 @@ function initializeSubmissionFormValues(submissionData: Submission | undefined)
...
+
+        /**
+         * Reconstruct the FacilitySubmission object from the current form field values
+         * (facilitySectionFields) instead of using submissionData.facility directly.
+         * This ensures that we capture the latest edited state for comparison or submission.
+         */
+        const reconstructedFacilitySubmission: FacilitySubmission = {
+            id: submissionData?.facility?.id || '',
```
ermish, line 481: lazy comment here, but I think we can utilize a clone method here.

components/moderation-panel/ModEditSubmissionForm.vue
```diff
@@ -456,30 +456,133 @@ function initializeSubmissionFormValues(submissionData: Submission | undefined)
...
+            healthcareProfessionalIds: [...facilitiesStore.facilitySectionFields.healthcareProfessionalIds],
+            __typename: 'FacilitySubmission' as const
+        }
+
+        /**
+         * Reconstruct the HealthcareProfessionalSubmission object from the current form field values
+         * (healthcareProfessionalSections) instead of using submissionData.healthcareProfessional directly.
+         * This ensures that we capture the latest edited state for comparison or submission.
+         */
+        const reconstructedHealthcareProfessionalSubmission: HealthcareProfessionalSubmission[] = [{
```
ermish, line 514: lazy comment here, but I think we can utilize a clone method here.

components/moderation-panel/ModEditSubmissionForm.vue
```diff
@@ -513,11 +616,23 @@ const hasHealthcareProfessionalChanges = (submissionBeforeChangesComparison: Sub
...
+    const initialSubmission = submissionBeforeChanges.value
+    if (!initialSubmission) {
+        return false
+    }
+
+    const hasNotesChanges = (initialSubmission.notes ?? '') !== currentSubmissionNotes.value
+    const facilityChanges = hasFacilityChanges(initialSubmission)
+    const hpChanges = hasHealthcareProfessionalChanges(initialSubmission)
+
+    return facilityChanges || hpChanges || hasNotesChanges
```
ermish, line 634: tip: I recommend using return variables! So you make this a variable like `hasChanges`, then return hasChanges. 
These make it easier to understand the check PLUS you can now debug and check the variable easily

components/moderation-panel/ModEditSubmissionForm.vue
```diff
@@ -456,30 +456,133 @@ function initializeSubmissionFormValues(submissionData: Submission | undefined)
     = submissionData?.healthcareProfessionals?.[0]?.specialties ?? []
     healthcareProfessionalSections.spokenLanguages
     = submissionData?.spokenLanguages ?? []
+
+    nextTick(() => {
```
ermish, line 460: instead of nextTick, I believe we can use `onMounted()` here. This should also resolve your problem with the function below getting called first

ermish (changes requested):
hasChanges = true!

components/moderation-panel/ModEditSubmissionForm.vue
Ghezz9: `nextTick()` is crucial here because it ensures that all reactive updates resulting from `submissionData` being loaded into the Pinia stores `(facilitiesStore.facilitySectionFields, healthcareProfessionalSections)` have been fully processed and flushed by Vue's reactivity system. Only then are the values read from the stores to create `initialSubmissionFormValueSnapshot` guaranteed to be the final and stable state.

If `submissionData` is loaded asynchronously,`onMounted()` would execute before `submissionData` is available, or before the stores are completely populated.

ermish: is this a LLM response? I'm not sure it's correct actually. 
I think onmounted will act the same since this is not an async call and might even fire on a DOM update before the whole DOM is rendered for onmounted

components/moderation-panel/ModEditSubmissionForm.vue
```diff
@@ -513,11 +616,23 @@ const hasHealthcareProfessionalChanges = (submissionBeforeChangesComparison: Sub
...
+        || ((submissionBeforeChangesComparison?.healthcareProfessionals?.[0]?.additionalInfoForPatients ?? '')
+          !== (hpSectionFields.additionalInfoForPatients ?? ''))
     )
 }
 
-const formHasUnsavedChanges = () => hasFacilityChanges(submissionBeforeChanges.value)
-  || hasHealthcareProfessionalChanges(submissionBeforeChanges.value)
+const formHasUnsavedChanges = () => {
+    const initialSubmission = submissionBeforeChanges.value
+    if (!initialSubmission) {
```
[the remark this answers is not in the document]
ermish: bump

ermish:
@Ghezz9 has some changes still in case you were waiting! I also responded to the nextTick comment

components/moderation-panel/ModEditSubmissionForm.vue
Ghezz9: No LLM response, i will try to do your solution.
Thank you

NabbeunNabi:
@Ghezz9 is there any update on this?

PR #1345, opened 2025-06-27
feat: clean unused keys

Ghezz9:
✅ Resolves #[1281]

## 🔧 What changed
@ermish @NabbeunNabi 
It's still on working but i thought it was a good idea discuss about best way for delete without deleting also keys that we are using for some strange errors i've prefered to create a txt called `potentially_unused_keys.txt`.
This file has the role of showing for each .json which kind of keys we are not using inside `findadoc-web` , the two screenshot show that it seems to work, i was curious about which is the best idea for deleting and then check if it's every keys is existing in the code. I think i can easily wrote 2 more functions that delete each keys that not appear on `.vue .ts .js` ecc.. and then run `yarn lint:locale`.
After this i will create a GitHubActions that check if some keys are not using anymore.

The idea of screenshots it's good? i've search other i18n keys like `t('footer.clickHere')` and it appears in the search on the right component (in this case `Footer.vue`) so i assume that if i write `t(specialties.orthodontics)` and it doesn't have any match it's because it exist only in the `.18n/locales/*.json`.

## 📸 Screenshots

-   ### Before

-   ### After

image

image


NabbeunNabi (changes requested):
What is globby? @Ghezz9 
Also per your concern we should run this check after it checks all our files in vue and stores for unused keys.

Then let's make it have a cron job with an action that weekly makes a PR if there are any changes.

Similar to the codegen one on the front end. So we can have a reviewer check and make sure

i18n/cleanUnusedLocaleKeys.js
```diff
@@ -0,0 +1,79 @@
...
+const localesDir = path.join(__dirname, '/locales')
+const sourceCodeDir = path.join(__dirname, '..')
+
+const translationFilePaths = fs.readdirSync(localesDir)
+    .filter(file => file.endsWith('.json'))
+    .map(file => path.join(localesDir, file))
+
+// 🔍 Extract used translation keys from code
+const extractUsedTranslationKeys = async () => {
+    const files = await glob(['**/*.{js,ts,vue,tsx}'], {
```
NabbeunNabi, line 20: We don't use tsx or js for keys

i18n/cleanUnusedLocaleKeys.js
```diff
@@ -0,0 +1,79 @@
...
+            })
+            .filter(([, unusedKeys]) => unusedKeys.length > 0)
+    )
+
+    if (Object.keys(unusedKeysByLocale).length === 0) {
+        //console.log('✅ No unused translation keys found.')
+        return
+    }
+
+    //console.log('\n📋 Potentially unused translation keys:\n')
```
NabbeunNabi, line 64: Console. Log commented out

NabbeunNabi (changes requested):
Getting better.

Looking better.

Have you looked into the cron GitHub action?

NabbeunNabi (approved):
@Ghezz9 Great job! just remember to add that `potentially_unused_keys.txt` to the `.gitignore`

i18n/cleanUnusedLocaleKeys.js
```diff
@@ -0,0 +1,136 @@
...
+
+    const output = Object.entries(unusedKeysByLocale)
+        .map(([file, keys]) => {
+            console.info(`📂 ${file}:`)
+            keys.forEach(key => console.info(`  - ${key}`))
+            return `${file}:\n${keys.map(k => `- ${k}`).join('\n')}\n`
+        })
+        .join('\n')
+
+    fs.writeFileSync('potentially_unused_keys.txt', output, 'utf-8')
```
ermish, line 131: let's clarify these are i18n keys, since there are other keys in the project

```suggestion
    fs.writeFileSync('potentially_unused_i18n_keys.txt', output, 'utf-8')
```

PR #1346, opened 2025-06-29
feat: refreshed search UI 

ermish:
✅ Resolves #[Issue number]

# 🔧 What changed

We have a brand new, refreshed search ui experience! 
This complements our new onboarding flow ✨ 

### Topnav
- Reworked topnav to no longer include the search dropdowns
- made it float over the map

### Map
- Removed unnecessary controls that made it noisy
- Changed the map design
- Moved it to fill the whole screen from the top, giving our elements a floating look

### Search Results
- Now a floating bottom sheet
- Draggable

### Search Result Details
- Now a bottom sheet that replaces the search results space. 
- Draggable
- Nice animations! 
- Removed the "doctors nearby" text which should have changed anyways
- Removed the non-functional filters button 

### Search Filters
- Created a new bottom sheet for search filters! ✨ 
-  Moved them to the search results area

## 🧪 Testing instructions
- Test the site on desktop and mobile to ensure it works
- Try on different browsers and screen sizes if possible
- Test language and search filter changes

## 📸 Screenshots

-   ### Before

-   ### After
TBD


components/BottomSheet.vue
```diff
@@ -0,0 +1,308 @@
+<!-- Inspired by https://github.com/vaban-ru/vue-bottom-sheet?tab=readme-ov-file,
+ but rewritten due to it no longer being maintained -->
+<!-- It's been customized to allow for multiple positions, simplified the code, and added a few features -->
+<template>
+    <Teleport to="body">
+        <div
+            ref="bottomSheet"
+            class="transition-[visibility]"
+            :style="{
```
NabbeunNabi, line 9: Can't do this in tailwind?

ermish: unless you mean adding the classes instead of styles?

NabbeunNabi: Yep. You can have dynamic classes in tailwind. But again I know what has gone into this so I am just curious.

And then in this case.

What do you think to pull this out into a rare style tag to make sure people do not touch this?

PR #1392, opened 2025-07-22
feat: create linter error for hardcoded colors

### LINKED ISSUE #1361: feat: add hard coded color checker to linter
# Problem
When we choose colors for text, bg, etc, we should only be using our theme colors and not hard coded colors. This will especially be important once we finish our dark modes and other color modes

(for ex, `text-primary-muted` = 😊 , `text-gray-500` = 😭)

Currently, our linter checker does not catch this. We want to change that so that it doesn't allow hard coded colors

# Success

When running the lint check, it should not allow hard coded colors: only theme colors

### Requirements

- [ ] When running the lint check, it should not allow hard coded colors: only theme colors

# Implementation notes

There will be some cases where we will need to ignore this linter check (for instance, the underline color of the <a> tags)

Ghezz9:
✅ Resolves [#1361](https://github.com/ourjapanlife/findadoc-web/issues/1361)

## 🔧 What changed

Created an `eslint-rules` that enforce the use of colors from our theming system, What the rules do?

- Prohibition of "hardcoded" Tailwind colors: The rule prevents the direct use of generic Tailwind color utility classes (like text-blue, text-red etc.)

**What is changed inside files?**
`eslint.config.js`: Applied ESLint rules for `.vue` files. Use `vueParser` to properly parse the `<template> and <script>` sections of Vue components. It also enables our custom `no-forbidden-color` rule on these files, ensuring that hardcoded Tailwind colors are prohibited within Vue templates.
Inside this file, we use a new dependency, i know we probably might create our custom rule but the process seems so hard.
`vue-eslint-parser`, we need this one because without ESLint is designed for standar JS files but Vue has also `template` part, i've searched more about how the mechanism works ( https://github.com/vuejs/vue-eslint-parser ) and i think it's essential for our purpose.
`no-forbidden-color-js`: Check inside each `.vue` file if there are hardcoded colors.
`ErrorLayout.vue`: it seems there was an warning line code
`SearchResultDetails`: changed `text-blue` to `text-hyperlink`

## 🧪 Testing instructions

image

## 📸 Screenshots

-   ### Before

-   ### After

### DIFF
--- a/assets/css/tailwind.css
+++ b/assets/css/tailwind.css
@@ -48,7 +48,6 @@
     --color-error: /* #ef4444 */ 239 68 68;
     --color-warning: /* #f59e0b */ 245 158 11;
     --color-info: /* #3b82f6 */ 59 130 246;
-    --color-hyperlink: /* #1e293b */ 30 41 59 / 1;
   }
 
   .theme-coral {
@@ -71,7 +70,6 @@
     --color-error: /* #d1523f */ 209 82 63;
     --color-warning: /* #f59e0b */ 245 158 11;
     --color-info: /* #3b82f6 */ 59 130 246;
-    --color-hyperlink: /* #1e293b */ 30 41 59 / 1;
   }
 
   .theme-violet {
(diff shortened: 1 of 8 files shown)


NabbeunNabi:
@Ghezz9 is there any update on this?

PR #1400, opened 2025-07-23
refactor: remove is under review since updates in background

NabbeunNabi:
✅ Resolves #[Issue number]

## 🔧 What changed
This removes unnecessary update necessary validation and toast since we run it before approving to the backend already


NabbeunNabi:
> is Under review is supposed to be for the case when multiple moderators are working at the same time. Is this not going to be an issue then?

This title is bad this just removes an unnecessary toast since you approve which called update and then tries to approve it. So this needs to be updated first is not necessary since that function doesn't action

PR #1406, opened 2025-07-23
feat: add new list card for moderation panel based on figma

NabbeunNabi:

## 🔧 What changed
We are updating our moderation panel design as we move forward. This is the first step to updating the list container. So in this PR I
- Changed the name of the container since it holds more than submissions
- Added a card that for now is only styled for submissions. Can easily be adapted for HP and Facility but to keep this PR small for now we will start with submissions and add the other two which follow same logic to component in future PR
- Added translations for this

## 📸 Screenshots
Adding portrait and landscape as other parts need to be updated based on the design

[Figma Link](https://www.figma.com/design/QXQxlw4jNFIhMxkKU7bbwq/Find-a-doc--Japan?node-id=2705-6769&t=yudM3fz2wdInk8mu-4)

-   ### Before
Portrait 
image

Desktop
image

-   ### After

https://github.com/user-attachments/assets/a580c808-b91a-4190-9d2a-33c28eba5d56

### DIFF
--- a/test/cypress/e2e/moderationEditSubmissionModal.cy.ts
+++ b/test/cypress/e2e/moderationEditSubmissionModal.cy.ts
@@ -33,7 +33,7 @@ describe('Moderation Edit Submission Modal', () => {
             cy.location('pathname').should('eq', '/moderation')
         })
 
-        it('should not display modal if user navigates back without making changes', () => {
+        it.skip('should not display modal if user navigates back without making changes', () => {
             cy.get('[data-testid="mod-submission-list-item-1"]').click()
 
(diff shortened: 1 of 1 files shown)


ermish:
before I check out the code, I have a few UX suggestions:
- make the row number faded and smaller
- move the tag (we might have more too) to be in the same line as the row number. It feels a bit too cluttered with too many rows
- make the tags just a bit smaller, and maybe a 10/20% opacity fade
- can you move the review button to the top right rather than middle right?
- change wording `created on` and `updated last`
- move the date part under the name so the name is the first thing you read.

ermish (approved):
woo! design!

PR #1410, opened 2025-07-27
fix: fix locale name input actions to be standard and consistent

NabbeunNabi:

## 🔧 What changed
Before with the previous fix, a few place still did not work as desired. Now they are working as desired in both pages and with consistent actions for user experience


ermish (approved):
looks good, only thing I can think of is if you should use an enum for editing vs creating mode.

NabbeunNabi:
> looks good, only thing I can think of is if you should use an enum for editing vs creating mode. 

I'll make an issue for this later. Good refactor.

Silly to do the flippy flop

PR #1447, opened 2025-08-11
fix: locale of data should match active locale of app

### LINKED ISSUE #1309: Display data in selected localization
# Problem
When a locale other than English is selected, the healthcare professional information in the modal and sidebar are still displayed in English.

# Success
Healthcare professional information is displayed in the selected locale for fields that are localized.

### Requirements
When Japanese is selected from the language dropdown:

- [ ] Healthcare professional's name is localized
- [ ] Facility name is localized
- [ ] Specialty is localized
- [ ] Address is localized

# Implementation notes

Image

ljbroersen:
✅ Resolves #1309

## 🔧 What changed
I created a helper function that checks what locale the app is currently set to, and then checks if the healthcare professional has a name assigned to that locale. If not, use the first name that is included in the array of names.

## 🧪 Testing instructions
Create a healthcare professional from the moderation panel, and then switch locales on the homepage.

## 📸 Screenshots

-   ### After
From left to right, locale switched to English, Japanese and German.
image
image
image

### DIFF
--- a/components/SearchResultsList.vue
+++ b/components/SearchResultsList.vue
@@ -113,7 +113,6 @@ function getLocalizedName(
     const selectedName
     = names.find(name => name.locale === currentLocale)
       || names.find(name => name.locale.startsWith(localePrefix))
-      || names.find(name => name.locale === 'en')
       || names[0]
 
     return `${selectedName.firstName} ${selectedName.lastName}`


components/SearchResultsList.vue
```diff
@@ -104,4 +103,18 @@ onMounted(() => {
...
+
+function getLocalizedName(
+  names: Array<{ locale: string, firstName: string, lastName: string }>
+) {
+    const currentLocale = localeStore.activeLocale.code
+    const localePrefix = currentLocale.split('-')[0]
+
+    const selectedName
+    = names.find(name => name.locale === currentLocale)
+      || names.find(name => name.locale.startsWith(localePrefix))
```
ermish, line 115: in a future PR: we really shouldn't hack around the locales having a `-` or not. I think this is a good, quick fix you did here, but we should really fix the locales to be more consistent.

ermish (approved):
lgtm!

PR #1452, opened 2025-08-12
fix: `createFacility` and `updateFacility` logic for mutation

### LINKED ISSUE #1451: Bug adding facilities to the database
# Problem
When trying to create a new facility, even though all necessary information is included, when clicking "Create",  it scrolls to the bottom of the page and does not create the account with no error messages.

# Success
Moderators should be able to create a facility entry successfully.

### Requirements

When clicking "Create",
- [ ] it should show the toast message that the action was successful, and if not, shows an error.
- [ ] you should get redirected to the Facilities overview

ljbroersen:
✅ Resolves #1451

## 🔧 What changed
Before, the code assigned the whole mutation response to `selectedFacilityData`, expecting it to be the facility itself. But the actual facility data was nested inside a property like `createFacility`. Because of this, the code was looking in the wrong place, causing errors.

Now, the code correctly grabs the facility data inside the mutation response (eg, `serverResponse.data.createFacility`) before setting `selectedFacilityData`. This makes sure the rest of the code works with the real facility object and avoids type and runtime errors.

Also, latitude and longitude are now converted to strings only if they exist, which stops problems when those values are missing or null.

## 📸 Screenshots

-   ### Before
image

### DIFF
--- a/components/moderation-panel/ModEditSubmissionForm.vue
+++ b/components/moderation-panel/ModEditSubmissionForm.vue
@@ -446,16 +446,29 @@ function initializeSubmissionFormValues(submissionData: Submission | undefined) 
                 facilitiesStore.facilityData.find(facility => facility.id === facilityId))
             .filter((facility): facility is NonNullable<typeof facility> => facility !== undefined)
 
-    healthcareProfessionalSections.acceptedInsurance
-    = submissionData?.healthcareProfessionals?.[0]?.acceptedInsurance ?? []
-    healthcareProfessionalSections.additionalInfoForPatients
-    = submissionData?.healthcareProfessionals?.[0]?.additionalInfoForPatients ?? ''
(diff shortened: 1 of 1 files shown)


ljbroersen:
> Did you check if this was changed for also...update, delete and for the submission and healthcare professional.
> 
> They all were using the same call and could have been affected

Healthcare professional:
- Creating works as intended
- Update throws a different error in the console but doesn't hinder the actual updating process.
- Delete works as intended

Submission
- Updating works as intended
- Adding a submission through the homepage works as intended. Appears in the submission list in the moderation environment afterwards

Rejecting / Accepting a submission does not make it move away from the "For Review" page though. It marks them as "Accepted" / "Rejected" and shows them in those sections, but the entry is also still visible on the "For Review" overview.

ermish (approved):
LGTM

PR #1477, opened 2025-09-01
fix: english and japanese prefecture names must match

### LINKED ISSUE #1396: English prefecture name and Japanese prefecture name has to be match
# Problem
Now on Facility/ Submission there is no mathcer for English/Japanese prefectures name

# Success
[What this task should accomplish when it's considered completed.]
### Requirements

- [ ] Add
- [ ] Requirements
- [ ] here

# Implementation notes

ljbroersen:
✅ Resolves #1396 

## 🔧 What changed
- When you create or update a Facility, it will check if both Japanese and English names of the Japanese prefectures match. If not, it will throw an error.
- When selecting the English prefecture name, it will autofill the Japanese equivalent.
- Added a string to the locales for the error message.
- I also edited `ModCreateFacilityOrProfessionalTopbar.vue` to add a line to invoke `resetCreateFacilityFields()` when discarding unsaved data. This seemingly wasn't there before, only for creating a HP entry and then discarding it.

## 📸 Screenshots
-   ### After
image
image


NabbeunNabi (approved):
LGTM. Great work. Let's see what @ermish thinks

components/moderation-panel/ModEditFacilitySection.vue
```diff
@@ -447,4 +468,26 @@ watch(() => healthcareProfessionalsToAddToFacility.value, newValue => {
...
+        facilityStore.facilitySectionFields.prefectureEn,
+        facilityStore.facilitySectionFields.prefectureJa
+    ],
+    ([en, ja]) => {
+        prefectureNameMatchError.value = !checkPrefectureNameMatch({ prefectureEn: en, prefectureJa: ja })
+    },
+    { immediate: true }
+)
+
+watch(
```
ermish, line 485: I think we can combine this watch with the one above to save some compute time

stores/facilitiesStore.ts
```diff
@@ -215,6 +221,11 @@ export const useFacilitiesStore = defineStore('facilitiesStore', () => {
     }
 
     async function updateFacility(): Promise<ServerResponse<Facility>> {
+        const { prefectureEn, prefectureJa } = facilitySectionFields
```
ermish, line 224: happy to see the validate here! I think we should make a function for validateCreateFacility that can be reusable. This can be the same one used by the component as well and then we can move in the prefectureMatch check in the broader validation.

ermish (changes requested):
this looks great! added a few comments, but love this feature

stores/facilitiesStore.ts
ljbroersen: Wouldn't it be better to keep the validation as it is now (for only the prefecture names) and then create a separate issue for `validateCreateFacility` and `validateUpdateFacility` afterwards?

ermish:
@ljbroersen you're still doing the centralized validation method for this right? Just checking if this was waiting on me

ljbroersen:
Yes I am! Don't worry

PR #1544, opened 2025-10-15
fix: home cy tests

### LINKED ISSUE #1509: bug: fix home.cy tests
# Problem
Because the desktop is being redone, the search bar tests will continue to fail, as users are unable to open it. I will mark them as skipped for now, but will need to be remade after the desktop work is complete 
# Success
home.cy tests should all pass 
### Requirements

- [ ] Add
- [ ] Requirements
- [ ] here

# Implementation notes

bnogimauemura:
✅ Resolves #1509 

## 🔧 What changed

- Updated Cypress tests to improve stability and reliability.
- Handled rendering delays, visibility issues, and transient overlays to make tests pass consistently in both local and CI environments.

**_Since there are a few skipped tests remaining, I will use this PR as a base to continue fixing the tests in `home.cy.ts`._**

## 📸 Screenshots

-   ### Before

-   ### After

### DIFF
--- a/components/SearchResultsList.vue
+++ b/components/SearchResultsList.vue
@@ -40,7 +40,7 @@
         >
             <div
                 id="searchResults"
-                class="results-list flex flex-col pb-40 landscape:pb-4"
+                class="results-list flex flex-col pb-40 mb-40 landscape:pb-22 landscape:mb-0"
                 data-testid="search-results-list"
             >
                 <div


bnogimauemura (commented):
After trying to ensure the element was fully rendered and accessible at its center point — including using `document.elementFromPoint()` and adding waits — the test kept timing out because the element wasn’t immediately painted in the viewport. To solve this, I switched to a more reliable approach:
1. Used `scrollIntoView()` to make sure the element was in the viewport.
2. Added `.should('be.visible')` to confirm Cypress could see the element.
3. Used `.click({ force: true })` to bypass any transient overlays or small rendering delays, ensuring the test passes both locally and in CI.
4. Extended the URL assertion timeout to ensure Cypress waits long enough for the navigation to complete.

Although I used `force: true`, I am still a bit cautious about it — if using it here is not recommended, I’m happy to explore alternative approaches.

NabbeunNabi (approved):

PR #1622, opened 2026-01-04
feat: fixing CI lint and adding shared linter settings

ermish:
## 🔧 What changed
- added in shared IDE settings for eslint so devs don't have keep configuring their IDE formatter settings manually to match our codebase eslint styles. 
- ran yarn lint --fix and fixed any remaining issues

## 🧪 Testing instructions
- check yarn lint --fix works with no errors
- CI should pass

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,8 +27,7 @@ eslint-results.sarif
 .env.test
 
 # VS Code
-.vscode
-!/vscode/settings.json
+!.vscode/settings.json
 
 # Jetbrains IDEs
 .idea
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,37 @@
+{
+    // Use ESLint as the formatter for JavaScript, TypeScript, and Vue files
+    "editor.defaultFormatter": "dbaeumer.vscode-eslint",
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -27,7 +27,7 @@ eslint-results.sarif
 .env.test
 
 # VS Code
-.vscode
+!.vscode/settings.json
```
NabbeunNabi: Would we just want to exclude this one but make sure anything else in `.vscode` isn't committed?

.vscode/settings.json
```diff
@@ -0,0 +1,37 @@
+{
+    // Use ESLint as the formatter for JavaScript, TypeScript, and Vue files
```
NabbeunNabi: Why is this highlighted red on my GitHub app? Just interesting

ermish: Weird I see that too. Probably a GitHub thing 🤷‍♂️

.gitignore
ermish: I was thinking that too, but then realized the recommended extensions file would be good to have as well. Maybe we leave it open in case other files might be good to add too and just monitor in reviews 🤔

PR #1623, opened 2026-01-04
feat: major package updates including breaking change fixes

### LINKED ISSUE #1598: Update all the dependencies!
# Problem
We have some breaking changes in the new versions of some packages such as Nuxt 4. Let's update them, make necessary changes, and verify everything works

### LINKED ISSUE #1604: chore: migrate to advancedMarker on map
Google Map's is deprecating the Marker element we're currently using. They encourage using AdvancedMarker instead so, let's port it to that.

`As of February 21st, 2024, google.maps.Marker is deprecated. Please use google.maps.marker.AdvancedMarkerElement instead. At this time, google.maps.Marker is not scheduled to be discontinued, but google.maps.marker.AdvancedMarkerElement is recommended over google.maps.Marker. While google.maps.Marker will continue to receive bug fixes for any major regressions, existing bugs in google.maps.Marker will not be addressed. At least 12 months notice will be given before support is discontinued. Please see https://developers.google.com/maps/deprecations for additional details and https://developers.google.com/maps/documentation/javascript/advanced-markers/migration for the migration guide.`

ermish:
✅ Resolves #1598 
✅ Resolves #1604

# 🔧 What changed

### bug fixes
- 🐛 fixed instances where we weren't using portait and landscape media queries (also removed `sm:` and `md:` so they aren't misused anymore)

### new features
- ✨Added umami tracking when a pin is clicked! 
- 🗺️ fixed pins to use new maps, tailwind, theme management changes 
- 🚧Migrated from Marker to AdvancedMarker for map since google is deprecating it.  (This took me many hours to finally get right)
- ❤️changed the pin icon from a cross to a heart to match our main icon 
- made a utility for accessing theme colors in JS

### package upgrade break change fixes
- 🥂 Updated vue toasts to new plugin format
- 💨 Migrated to new tailwind postcss plugin approach
- :painful: Updated nuxt config format
- fixed footer for tailwind changes
- updated snapshot tests
- updated github ci action versions
- updated node versions in build, netlify, and github actions

# 🧪 Testing instructions
- make sure everything looks ok in mobile and desktop view. 

## Known issues
- lint job is failing unrelated to this PR. I added a PR to fix our CI builds here #1622 

- js error
image

this is caused by google maps api and vue3-google-maps libraries not having proper type alignment. We could either rewrite our entire maps implementation or just wait for a library update. Since it's not breaking out app, I've voted to wait until it's fixed by them. 
You can follow this [github issue](https://github.com/inocan-group/vue3-google-map/issues/362#issuecomment-3707184354) related


components/MapContainer.vue
```diff
@@ -19,135 +20,202 @@
...
-                            :class="[
-                                'w-[45px] h-[73px]',
-                                searchResultsStore.activeFacility?.id === location.id
-                                    ? 'text-secondary fill-secondary'
-                                    : 'text-primary fill-primary',
-                            ]"
-                        />
-                    </div>
-                </CustomMarker>
+                    <template #content>
```
NabbeunNabi: Why are there two templates in this file? Is this based on how Google says to do it?

components/MapContainer.vue
```diff
@@ -19,135 +20,202 @@
...
+                        <img
+                            :src="renderMarkerIcon(location.id)"
+                            class="w-[45px] h-[73px] block gmp-clickable"
```
NabbeunNabi: I know px was copied from above but just curious if this is the best choice for choosing width and height. Can of course be updated in future PR

tailwind.config.js
```diff
@@ -1,9 +1,6 @@
 /** @type {import('tailwindcss').Config} */
 
-// eslint-disable-next-line
-const colors = require("tailwindcss/colors");
-
-module.exports = {
+export default {
```
NabbeunNabi: With updating Tailwind from what I recall, postcss and the tailwind config are moved with vite.

The files aren't necessary and it is just in css as explained [here](https://tailwindcss.com/docs/upgrade-guide)

And I believe you would put it in Nuxt this [way](https://tailwindcss.com/docs/installation/framework-guides/nuxt).

Maybe this can be updated again later if it is too much refactoring

NabbeunNabi (changes requested):
Left some comments. But this is a lot of great work! Thanks for doing this

components/SearchResultsListItem.vue
```diff
@@ -36,7 +36,7 @@
                     <div
                         v-for="(spokenLanguage, index) in formattedLanguages"
                         :key="index"
-                        class="px-2 py-[1px] mr-2 border border-primary/40 rounded-full shadow text-sm
+                        class="px-2 py-px mr-2 border border-primary/40 rounded-full shadow text-sm
```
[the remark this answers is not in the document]
ermish: I was just following the tailwind rule recommendations they gave but yeah i like your point more. Also why is it following after a px-2 😆 
I'll update this one

tailwind.config.js
ermish: ok dug into the docs and I remember my concern with the vite approach vs postcss. I didn't want us to lose vendor prefixing from postcss, but I found a line `Additionally, in v4 imports and vendor prefixing is now handled for you automatically, so you can remove postcss-import and autoprefixer if they are in your project:` that looks like they have prefixing support built in, so updating it to the recommended approach you added 💯

components/MapContainer.vue
ermish: you can have nested templates in vue 👍 
Yes, this is what the AdvancedMarker accepts. Without the matching template tag, it breaks 
image

components/MapContainer.vue
ermish: great callout

NabbeunNabi (approved):
Great work and improvements. 💃 

I am still not sure if we need the tailwindconfig at all since it is all handled in the css file. But I don't want to hold this up as that can be looked at later as we want to get rid of the marker that will be deprecated.

Thanks for this!

ermish:
> Great work and improvements. 💃
> 
> I am still not sure if we need the tailwindconfig at all since it is all handled in the css file. But I don't want to hold this up as that can be looked at later as we want to get rid of the marker that will be deprecated.
> 
> Thanks for this!

oh, you're right! I'll move portrait and landscape to css as well, then delete the file 👍

PR #1626, opened 2026-01-06
fix: bottomsheet scrolling on smaller viewports

### LINKED ISSUE #1600: bug: map bottomsheet content not fully visible on smaller viewports
# Problem
The bottomsheet on the map view cuts off content on smaller viewports.

In addition, another minor UI flash issue during navigation between search and specific facilities if the map needs to shift.

# Success
The two above issues are fixed.

### Requirements

- [ ] Bottomsheet content is viewable regardless of viewport size with scrollability as needed
- [ ] UI flash should not happen when navigating between search and specific facilities

# Implementation notes
Review the current library being used and adjust as necessary

jonathanrogenes:
✅ Resolves #1600 

## 🔧 What changed
Changed parent div in BottomSheet.vue as styling was overriding the child div by hiding overflow on y axis. 
Changed parent div in SearchResultDetails.vue to allow scrolling / full viewing on smaller viewports.
Hammer on bottomSheetMain.value was overriding scrollability by dragging the entire sheet, so these lines were removed. This functionality is already on the bar/handle at top of bottomsheet.

## 🧪 Testing instructions

Check in deploy preview.
## 📸 Screenshots

-   ### Before

https://github.com/user-attachments/assets/fefd7edf-8d92-49d0-9647-23346a55c10b

-   ### After

https://github.com/user-attachments/assets/58630dc5-b5ce-4d3a-b4cc-19079b2e7b60


jonathanrogenes:
@NabbeunNabi Thanks for pointing that out; I just checked it on mine and it is not... immediate thought is perhaps its a padding issue.

NabbeunNabi:
> @NabbeunNabi Thanks for pointing that out; I just checked it on mine and it is not... immediate thought is perhaps its a padding issue.

It is still hiding the overflow. That usually disables scroll bars.

Wonder if we just want to disable the x but not the y?

ermish:
https://github.com/user-attachments/assets/206fca83-f563-40c3-8dfe-4cf77ce57ffa

@jonathanrogenes can you make the scrolling movement match the list style where the whole card moves up at the same time as the swipe? The scrolling feels a bit awkward on the half card height.

jonathanrogenes:
> Check out the last comment

It is working now! please let me know thoughts

ermish:
> > > Check out the last comment
> > 
> > 
> > 
> > It is working now! please let me know thoughts
> 
> When I'm scrolling to make it the full size before scrolling, it scrolls and makes it the larger one. I think @ermish is saying it should scroll only on the full size and should scroll after it is full size. At the same time is a little jarring for me.
> 
> But I could be wrong 

I agree it with Will, it feels natural in the results list but here it feels a little jarring. I thiiink the problem is the details page just needs to start bigger. But then the map pin won't be seen 🤔 migut need to think through this a bit, but I'm thinking it expands first then can be scrolled only when expanded would be best for our current design 🤔

jonathanrogenes:
Latest commit now allows scrolling whilst keeping Hammer in, but now its back to dragging and scrolling at same time

ermish:
Not the greatest video but I noticed the swipe up on the search results lost its smoothness now. 
I think the search result details part is fine to merge, but can we fix the other part first? I can help put this weekend if you're stuck @jonathanrogenes

jonathanrogenes:
> Interesting, latest commit does what we want when viewing doctor details after clicking on a location from the map! - drags first when you swipe on the content, and then becomes scrollable once its large. However, when you view doctor details after clicking a location from the search results list, the content only scrolls and doesn't drag unless you swipe the handle.

clicking from the search results list, the sheet is typically already at 25% position (if it was previously expanded), so isScrollingEnabled gets set to true immediately, allowing scrolling instead of dragging. This is understood because when you click from the list while its in initial 50% position, it functions as we want.

jonathanrogenes:
Everything is working! Please check your deploy preview.

PR #1639, opened 2026-01-10
chore: update PR template to remove preliminary check requirement

NabbeunNabi:
✅ Resolves #[Issue number]

## 🔧 What changed
As discussed before in the frontend either @ermish or @NabbeunNabi can approve. So removed the extra step that also is not necessarily clear for newcomers

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,8 +1,8 @@
 ✅ Resolves #[Issue number]
+
 - [ ] PR title follows [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/#summary) specifications
 - [ ] PR assignee has been selected
 - [ ] PR label has been selected
-- [ ] @NabbeunNabi has been selected for preliminary review
 
 ## 🔧 What changed
 
@@ -10,8 +10,5 @@
 
 ## 📸 Screenshots
 
--   ### Before
-
--   ### After
-
-
+- ### Before
+- ### After


components/onboarding/WelcomeScreen.vue
```diff
@@ -1,120 +1,154 @@
 <template>
-    <div
-        class="relative h-screen w-full overflow-hidden bg-primary-bg"
-        :style="{ opacity: isFadingIn ? 1 : 0, transition: 'opacity 0.5s ease' }"
+    <Transition
```
ermish: Any reason you prefer the transition over native tailwind? 

Personally, I like the more fine-grained control of tailwind to mix different animations and it keeps us consistent to just tailwind for styling across the app.

components/onboarding/WelcomeScreen.vue
```diff
@@ -1,120 +1,154 @@
 <template>
-    <div
```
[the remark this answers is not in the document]
ermish: Moving the style section tags into tailwind though make sense if you can keep the same animations 👍 if you need to keep the keyframes, we can move into tailwind.css to have a reusable animation 

If you need help, I can help fix it

ermish (changes requested):
Needs a few UX changes

components/onboarding/WelcomeScreen.vue
NabbeunNabi: this was not intended to be in this PR. Just tired and was playing around with stuff and added all the files. Should just be the template. Let me fix this quickly

PR #1649, opened 2026-01-16
fix: try to fix home.cy.ts error

Ghezz9:
✅ Resolves #[Issue number]

WIP

## 🔧 What changed

Google Maps `MarkerClusterer` was still running async calculations (`getClusters`) while the component was unmounting during navigation. This caused a race condition where the clusterer tried to access properties `range` that no longer existed.
I've added `onBeforeRouteLeave` to destroy the `MarkerClusterer` before navigation starts.

On the first commit i've only added `onBeforeUnmount` but that one seems not to change the issues (strange i don't know why maybe i was cleaning the marker in the wrong way), still used but i think `onBeforeRouteLeave` is more strong in term destroy before starting Vue create that component, maybe more safe in term of race condition.

Feel free to reply to my doubt

## 📸 Screenshots

- ### Before
- ### After

### DIFF
--- a/components/MapContainer.vue
+++ b/components/MapContainer.vue
@@ -55,6 +55,7 @@
 
 <script setup lang="ts">
 import { ref, watch, nextTick, onMounted, onBeforeUnmount } from 'vue'
+import { onBeforeRouteLeave } from 'vue-router'
 import { GoogleMap, AdvancedMarker, MarkerCluster } from 'vue3-google-map'
 import type { Renderer } from '@googlemaps/markerclusterer'
 import { useUmami } from '~/composables/useUmamiTracking'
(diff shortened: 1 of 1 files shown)


NabbeunNabi:
> hey @Ghezz9 @NabbeunNabi int retrospect, I mentioned in my PR, but I should have noted in slack that the range error is part of the vue3-google-maps library that has an issue being tracked for a fix. It doesn't break anything so we should be fine to ignore.
> 
> If this is unrelated to the range console error and just about tests, then disregard

Ahh yea I didn't remember that.

The main goal is to fix the test. Because we don't want it to fail due to the range thing.

This test actually caught the privacy and terms and conditions being hidden before.

So would like to keep it. But thanks for the context 💃

PR #1672, opened 2026-02-26
feat: filter language rework

### LINKED ISSUE #1631: Rework Languages Filter
# Problem
Let's rework languages to be more visible. 
Likely just listing them out in two columns, or in multiple rows. English and Japanese can be colored/highlighted slightly differently in order to give preference as they are the most commonly used.

Check out the design doc here:
https://docs.google.com/document/d/1WXBaas2gklyNH3Cm6Ker-Zszm-kymExnmv-6boLUyUk/edit?usp=sharing

# Images
-Before
-After

jonathanrogenes:
✅ Resolves #1631

## 🔧 What changed
Header for Select Languages
Entire new selection styling
Button style grid of all the languages in the system
Search functionality remains unchanged
Helper function for selecting language changed to provide simpleText rather than displayText as in the data files

## 🧪 Testing instructions
Please test a search on your machine. Be aware the vast majority of these languages aren't registered with a facility
## 📸 Screenshots

- ## Before
Screenshot 2026-02-26 at 19 10 21

- ### After
Screenshot 2026-02-26 at 19 04 23

### DIFF
--- a/utils/facilitiesUtils.ts
+++ b/utils/facilitiesUtils.ts
@@ -1,5 +1,6 @@
 import { prefectureDisplayName } from '~/stores/locationsStore'
 import type { CreateFacilityInput, UpdateFacilityInput, PhysicalAddressInput } from '~/typedefs/gqlTypes'
+import type { Prefecture } from '~/typedefs/locationTypes'
 
 export interface FacilityValidationResult {
     valid: boolean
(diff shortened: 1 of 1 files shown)


i18n/locales/en.json
```diff
@@ -110,7 +110,7 @@
   },
   "searchBar": {
     "selectSpecialty": "Select a specialty...",
-    "selectLocation": "Select a location...",
+    "selectLocation": "Select Location",
```
aaronculbert: I'm not against this change but what was the reasoning behind it?

The other locale strings follow a similar pattern so it might be worth updating all of them to follow the same convention without the `...`

PR #1691, opened 2026-03-13
fix: buttons now scroll to their respective sections on the about page.

### LINKED ISSUE #1651: bug: buttons not working on about page
# Problem
The buttons are supposed to scroll the user to the appropriate section on the page. But it currently is not doing anything when pressed

# Success
Buttons scroll to the appropriate sections

### Requirements

- [ ] Buttons scroll again when pressed
- [ ] the corresponding buttons go to the correct section (and not another one)
- [ ] no other functionality on the page is affected

# Implementation notes
check out `pages/about.vue`

aaronculbert:
✅ Resolves #1651 

## 🔧 What changed
Updated smooth scroll implementation to use native APIs to fix each button scrolling to their respective section.

## 🧪 Testing instructions
Open the about page and start by clicking on Learn More and then navigate with the various buttons.

## 📸 Screenshots

https://github.com/user-attachments/assets/97aac93f-185a-4ab7-ae57-0aae75bb0e84

### DIFF
--- a/tests/vitest/snapshotTests/__snapshots__/about.spec.ts.snap
+++ b/tests/vitest/snapshotTests/__snapshots__/about.spec.ts.snap
@@ -12,7 +12,7 @@ exports[`about.vue (vitest snapshot) > renders correctly 1`] = `
     </div>
     <anonymous-stub role="img" alt="Characters Together SVG Icon" title="Characters Together" class="landscape:w-1/3 landscape:mr-4"></anonymous-stub>
   </div><!-- Our Story section -->
-  <div id="about-header-container" data-testid="about-header-container" class="flex justify-center items-center">
+  <div id="about-header-container" data-testid="about-header-container" class="scroll-mt-4 flex justify-center items-center">
     <div id="about-header" data-testid="about-header" class="flex w-4/5 my-12">
       <div class="flex-1 border-currentColor/70 border self-center"></div>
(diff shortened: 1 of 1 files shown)


aaronculbert:
No worries Davide, I'll answer your questions anytime!

> * `about.spec.ts.snap` how do you know that was a file to change?

I didn't realise initially but the Snapshot Tests failed and this was how to fix the test.

> * So if i understood well `scroll` was broke, before when you click on contributors etc it doesn't worked well?

Yep, nothing happened when clicking on the buttons.

> * Adding `scroll-mt-4` on class resolve this on everytime we will have to handle a scroll after click?

scroll-mt-4 is a Tailwind utility. It adds space above an element that only applies during scrollTo-like events.

I added this to have that space above each section header. If we removed it, the scrolling would happen in a way that the navbar cuts off the section header text.

> * You delete 90% `useSmoothScroll` why? passing to a native solution?

Right, initially it was implemented extending the original solution but found that there is a simpler way using browser's native APIs nowadays. Usually try to be careful doing this for backwards compatibility reasons but `scrollIntoView` is well-adopted across all modern browsers for quite a few years now and we don't have a large user base that requires supporting super old browsers.

Also generally simpler code is a win whenever we find the opportunity 👍

aaronculbert:
@NabbeunNabi Do we need to do anything to get Pinia test run to finish or is that a known issue?

Also, feel free to merge if you fix the Pinia run or we're good to bypass. 🙇

PR #1726, opened 2026-04-11
fix: form validation and error placement

BobbyRobBob:

## 🔧 What changed
I noticed that the validation message was being displayed by default on page load and was above the drop-down while others were below. Amended on both primary and secondary languages and made some small spacing changes so the errors aren't too close to next headers.

## 🧪 Testing instructions
-Run yarn dev:localserver
-Click on "Add a doctor"
-Observe error is displayed below drop-down only if user try's to submit empty.

Expected result: Error should not be displayed on load and should be below drop-down if triggered.

## 📸 Screenshots

- ### Before
 #### First load of screen
 
image

 #### After clicking submit 
image

- ### After
 #### First load of screen
image

 #### After clicking submit 
image

## Summary by CodeRabbit

* **Style**
  * Adjusted vertical spacing of validation containers and restructured how language validation messages are positioned and displayed in the submission form.

* **Bug Fixes**
  * Corrected initial validation state so the primary language error is not shown before user interaction.

* **Tests**
  * Updated end-to-end test to assert the spoken-language validation message is initially hidden and appears after submission.

### DIFF
--- a/components/SubmissionForm.vue
+++ b/components/SubmissionForm.vue
@@ -31,7 +31,7 @@
                 :placeholder="t('submitPage.location')"
                 @blur="initialValidationCheck(location, 'googleMaps')"
             >
-            <div class="google-maps-validation-container flex text-error text-xs font-sans h-3 mt-1 mb-3">
+            <div class="google-maps-validation-container flex text-error text-xs font-sans h-3 mt-1 mb-4">
                 <span
                     v-show="!isValidInput.googleMapsUrl.value"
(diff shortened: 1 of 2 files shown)


NabbeunNabi (changes requested):
Since you are working on this do you mind also aligning the error message to the left for cleaner ui instead of centered. So lined up with the language input. Also hopefully this validation is less necessary when the button becomes `disabled` if you do that follow up

PR #1751, opened 2026-06-11
perf: Graphql healtcare professional update

tanukilandslide:
✅ Resolves #[1469]

## 🔧 What changed
When updating a healthcare professional, the graphql mutation gets created by looking at what fields have changed. It also creates the payload to be sent as well.

## 📸 Screenshots

- ### Before
- ### After


aaronculbert:
This [PR](https://github.com/ourjapanlife/findadoc-server/issues/998) has the backend fix. Once merged, we can run some tests to see if this is now working and then iterate on the code itself.

aaronculbert (commented):
Following up on my earlier review — I dug into this a bit more against the current `main` with the backend fix in place, and it's actually good news for you here. 🙂

It turns out the frontend on `main` **already does the partial update correctly**. The diff-based logic (only sending the fields that actually changed) was added to `main` in a separate, unrelated PR (#1705, "frontend code cleanliness"), and the real blocker was always server-side: `updateHealthcareProfessional` was failing purely because of a JSONB serialization bug on the backend (#998), which is now fixed by server #1002.

So the root cause was backend, not frontend — and your work digging into this is exactly what surfaced that backend bug in the first place, which was genuinely valuable. The frontend just happened to get its matching fix in that other PR while this was in flight.

Because of that, I think it's totally fine to **close this PR** — `main` + the backend fix already cover the partial-update behavior, so there's nothing left to merge here. Definitely not wasted effort: the backend bug you helped identify was the actual thing standing in the way, and it's resolved now.

Thanks again for chasing this one down! 🙏

PR #1771, opened 2026-07-28
fix: ModInputField now checks all numbers for full width characters and converts them to half width

tanukilandslide:
✅ Resolves #[1762]

## 🔧 What changed
ModInputField.vue now searches for full width numbers and changes them to half width numbers. This creates both better consistency across the site, but also ensures numbers will pass validation.

## 📸 Screenshots

- ### Before
- ### After

## Summary by CodeRabbit

* **Bug Fixes**
  * Numeric input now automatically converts full-width digits to standard ASCII digits.
  * Validation and stored values remain consistent regardless of the digit format entered.

### DIFF
--- a/components/moderation-panel/ModCreateFacilitySection.vue
+++ b/components/moderation-panel/ModCreateFacilitySection.vue
@@ -33,6 +33,7 @@
                     type="text"
                     :placeholder="t('modFacilitySection.placeholderTextFacilityPhoneNumber')"
                     :required="true"
+                    :number-converter="true"
                     :input-validation-check="validatePhoneNumber"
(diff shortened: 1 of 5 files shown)


NabbeunNabi (changes requested):
This is nice! But can we make it an arrow function. And move it to utils. And then add some tests if possible

NabbeunNabi (changes requested):
There is a console log. If the numbers are already the correct half width what happens? Would the tests still pass for that particular function and not throw an error? Those are the type of tests I want. Not just the whole phone number but for that specific function.

Issue #1787, opened 2026-08-27
feat(seo): enable hybrid rendering for public routes


aaronculbert:
**Blocked by:** #1786, #1785

# Problem

`ssr: false` (`nuxt.config.ts:24`) plus `nuxi generate` in `prod:build` (`package.json:15`) means the production build is a single prerendered empty shell. Measured on the live site:

| Measurement | Value |
|---|---|
| Total HTML size | 3,880 bytes |
| Visible body text | **0 characters** |
| `<h1>` elements | **0** |
| `data-ssr` in payload | `false` |

Google does execute JavaScript, but rendering is queued, budgeted, and deprioritised for low-authority domains. On a site where all content requires JS plus a cross-origin GraphQL round trip to `api.findadoc.jp`, effectively nothing reaches the index.

AI crawlers are stricter — `OAI-SearchBot` (ChatGPT Search) and `PerplexityBot` execute **no JavaScript at all**, so the current shell is literally blank to them.

This ticket unblocks all of Phase 3 and Phase 4.

# Success

Public directory routes are server-rendered and return real content in the initial HTML response. Authenticated routes stay client-only.

### Requirements

- [ ] `ssr: true` set in `nuxt.config.ts`
- [ ] `my-page*`, `moderation*`, and `login` remain client-only via `definePageMeta({ ssr: false })`
- [ ] `routeRules` configured — prerender/ISR for public routes, SPA for authenticated ones
- [ ] `curl https://www.findadoc.jp/` returns a document with a real `<h1>` and non-trivial body text
- [ ] No hydration mismatch warnings in the console on any public page
- [ ] `yarn test` passes, including the Playwright e2e suite
- [ ] Auth0 login flow verified working end to end on a preview deploy

# Implementation notes

**This ticket carries a deployment-model decision and should not start before it is made.** Two viable shapes:

1. **Pure prerender** — keeps the static Netlify model that `online-booking.md` §A.4 assumes. Requires regenerating ~1,100 pages whenever a moderator approves a facility, via a build webhook from `findadoc-server` or a scheduled rebuild. The staleness window needs to be acceptable.
2. **ISR / server runtime** — findadoc-web gains a server runtime. Removes the rebuild-cadence problem and scales to the Phase 4 locale matrix (~11,000 pages), but changes the deployment model.

[...]

PR #1800, opened 2026-08-30
refactor: remove SSR blockers ahead of enabling server rendering

### LINKED ISSUE #1786: refactor: remove SSR blockers ahead of enabling server rendering
# Problem

`ssr: false` (`nuxt.config.ts:24`) is the root cause of the site being unindexable, but flipping it will break at runtime until a handful of browser-only assumptions are removed.

The good news is the blast radius is small and fully enumerable. **All six plugins are already `.client.ts`** — including `vue-auth0-plugin.client.ts` — so authentication never runs server-side. What remains:

- **10 module-scope `useToast()` calls**, which execute at import time and will run during SSR: `stores/searchResultsStore.ts:19`, `components/TopNav.vue:224`, `components/SubmissionForm.vue:169`, `components/HamburgerMenu.vue:335`, plus 6 files under `components/moderation-panel/`.
- **2 components depending on browser-only libraries**: `MapContainer.vue` (`vue3-google-map`) and `BottomSheet.vue` (`hammerjs`).
- **16 files referencing `window`/`document`/`localStorage`**, which need auditing for server-side execution paths.

Doing this as its own PR keeps #1787 reviewable — that one should be a config change, not a config change plus 20 file edits.

# Success

The codebase is SSR-safe. `ssr: true` can be set without runtime errors.

### Requirements

- [x] All module-scope `useToast()` calls hoisted into setup scope
- [x] `MapContainer.vue` and `BottomSheet.vue` wrapped in `<ClientOnly>`
- [x] The 16 files touching `window`/`document`/`localStorage` audited; guarded with `import.meta.client` where they can run server-side
- [x] `yarn test` passes
- [ ] Verified by temporarily setting `ssr: true` locally and confirming no server-side errors — the flag itself flips in #1787

# Implementation notes

Find the toast calls with:

```bash
grep -rn "^const toast = useToast()" stores utils composables components
```

`vue-toastification` is registered in a `.client.ts` plugin, so `useToast()` has no provider during SSR — hence the failure. Moving the call inside `setup()` is sufficient; no need to make toasts themselves client-only.

For the `window`/`document` audit:

```bash
grep -rl "window\.\|document\.\|localStorage" stores utils composables components pages
```

Most will be inside lifecycle hooks or event handlers and are already safe. The ones to fix are module scope and anything running in `setup()` directly.

NabbeunNabi:
## Summary
- Makes the app SSR-safe without flipping `ssr: true` (that stays in #1787): wrap `MapContainer` and `BottomSheet` in `<ClientOnly>` / `*.client.vue`, and guard `window`/`document`/`localStorage` with `import.meta.client`.
- Provide toasts from the existing **client plugin** (`$toast`) and `useAppToast()`. Server/prerender gets a no-op so Nitro never imports `useToast` from vue-toastification's CJS build (that named export does not exist and breaks `nuxi generate`).
- Share viewport breakpoints in `utils/viewport.ts` so `nuxt.config` and `useScreenOrientation` stay in sync. Orientation refs start at the desktop fallback size, then update in `onMounted`.
- Drop Nuxt 2 `hid` from `app.head.meta`. Unhead dedupes on `name` / `property`.

Closes #1786

## Test plan
- [ ] Temporarily set `ssr: true` in `nuxt.config.ts`, run the app, and confirm no server-side `window`/`document`/`localStorage` errors
- [ ] Confirm map and bottom sheet still render on client (desktop landscape and mobile portrait)
- [ ] Confirm theme, onboarding, auth login/logout, toasts, and moderation search still work
- [ ] `nuxi generate` (or Netlify `yarn prod:build`) prerenders without `useToast` / `index.cjs` errors
- [ ] `yarn test`
- [ ] Revert `ssr` to `false` after local verification on this PR — do not merge that flag here

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved stability during page loading and server-rendered access by preventing browser-only features from running prematurely.
  - Improved map and bottom-sheet rendering across desktop and mobile layouts.
  - Prevented theme, scrolling, authentication, onboarding, and form interactions from causing errors during initial loading.

- **Improvements**
  - Standardized notifications across the application for more consistent toast behavior.
  - Centralized viewport breakpoints for more predictable responsive layouts.
  - Improved initial viewport sizing and theme handling.


aaronculbert (approved):
Thanks! Looks good, tested the live preview a bit especially on mobile. No issues that I could see.

aaronculbert (changes requested):
Ah hang on, I found an issue on mobile, the bottom sheet is broken - you can't pull it up again after clicking on the map.

NabbeunNabi:
> Ah hang on, I found an issue on mobile, the bottom sheet is broken - you can't pull it up again after clicking on the map.

could you before?

aaronculbert:
> > Ah hang on, I found an issue on mobile, the bottom sheet is broken - you can't pull it up again after clicking on the map.
> 
> could you before?

Yeah just confirmed on the live site.

aaronculbert (approved):
LGTM - confirmed working on mobile!

PR #1802, opened 2026-08-30
feat(seo): enable hybrid rendering for public routes

### LINKED ISSUE #1787: feat(seo): enable hybrid rendering for public routes

NabbeunNabi:
## Summary
- Stacked on #1800 (`brammer/remove-ssr-blockers`). Do not merge until that PR lands.
- Sets `ssr: true` and prerenders public pages (`/`, `/about`, `/terms`, `/privacypolicy`, `/submit`) so crawlers get real HTML (`<h1>` + body text). Keeps the current static Netlify `nuxi generate` model; ISR would need a server runtime (see #1787).
- `login`, `my-page*`, and `moderation*` stay client-only via `routeRules` and `definePageMeta({ ssr: false })`.

Closes #1787

## Test plan
- [ ] Confirm this PR targets `brammer/remove-ssr-blockers`, not `main`
- [ ] `curl http://localhost:3000/about` (and `/`, `/terms`) includes an `<h1>` and visible copy
- [ ] `/login`, `/my-page`, `/moderation` still work as SPA (Auth0 login end to end)
- [ ] No hydration mismatch warnings on public pages
- [ ] Map / bottom sheet still render after #1800 client-only components
- [ ] `yarn test`

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

* **New Features**
  * Public pages are now server-rendered and prerendered for faster initial loading.
  * Authenticated and interactive pages continue to load client-side for reliable state handling.
  * Onboarding progress now hydrates safely from local storage and persists across visits.

* **Bug Fixes**
  * Prevented incorrect onboarding or completed-app screens from appearing before hydration.
  * Improved handling of invalid or corrupted saved onboarding data.

* **Tests**
  * Added coverage for onboarding defaults, hydration, and persistence.

### DIFF
--- a/components/MainContentContainer.vue
+++ b/components/MainContentContainer.vue
@@ -37,7 +37,9 @@
                 class="h-svh"
             >
                 <Loader />
-                <BottomSheetContainer />
+                <ClientOnly>
+                    <BottomSheetContainer />
+                </ClientOnly>
                 <ClientOnly>
                     <MapContainer
                         class="h-[calc(100vh)]"
--- a/components/MainContentContainer.vue
+++ b/components/MainContentContainer.vue
@@ -1,66 +1,74 @@
 <template>
     <div class="h-full w-full">
-        <Onboarding
-            v-if="onboardingState === OnboardingState.NotStarted"
-        />
-        <!-- Search Container -->
-        <div
-            v-if="onboardingState === OnboardingState.Completed"
-            id="search-container"
-            class="flex flex-col h-svh overflow-hidden"
-        >
-            <TopNav class="absolute top-0 left-0 right-0 z-10 mx-2" />
-            <!-- Landscape / Desktop Mode -->
+        <!-- Decide onboarding vs app on the client so SSR/prerender cannot
+             bake "not-started" HTML that ignores localStorage. -->
+        <ClientOnly>
+            <Onboarding
+                v-if="onboardingState === OnboardingState.NotStarted"
+            />
+            <!-- Search Container -->
             <div
-                v-if="isLandscape"
-                id="search-landscape"
-                class="flex flex-col flex-1 overflow-hidden "
+                v-if="onboardingState === OnboardingState.Completed"
+                id="search-container"
+                class="flex flex-col h-svh overflow-hidden"
             >
-                <Loader />
-                <div class="flex flex-1 overflow-hidden">
-                    <LeftNavbar
-                        class="bg-primary-bg w-96 z-20
+                <TopNav class="absolute top-0 left-0 right-0 z-10 mx-2" />
+                <!-- Landscape / Desktop Mode -->
+                <div
+                    v-if="isLandscape"
+                    id="search-landscape"
+                    class="flex flex-col flex-1 overflow-hidden "
+                >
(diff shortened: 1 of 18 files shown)


aaronculbert (approved):
Approving but two callouts since it will affect our SEO:

- / root page renders blank
- /submit page renders the form but defaults to hidden (revealed by JS)

Easiest way to get a quick visual is disabling JS from browser tools. Search engines can index rendered html pretty quickly, a JS pass might take some additional time.

These are existing issues that SSR has exposed. After this ticket, my recommendation is we drop the onboarding step which is basically only a splash page with no content and allow users to dive into search asap, and then remove any client-side gating on the submit form.

PR #1806, opened 2026-09-01
fix(seo): give every page a unique, branded title

### LINKED ISSUE #1783: fix(seo): titleTemplate drops the brand name from every page title
# Problem

`nuxt.config.ts:36-37`:

```ts
titleTemplate: 'Health Services in Japan',   // no %s placeholder
title: SITE_TITLE,                            // 'Find a Doc, Japan!' — discarded
```

A `titleTemplate` without a `%s` placeholder replaces the title outright rather than wrapping it. The live site confirms this:

```
$ curl -s https://www.findadoc.jp/ | grep -o '<title>[^<]*</title>'
<title>Health Services in Japan</title>
```

**No page on the site has "Find a Doc" in its title tag.** Branded searches do not match the title, and every page shares one identical title, so even once pages are indexable they cannot be told apart in results.

# Success

Every public page has a unique, descriptive title that includes the brand name.

### Requirements

- [ ] `titleTemplate` uses a `%s` placeholder, e.g. `'%s | Find a Doc, Japan'`
- [ ] Each public page sets its own `title` — index, about, submit, terms, privacypolicy
- [ ] The homepage title reads naturally and is not doubled up (`titleTemplate` should not produce "Find a Doc, Japan | Find a Doc, Japan")
- [ ] Titles stay under ~60 characters so they are not truncated in results
- [ ] Verified by curling each public page in a preview deploy

# Implementation notes

Set per-page titles with `useHead({ title: ... })` in each page's `<script setup>`. Note only three pages currently call `definePageMeta`, and all three set `layout` only — there is no existing per-page head convention to follow, so this ticket establishes one.

Titles should be translated rather than hardcoded English — see #1796. If #1796 lands first, wire titles through `t()` from the start.

9426Haruka:
## Problem

`titleTemplate` had no `%s` placeholder, so it replaced the title outright on every page. The result (verified on the live site in #1783): every public page shared the single title `Health Services in Japan`, and the brand name "Find a Doc" never appeared in any `<title>` tag.

Closes #1783

## Approach

1. **Global template moved to runtime.** Function-style `titleTemplate` must be set at runtime (via `useHead` in `app.vue`), because head config from `nuxt.config.ts` is serialized into the app manifest and functions are dropped there. I verified this: a function template in `nuxt.config.ts` silently never applied, while the same function in `app.vue` works for every page.

   A string template (`'%s | Find a Doc, Japan!'`) cannot be used alone, because pages without an explicit title then render a broken `" | Find a Doc, Japan!"`. I tested all three combinations locally:

   | Config | Result |
   |---|---|
   | `%s` string + no page title | `"< \| Find a Doc, Japan!"` (broken) |
   | `%s` string + global `title: SITE_TITLE` | `"Find a Doc, Japan! \| Find a Doc, Japan!"` (doubled) |
   | function + no page title | `"Find a Doc, Japan!"` (clean fallback) ✅ |

   The function template handles both cases: pages with a title get wrapped with the brand, pages without fall back to the brand alone. This also removes the global `title` entry that would produce a doubled title once SSR lands (#1787).

2. **Per-page titles.** Each public page (`index`, `about`, `submit`, `terms`, `privacypolicy`) now sets its own title via `useHead({ title: () => t('pageMeta.*') })`, reactive to locale changes. This establishes the per-page head convention the issue asked for — the repo previously had zero `useHead` usages.

3. **Translated titles.** Added a `pageMeta` namespace with translations for all 10 locales (en, ja, cn, de, fr, it, pt, ru, tl, vi), rather than hardcoding English. `SITE_TITLE` was also extracted to `utils/site.ts` so the brand name has a single source of truth shared by `nuxt.config.ts` and `app.vue`.

## Verification

Ran locally (`nuxi dev`) and read `document.title` for each route:

| Route | Title |
|---|---|
| `/` | `Health Services in Japan \| Find a Doc, Japan!` |
| `/about` | `About Us \| Find a Doc, Japan!` |
| `/submit` | `Submit a Provider \| Find a Doc, Japan!` |
| `/terms` | `Terms of Service \| Find a Doc, Japan!` |
| `/privacypolicy` | `Privacy Policy \| Find a Doc, Japan!` |
| `/my-page` (no explicit title) | `Find a Doc, Japan!` (fallback) |

All titles are unique, include the brand, and stay under 60 characters.

- `yarn lint:ci` ✅ (2 pre-existing a11y warnings in `WelcomeScreen.vue`, untouched by this PR)
- `yarn lint:locales` ✅
- `yarn test:unit` ✅ 47/47

One pre-existing oddity worth noting for the SEO epic: `/login` shows `"Find a Doc, Japan"` (no `!`). I traced it with a `document.title` setter breakpoint — it is written outside of unhead's update cycle, both before and after this change, so I left it out of scope here.

### DIFF
--- a/app.vue
+++ b/app.vue
@@ -7,6 +7,16 @@
 
 <script setup lang="ts">
 import { initializeGqlClient } from './utils/graphql.js'
+import { SITE_TITLE } from '~/utils/site'
 
 initializeGqlClient()
+
+// Wrap each page's title with the brand name. Pages without an explicit
+// title fall back to the brand name alone, so we never render a broken
+// " | Brand" or a doubled "Brand | Brand" title. This must be set at
+// runtime (not in nuxt.config) because function templates are not
+// serializable into the generated app config.
+useHead({
+    titleTemplate: (title?: string) => title ? `${title} | ${SITE_TITLE}` : SITE_TITLE
+})
 </script>
--- a/i18n/locales/cn.json
+++ b/i18n/locales/cn.json
@@ -1,4 +1,11 @@
 {
+  "pageMeta": {
+    "homeTitle": "日本的医疗服务",
+    "aboutTitle": "关于我们",
+    "submitTitle": "提交医疗服务",
+    "termsTitle": "服务条款",
+    "privacyTitle": "隐私政策"
+  },
   "common": {
     "menu": "菜单",
     "siteName": "Find a Doc Japan",
--- a/i18n/locales/de.json
+++ b/i18n/locales/de.json
@@ -1,4 +1,11 @@
 {
+  "pageMeta": {
+    "homeTitle": "Gesundheitsdienstleistungen in Japan",
+    "aboutTitle": "Über uns",
+    "submitTitle": "Anbieter einreichen",
+    "termsTitle": "Nutzungsbedingungen",
+    "privacyTitle": "Datenschutzerklärung"
+  },
   "common": {
     "menu": "Menü",
     "siteName": "Find a Doc Japan",
--- a/i18n/locales/en.json
+++ b/i18n/locales/en.json
@@ -1,4 +1,11 @@
 {
+  "pageMeta": {
+    "homeTitle": "Health Services in Japan",
+    "aboutTitle": "About Us",
+    "submitTitle": "Submit a Provider",
+    "termsTitle": "Terms of Service",
+    "privacyTitle": "Privacy Policy"
+  },
   "common": {
     "menu": "Menu",
     "siteName": "Find a Doc Japan",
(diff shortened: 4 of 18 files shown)


aaronculbert:
Thanks for the interest. Contributions to Findadoc start with an intro in #dev-team on our Slack ([join here](https://find-a-doc.slack.com/join/shared_invite/zt-2lb3j2t1z-O9F4thzcFktqsIXNzzjzHw#/shared-invite/email)) and getting assigned to the issue before opening a PR. Closing this for now. If you join Slack and let us know which issue you'd like to work on, we're happy to review a fresh PR.

For reference, our docs - https://docs.findadoc.jp/

PR #1815, opened 2026-09-02
feat: give the homepage a job and move the search app to /search

aaronculbert:
`/` was a splash screen and a hardcoded 3-second timer. The category step it was built around **never rendered** — nothing ever set `currentStep` to `Categories` — so a first-time visitor clicked one button, waited three seconds for a timer that did no work, and landed on an unfiltered map. The page carried no nav and no footer, so there was no link to About or Submit and no language selector on the surface most likely to be a stranger's first contact with the site.

`/` is now a content homepage. The map application moves to `/search`, entered with the homepage's filters already applied.

## It prerenders

Measured on the built output, before and after:

| | Before | After |
|---|---|---|
| `/` visible text | **0 chars** | **3,310 chars** |
| `/` `<h1>` | none | `Find a doctor who speaks your language` |
| `/` internal links | 0 | **17** into `/search?…` |
| `/npo` | — | 1,112 chars |

`/` previously emitted nothing because `MainContentContainer` sat inside `<ClientOnly>` so prerender could not bake the wrong onboarding branch — a workaround this change removes the need for.

`/search` is `ssr: false`: it mounts the Maps SDK, reads filters from the query string and fetches client-side, so a prerendered shell would only ship an empty result list.

The homepage loads **no map, no GraphQL, no search waterfall** — 0 API calls against `/search`'s 10. Auth0 still initialises on every route; that is bundle-splitting work and not addressed here.

## The form works without JavaScript

Because the page is prerendered, the form is in the HTML and clickable **before Vue hydrates** — and it is the primary control on the page. An early submit fired a native GET back to `/` and silently discarded the selection. A fast user hits this, not just a test.

It now carries `action="/search"`, `method="get"` and named selects, so that submit lands on `/search` with the filters applied — the same result as the JS path, and it works with JavaScript disabled entirely. Held by a test that submits at `waitUntil: 'commit'`.

## Search filters, from live data

| Filter | Offered | Why |
|---|---|---|
| Specialty | 31 | all of them |
| Language | **14** | every language ≥1 registered professional speaks. The enum has 40+; the rest return a guaranteed empty result. |
| Area | **33** | every prefecture with a facility, ordered by count, each showing its count so nobody picks Aichi expecting more than one |

Areas are prefectures rather than cities because city names are still unnormalised upstream — `Chuo Ward` (21) and `Chuo City` (9) are the same place. Prefecture filtering is new in `searchResultsStore`.

## Also in this change

**Onboarding is deleted** — three components, its store, its layout, its Storybook stories, its e2e spec, and the `skipOnboarding` fixture every other spec had to call before visiting any page.

**`layouts/default`** no longer locks to `h-dvh` with an internally scrolling `<main>`. That is the app-shell shape, and it left content pages unscrollable at the document level with the footer pinned to the viewport — 2,682px of homepage inside a 761px box. `/search` has its own layout now.

**The footer was `portrait:hidden`** — Privacy and Terms were unreachable on mobile entirely. It renders at every size now; the search page hides it in portrait at the call site, where the bottom sheet owns that edge.

**The header search control** called `showBottomSheet` for a sheet only mounted on the search page's *portrait* branch, so everywhere else it set state and nothing appeared. It now navigates to `/search` off the search page and opens the filter panel on it.

**Statutory NPO disclosures move to `/npo`.** The balance sheet and registration number existed only in the mobile hamburger menu — desktop visitors could not reach them at all. The registration number is plain text: it is a fact about the entity, and the directory it used to link is a third party, not a registry.

**i18n gained a message-level fallback.** Only `detectBrowserLanguage.fallbackLocale` was set, which selects a locale but does nothing for a key missing from it, so an untranslated key rendered as its own path on the page.

## Two things to flag before merge

**The translations have not had a native review.** 29 keys across nine languages, written as part of this change, on a site whose purpose is serving people who do not read the local language. The Japanese matters most. I caught one factual error in my own draft — an unsupported `認定` claim on the disclosure page — which is exactly why this needs a second pair of eyes.

**The counts are hardcoded**, measured against production on 2026-09-02: 465 clinics, 400 professionals, 14 languages, and the per-prefecture figures. The API exposes no facet endpoint, so deriving them at runtime means paging the whole facility table — the cost the homepage exists to avoid. They need re-measuring as the directory grows.

## Not addressed

The search waterfall. `/search` still makes 10 GraphQL calls and takes ~5.9s. Shipping a fast homepage that hands users into a slow search makes that contrast sharper, not softer — it is the next thing worth doing, and it is blocked on a server-side location filter.

## Tests

Lint clean, unit 55/55, snapshots 4/4, **e2e 48/48**, locale check clean, and `nuxi generate` prerenders 29 routes.

## Summary by CodeRabbit

- **New Features**
  - Added a localized homepage with search, category and prefecture browsing, workflow information, and contribution links.
  - Added a responsive search page with URL-based filters and location-based filtering.
  - Added an NPO disclosure page with organization details and statutory documents.
- **Improvements**
  - Added translated loading text and fallback translations.
  - Updated footer and menu links, branding, and legal information.
  - Improved direct access to search pages and document flow.
  - Removed the previous onboarding flow.
- **Documentation**
  - Added online booking requirements and SEO planning documentation.


NabbeunNabi (approved):

NabbeunNabi:
https://github.com/user-attachments/assets/9a4a13b1-f878-4122-8116-b3a1d1a58f2d

PR #1816, opened 2026-09-04
feat: design system, chrome restyle, and list-first search

aaronculbert:
Brings the rest of the site up to the design language the homepage established in #1815, and rebuilds `/search` around it.

Large, but mostly one change: a design system, then everything applied to it. Suggested review order is the table below.

## What changed

| Area | Summary |
|---|---|
| **Foundations** | Themes collapse to **light / dark / auto**. Component primitives and one global focus ring. Fonts move to subsetted woff2. |
| **Chrome** | TopNav, Footer, HamburgerMenu, LocaleSelector, error layout. |
| **Content pages** | `/about`, `/submit`, `/terms`, `/privacypolicy`, `/npo`, `/login`. |
| **Search** | `/search` is list-first, with URL state, in-memory filtering and an opt-in map. |

### Foundations — `assets/css/tailwind.css`, `docs/brand.md`

The dark palette now actually passes AA; the previous one used a near-white accent token, so every border glowed. An inline `<head>` script applies the stored scheme before first paint and migrates the old five-colourway keys, so returning visitors don't get a light flash. `prefers-contrast` and `prefers-reduced-motion` are honoured globally rather than as separate themes.

Primitives (`.btn-*`, `.card`, `.chip`, `.field*`, `.page-section`, `.page-title*`, `.prose-legal`) replace repeated utility strings. `docs/brand.md` documents every token with its measured contrast ratio.

8.3 MB of TTF fonts are deleted — they were declared but never actually downloaded, because nothing referenced the family. Character illustrations move to `/public` as cached files; inlining one of them put 130 KB into every prerendered page.

### Search — `pages/search.vue`, `stores/searchResultsStore.ts`, `components/search/`

Results are the primary surface; the map is opt-in and only fetches the Google SDK when opened. Filters and the open facility live in the query string, so links are shareable and Back closes the panel rather than leaving the page.

The store loads the directory once and filters in memory, so every later filter change is instant and costs no request. Anonymous search no longer blocks on Auth0 initialisation, and requests carry a real per-attempt deadline — previously a connection the server accepted and never answered left the page on skeletons forever with the retry button unreachable.

Contact values are normalised before they become links. Five of 465 facilities store a website with no scheme, which a browser reads as a *relative path*, so those buttons navigated inside our own site. One phone number keeps a trunk zero after `+81` and could not dial.

The map-first stack is removed: `SearchBar`, `FiltersPanel`, `SlidingRightPanel`, `BottomSheet*`, `bottomSheetStore`, and the `hammerjs` dependency.

## Notable fixes found along the way

- Modal backdrops used a token that **inverts to near-white in dark mode**, so scrims were a white wash.
- The global focus ring was invisible on the primary band — including the homepage's main call to action. It is now two-tone.
- `bg-opacity-40` does not exist in Tailwind v4, so the confirmation dialog rendered as an **opaque coral sheet**.
- Search components read a store that only updates via the language picker, so a visitor arriving on the `ja-JP` cookie got a Japanese UI full of English clinic names.

## Testing

Lint clean · vitest **97** · snapshots **4** · `nuxi generate` **29 routes** · Playwright **58** against the production build · 10 locales aligned · `/search` verified against the live `api.findadoc.jp`.

`tests/e2e/search.spec.ts` mocks GraphQL, and the mock now **enforces the real API's page ceiling**. A permissive mock previously passed every test while production was broken.

## Notes for the reviewer

- **Translations beyond English and Japanese are English placeholders** for new keys, and the Japanese has had no native review.
- `docs/` is included here; it was previously on an unmerged `docs/refresh-handoff` branch with no PR. `docs/handoff.md` records the full change set and what remains.
- The **backend location filter** (`docs/handoff.md` §2a) is still the next piece of work. Once it lands, `loadDirectory()` is the single place to swap so the client stops downloading the whole directory.
- `pages/terms.vue` still needs real legal copy and i18n extraction.

## Summary by CodeRabbit

- **New Features**
  - Added a responsive directory search with filters, map view, result cards, pagination, URL-synced searches, and detailed facility panels.
  - Added light, dark, and automatic color-scheme controls.
  - Added accessible focus management, loading states, improved form validation, and map-unavailable messaging.
- **Improvements**
  - Refreshed homepage, About, submission, legal, footer, and navigation layouts.
  - Introduced consistent buttons, cards, fields, typography, and spacing.
  - Improved localized search, navigation, theme, and error messaging.
- **Bug Fixes**
  - Added safeguards for stalled requests and improved directory loading reliability.


aaronculbert:
## CodeRabbit review — triage

All twelve findings reviewed. Ten actioned, two declined with reasons.

| Finding | Decision |
|---|---|
| Drawer needs a focus trap / Escape off the panel | **Fixed** — and the search details panel had the identical defect, so both now share one composable |
| `closeDetails` can reopen the previous facility | **Fixed** — Back is only correct when the previous entry has no `?facility=` of its own |
| Clearing optional second language shows an error | **Fixed** — display and `validateFields()` now agree |
| `aria-haspopup` inconsistent with `role="group"` | **Fixed** — replaced with `aria-controls`, it's a disclosure not a menu |
| `/about` jump list has the wrong accessible name | **Fixed** — now "On this page" |
| brand.md table row missing a cell | **Fixed** |
| handoff.md Auth0 claim overstated | **Fixed** — generated HTML vs runtime SDK now distinguished |
| handoff.md stale uncapped-concurrency note | **Fixed** — the cap was added later in the same pass |
| handoff.md contradictory `localeErrors.notFound` status | **Fixed** |
| audit summary over-broad on server-side filtering | **Fixed** — the API filters specialty and language, not location |
| Untranslated strings in `cn.json` | **Declined** |
| Untranslated copy in `pt.json` | **Declined** |

**On the two declined:** new keys carry English placeholders in the eight locales beyond English and Japanese. That's a known, recorded tradeoff — machine-translating medical-adjacent UI copy into languages nobody on the review can check is worse than an obvious placeholder. It needs native speakers, and it's flagged in the PR description and `docs/handoff.md`.

Two findings were worth more than their labels suggested. The focus trap was reported against the drawer only; the same defect existed in the search details panel, so both are fixed together. And `closeDetails` was something I had spotted myself and judged low-risk — two independent reads landing on it was reason enough to fix it properly rather than reason about reachability.

## Also in this push

**CI was failing for an unrelated reason.** `nuxi start` shells out to `npx serve`, and `serve` isn't a dependency, so every run downloaded it inside the 60-second `wait-on` budget. That timed out on 2 of 3 runs and the Playwright tests never ran at all. Added as a devDependency.

**Map auth failures are now visible.** The Maps SDK logs to the console and leaves an empty box, which is exactly how the deploy-preview referrer problem went unnoticed. `gm_authFailure` now renders a short message instead.

Local: lint clean, 97 unit, 4 snapshot, 60 e2e (2 new regression tests), 29 routes prerendered, 10 locales aligned.

NabbeunNabi (approved):

PR #1817, opened 2026-09-04
ci: add TypeScript typecheck script and CI job

### LINKED ISSUE #1801: chore(ci): add nuxi typecheck script and CI workflow
# Problem

There is no full TypeScript check in this repo. CI runs ESLint (`lint.yml` → `yarn lint:ci`), Vitest, Playwright, locale lint, and GraphQL codegen — not `nuxi typecheck` / `vue-tsc`.

`package.json` also has no `typecheck` script. ESLint uses `typescript-eslint` with `project: true` on `.ts`/`.js` for a few rules (`no-unused-vars`, `no-shadow`). It does **not** use type-checked configs, so it is not a substitute for `vue-tsc`.

That gap showed up in #1786 / #1800: invalid Nuxt 2 `hid` meta fields and a removed `@nuxtjs/i18n` v10 `lazy` option were only caught by the IDE language server, not by `yarn lint`.

# Success

PRs fail CI when TypeScript / Vue SFC types are wrong, and contributors can run the same check locally.

### Requirements

- Add a `yarn typecheck` script that runs `nuxi prepare` (or reuse `prepare:ci`) then `nuxi typecheck`
- Add a GitHub Actions workflow (new file or a job on `lint.yml`) that runs `yarn typecheck` on PRs and pushes to `main`, matching the existing Node 22.12.0 + yarn immutable install pattern
- Document the script next to `lint` / `test` if README or CONTRIBUTING lists those commands
- First PR may need to fix existing type errors so the job starts green (or track remaining errors in follow-ups if the initial set is large)

# Implementation notes

```bash
yarn nuxi typecheck
```

`tsconfig.json` extends `./.nuxt/tsconfig.json`, so `nuxi prepare` must run first (same as `lint:ci` / `test:unit`).

Do not treat ESLint as the typecheck. Keep lint and typecheck as separate scripts/jobs.

Related: #1800

pranavsivaprasad230501:
## Summary
CI runs ESLint, Vitest, Playwright, locale lint, and GraphQL codegen, but nothing runs `nuxi typecheck` / `vue-tsc`. `typescript-eslint` here isn't configured with type-checked rules, so it isn't a substitute — as noted in the issue, #1786/#1800 had type errors only caught by the IDE.

- Added a `yarn typecheck` script (`nuxi prepare` via `prepare:ci`, then `nuxi typecheck`).
- Added `vue-tsc` as a devDependency pinned near the project's `typescript` version — without it, `nuxi typecheck` falls back to an ad-hoc `npx` install that can resolve an incompatible `typescript`, and fails immediately with `ERR_PACKAGE_PATH_NOT_EXPORTED`.
- Added a `typecheck` job to `lint.yml`, mirroring the existing `eslint` job's Node 22.12.0 / immutable-install setup.
- Fixed the 5 pre-existing type errors so the job starts green (per the issue's guidance):
  - `utils/arrayUtils.ts`: the Fisher-Yates swap needed non-null assertions under `noUncheckedIndexedAccess` (indices are always in-bounds by construction).
  - `components/moderation-panel/ModMainContent.vue`: a path segment from `.split('/')[2]` is `string | undefined` under `noUncheckedIndexedAccess`; defaulted to `''` before the `Array.includes` check (same runtime behavior).
  - `stores/moderationSubmissionUnsavedStore.ts`: the deprecated `registerEditSubmissionTryClose` shim called the new leave-callback (`onLeave`, which takes no arguments) with an argument — a real bug, though currently dead code (no callers). Fixed to call the legacy callback and then proceed with the leave.
  - `plugins/unsaved-changes-plugin.client.ts`: `nuxtApp.$pinia`/`$router`/`$i18n` don't resolve to their real types at this call site, so I typed them locally (imported `Pinia`/`Router` types, and a minimal local interface for the one `$i18n` method used) rather than widening anything project-wide.

No behavior changes intended outside the one dead-code bug fix.

Fixes #1801

## Test plan
- [x] `yarn typecheck` — passes (0 errors after the fixes above).
- [x] `yarn lint:ci` — passes.
- [x] `yarn test:unit --run` — all 57 existing tests pass, including `arrayUtils.spec.ts` covering the shuffle fix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed unsaved submission edits not closing correctly when users attempt to leave.
  * Prevented route-screen selection errors when a route has an incomplete path.
* **Developer Experience**
  * Added automated TypeScript type checking to the lint workflow.
  * Improved type safety across navigation, internationalization, and array utilities.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -38,3 +38,24 @@ jobs:
         run: yarn install --immutable
       - name: Run ESLint
         run: yarn lint:ci
+
+  typecheck:
+    name: TypeScript ✨
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node-version: [22.12.0]
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: Install Yarn/ Node ${{ matrix.node-version }}
+        uses: actions/setup-node@v4
+        with:
+          node-version: ${{ matrix.node-version }}
+          cache: "yarn"
+      - name: Install Dependencies
+        run: yarn install --immutable
+      - name: Run typecheck
+        run: yarn typecheck
--- a/components/moderation-panel/ModMainContent.vue
+++ b/components/moderation-panel/ModMainContent.vue
@@ -101,7 +101,7 @@ function syncSelectedModerationViewFromQuery() {
 const setActiveScreenBasedOnRoute = async () => {
     await nextTick()
     // grab the relevant part of the path
-    const pathSnippet = routePathForModerationScreen.value.split('/')[2]
+    const pathSnippet = routePathForModerationScreen.value.split('/')[2] ?? ''
 
     // for routes that require a selected ID, check before entering the switch
     const editRoutes = ['edit-submission', 'edit-facility', 'edit-healthcare-professional']
--- a/package.json
+++ b/package.json
@@ -12,6 +12,7 @@
         "lint:locales": "node i18n/checkLocaleKeys.js",
         "lint:cleanlocales": "node i18n/cleanUnusedLocaleKeys.js",
         "lint:ci": "yarn prepare:ci && ESLINT_USE_FLAT_CONFIG=true eslint . -c eslint.config.js",
+        "typecheck": "yarn prepare:ci && nuxi typecheck",
         "prod:build": "yarn && yarn generate && nuxi generate",
         "prod:start": "nuxi start",
         "ci:build": "yarn && yarn generate && nuxi generate",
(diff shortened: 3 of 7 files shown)


aaronculbert:
Thanks for the interest. Contributions to Findadoc start with an intro in #dev-team on our Slack ([join here](https://find-a-doc.slack.com/join/shared_invite/zt-2lb3j2t1z-O9F4thzcFktqsIXNzzjzHw#/shared-invite/email)) and getting assigned to the issue before opening a PR. Closing this for now. If you join Slack and let us know which issue you'd like to work on, we're happy to review a fresh PR.

For reference, our docs - https://docs.findadoc.jp/

Issue #1790, opened 2026-08-27
feat(seo): add healthcare professional detail pages


aaronculbert:
**Blocked by:** #1787

# Problem

Same as #1789: there is no professional page. All 400 healthcare professionals are viewed inside the search UI at one URL, with `activeProfessional` held in Pinia (`stores/searchResultsStore.ts:31`) and no URL sync.

Professionals are the higher-intent entity for the site's core use case — people search for "an English-speaking psychiatrist", not for a building.

# Success

Every healthcare professional has its own server-rendered, shareable, indexable URL.

### Requirements

- [ ] Route at `/doctor/[slug]-[id]` renders a professional page — **400 pages**
- [ ] Page shows names (all locales), degrees, specialties, spoken languages, accepted insurance, and affiliated facilities
- [ ] Unique `<title>` and meta description per professional
- [ ] Affiliated facilities link to their pages (#1789)
- [ ] Search results link to these URLs instead of setting Pinia state
- [ ] Unknown or invalid IDs return a real 404
- [ ] Server-rendered — content present in the initial HTML response

# Implementation notes

`searchProfessionalsQuery` already fetches everything needed: `names { firstName lastName middleName locale }`, `degrees`, `specialties`, `facilityIds`, `spokenLanguages`, `acceptedInsurance`, `additionalInfoForPatients`, `createdDate`, `updatedDate`.

**Thin-content note:** only **17 of 400** professionals have `additionalInfoForPatients`. These pages will be largely structured facts. That is normal and acceptable for a directory, but it makes the internal linking in #1794 and the structured data in #1795 more important than usual — they are what give the pages substance.

**Privacy decision required before this ships.** Making 400 named practitioners individually indexable is a different exposure from listing them inside a search UI, and may interact with the APPI considerations in `online-booking.md`. See `seo.md` §8.5. If the answer is no, this ticket closes and #1792's language facets carry more weight.

Names are multi-locale — pick the slug from the `en_US` name where present, falling back sensibly.

PR #1821, opened 2026-09-07
feat(seo): add facility detail pages

### LINKED ISSUE #1789: feat(seo): add facility detail pages
**Blocked by:** #1787, city normalisation in `findadoc-server`

# Problem

**There is no facility page.** `pages/` contains only index, about, login, submit, terms, privacypolicy, my-page*, moderation*, and `u/[username]`.

All 465 facilities are viewed inside the search UI at the single URL `https://www.findadoc.jp/`. `stores/searchResultsStore.ts:29-31` holds `activeFacilityId` / `activeFacility` in Pinia, and a grep across every search component for `useRoute`/`useRouter` returns **zero matches**.

Consequences:

- A user who finds a clinic **cannot link anyone to it and cannot bookmark it**.
- Google has no address to index even once #1787 makes the site renderable.

For a search directory, per-entity URLs *are* the SEO product. This is the single highest-leverage ticket in the plan.

# Success

Every facility has its own server-rendered, shareable, indexable URL.

### Requirements

- [ ] Route at `/clinic/[prefecture]/[city]/[slug]-[id]` renders a facility page — **465 pages**
- [ ] Page shows name (En/Ja), address, phone, website, Google Maps link, and affiliated professionals
- [ ] Unique `<title>` and meta description per facility
- [ ] Affiliated professionals link to their pages (#1790)
- [ ] Search results link to these URLs instead of setting Pinia state
- [ ] Unknown or invalid IDs return a real 404
- [ ] Server-rendered — content present in the initial HTML response

# Implementation notes

Every field needed is already fetched by `searchFacilitiesQuery` in `stores/searchResultsStore.ts` — `nameEn`, `nameJa`, `mapLatitude`, `mapLongitude`, `healthcareProfessionalIds`, and the full `contact { address { ... } email googleMapsUrl phone website }`. No API change required beyond the normalisation below.

**Carry the ID as a URL suffix** so slug changes never break links, and resolve purely on the ID — the slug is decorative.

**Blocked on data normalisation.** The API returns both `"Chuo Ward"` (21 facilities) and `"Chuo City"` (9) for the same place. Left as-is this produces two competing URLs for one city. Must be fixed in `findadoc-server` first — tracked in ourjapanlife/findadoc-server#1004.

Structured data is deliberately deferred to #1795 to keep this ticket reviewable.

NabbeunNabi:
✅ Resolves #1789

## 🔧 What changed

Every facility gets a shareable, server-rendered URL at `/clinic/{prefecture}/{city}/{slug}--{id}`. The page resolves on the UUID only (slug and location are decorative) and 301s to the canonical path when they drift. Unknown IDs are a real HTTP 404 — there is no `/clinic` SPA rewrite.

Search result cards (and map select) now link to those URLs. `?facility=` still opens the in-search panel so old shared links keep working. Affiliated professionals are listed as text until #1790.

`nuxi generate` pages the directory in a handful of requests, writes it to `.nuxt/clinic-prerender-cache.json`, and fills clinic HTML from that. Hitting `facility(id)` per page 429'd production (~900 calls) and failed Netlify plus the old snapshot generate. Runtime / `nuxi dev` still fetches by ID.

IDs are UUIDs, so the slug/id join is `--` rather than a single hyphen. City-name normalisation (findadoc-server#1004) is not required here.

Also extracts `graphqlEndpoint()` so runtime GraphQL and generate-time listing share one local/prod URL.

Removes the Vue snapshot suite (workflow, `vitest.snapshot.config.ts`, snapshot tests, `@testing-library/vue`). Playwright covers the UI; `@vue/test-utils` stays as an `@nuxt/test-utils` peer.

## 🧪 Testing instructions

- [ ] Unit: `yarn test:unit` (new `clinicPath`, `clinicPrerender`, `graphqlEndpoint`, entity-title, and no-`/clinic`-rewrite cases)
- [ ] Open a search result card and land on `/clinic/...--{id}` with name, address, phone, website, maps, and staff
- [ ] Change the slug or city in the URL and get a 301 to the canonical path
- [ ] Hit a made-up UUID under `/clinic/...` and get HTTP 404, not a 200 SPA shell
- [ ] `/search?facility={id}` still opens the details panel
- [ ] `nuxi generate` against production should not 429; a seeded clinic URL is in the HTML with a unique `<title>` and meta description
- [ ] Snapshot workflow is gone; no `yarn test:snapshot`

## 📸 Screenshots

- ### Before
  Clinic details only existed inside the search UI (Pinia / `?facility=`).
- ### After
  Each clinic is its own document at `/clinic/{prefecture}/{city}/{slug}--{id}`.

## Summary by CodeRabbit

- **New Features**
  - Added dedicated clinic detail pages with localized names, facility information, back-to-search navigation, and localized metadata.
  - Clinic pages now use canonical, indexable URLs and are included in generated routes and sitemaps.
  - Search results link directly to dedicated clinic pages.
  - Added support for displaying facility names in the secondary language.

- **Bug Fixes**
  - Invalid clinic URLs now return proper 404 pages.
  - Page titles and social metadata update correctly for clinic details and localized routes.
  - Canonical redirects prevent duplicate clinic URLs.


aaronculbert (approved):
Overall LGTM but we need to consider these two follow-ups:
1. These new pages should be added to the sitemap
2. Opening results in a new page instead of the side drawer has a tradeoff in UX since it takes them away from the search unless they open in a new tab. The back button also doesn't persist the previous search state so it resets. I think it's particularly jarring when clicking results from the map markers instead of the search results list.
SInce the functionality still exists maybe we just change the links for the map markers so they open the side drawer.

NabbeunNabi:
I can update this one based on your suggestion since I have to rebase. I did notice the flow was a little jarring

